PR #11, opened 2012-08-01
Fix mismatched code tag

ForbesLindesay:
The code tag was missing an opening angle bracket.


tj:
damn i've done that a few times now hahaha, damn html! Jade ftw

PR #12, opened 2012-08-01
change body font color and weight

sreekumar-kr:
font color to #777 and weight to 400... that improves readability

### DIFF
--- a/style.css
+++ b/style.css
@@ -1,4 +1,3 @@
-
 /* general */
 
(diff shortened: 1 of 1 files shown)


ovaillancourt:
+1 gotta tweak my laptops screen angle when I get on the site just to read the text cause it's too light :P.

tj:
merged

PR #17, opened 2012-08-01
Update applications.html

ravidsrk:


tj:
we've gotta patch it in the jade file

PR #19, opened 2012-08-02
Fixed title.

geon:


tj:
we have to edit the jade not the html

PR #21, opened 2012-08-08
Dead link in the FAQ

emilecantin:
The link to 'Routes listing' in the FAQ is dead, this should fix it. I did not build, nor test, as I am currently on a machine without 'make'. It is a very simple change; it shouldn't break anything.


tj:
thanks!

PR #22, opened 2012-08-08
Fix typo

arlo:
Dont use this for describing req.secure usage


tj:
well the getter is defined using `this`, but yeah we should just do `'https' == req.protocol` and remove the return

arlo:
done.

PR #36, opened 2012-10-04
maxAge for cookies

hectorcorrea:
Notes that maxAge for cookies is expressed in seconds, not milliseconds. 
For more info see: http://mrcoles.com/blog/cookies-max-age-vs-expires/

### DIFF
--- a/api.html
+++ b/api.html
@@ -505,13 +505,13 @@ option defaults to "/".
 </p><pre class="js"><code>res.cookie('name', 'tobi', { domain: '.example.com', path: '/admin', secure: true });
 res.cookie('rememberme', '1', { expires: new Date(Date.now() + 900000), httpOnly: true });
 </code></pre><p>The <code>maxAge</code> option is a convenience option for setting "expires"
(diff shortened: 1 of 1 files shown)


tj:
it's milliseconds here, because ms is dominate in js-land, things like ms.js give you milliseconds `ms('5 hours')` etc

hectorcorrea:
Then we've got a bug. Either Express or Connect has a bug because these two cookies end up with a totally different expiration date: 

  res.cookie('maxage', 'test', {maxAge: 900000});
  res.cookie('expires', 'test', {expires: new Date(Date.now() + 900000)});

I think the problem is that Express/Connect are setting both maxAge and the expiration date when maxAge is used. This is what comes on the response header: 

Set-Cookie:maxage=test; Max-Age=900000; Path=/; Expires=Thu, 04 Oct 2012 16:42:50 GMT

This is incorrect. If you look at the expiration that the browser is displaying (in Chrome or Safari) it will say something like October 15th because apparently max-age takes priority.

I would just update the documentation to say that max age is in seconds and call it a day but that's up to you :)

PR #37, opened 2012-10-12
Put in a link to Connect Middleware

thesteve0:
Since Express bundles a bunch of Connect middleware I think a link to their middleware will help people who want to understand what the middleware does. It would be helpful to know if all the connect middleware is included with Express.

### DIFF
--- a/en/api/app-use.jade
+++ b/en/api/app-use.jade
@@ -85,3 +85,9 @@ section
     app.use(express.static(__dirname + '/public'));
     app.use(express.static(__dirname + '/files'));
     app.use(express.static(__dirname + '/uploads'));
+
+   p.
+    Many of the Express middleware packages come from Connect. 
+    Please consult 
+    a (href="http://www.senchalabs.org/connect/") their docs
+    if you want to learn more


tj:
I've started adding them straight to these docs since the Connect site sucks, just need to add the last few

thesteve0:
Do you need help

tj:
sure if you dont mind! that would be great, I'm pretty much just copy/pasting the existing connect middleware docs over for now http://expressjs.com/api.html#middleware

thesteve0:
Meet you in Node.js on freenode later tonight

thesteve0:
I am available now - where are you at and what would you like me to do?

thesteve0:
So I saw where you are adding in the doc - do you have access to the Jade docs for the Connect middleware pages. There seems to be a lot missing on the pages (such as https://github.com/visionmedia/expressjs.com/blob/gh-pages/en/api/mw-bodyParser.jade) when compared to the connect site  - http://www.senchalabs.org/connect/bodyParser.html.

thesteve0:
Did an autotranslation in the other pull request - I don't know what it looks like. Tell me if it is good because it should be fairly easy to do the rest - tedious, but easy

tj:
it's not too important that we have middleware source etc, so the ones I added are not 100% identical with the Connect docs but that's fine

thesteve0:
I actually liked the doc on the connect site - can we just copy those over.
Your docs there are more complete than what you showed me before.
Thanks
Steve

jonathanong:
although i agree we need to link to connect, this is the wrong place to put it. we want to put it at the top that says, "express extends connect", and in the middleware section saying "these are all available via connect and can be used without express"

#99 is an attempt at the second part.

defunctzombie:
Closing since we have determined that the connect docs will be included in another fashion.

thesteve0:
Just happy they are brought together
Thanks

PR #55, opened 2013-03-01
Fix API menu order

thomseddon:
The order on the page differed from the order in the menu which
was a little confusing.

Also fixes visionmedia/express#1515


thomseddon:
Assuming not interested

PR #56, opened 2013-03-01
Fix previous build

thomseddon:
Build in 9a4e778 wasn't quite right (by a single semi-colon)

Do I win a prize for the smallest pull request ever? :)


thomseddon:
Fixed in  d9a982a

PR #59, opened 2013-03-10
Add documentation that app.locals is a function

jonathanong:
so people stop complaining that `locals.name` does not work.


jonathanong:
just realized we shoudl add somethign about the other keywords like `cache`.

PR #65, opened 2013-04-17
Add info to the guide about triggering errors

MrTrick:
The guide explains how to handle errors, but not trigger them.
This is apparently part of the connect library, but given the time it
took me to figure out triggering I think it's justified having it here.


jonathanong:
this is a little too complicated of an example. should just be something like `next(new Error('this is a custom error'))` and/or `doSomething(next)`

PR #90, opened 2013-09-26
mbriot : prevent user to not use named properties with app.locals

mbriot:


defunctzombie:
LGTM, anyone else got an opinion?

jonathanong:
i don't mind, but we should add something about it being "fixed" in v4.

jonathanong:
actually, that'll just confuse people. we'll fix the docs when v4 comes out.

PR #98, opened 2013-10-15
Updated use() documentation to reflect that use use must be called before VERBs.

Jacob-Christian-Munch-Andersen:
See https://github.com/visionmedia/express/issues/1777

### DIFF
--- a/en/api/app-use.jade
+++ b/en/api/app-use.jade
@@ -29,6 +29,11 @@ section
     pathname.
 
   p.
+    Note that <strong>all</strong> <code>use</code> declarations must be invoked before
+    <strong>all</strong> <code>get</code> and <code>post</code> declarations. Otherwise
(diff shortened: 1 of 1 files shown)


mrlucmorin:
Though in a practical sense what you added is valid, it is not 100% true.

It is conceivable that someone would want to place an app.use(...) call AFTER the router, for example as a "last chance" middleware in the case where none of your routes were matched.

Maybe an emphasis should be put on the paragraph that says _The order of which middleware are "defined" using app.use() is very important, they are invoked sequentially, thus this defines middleware precedence._

Jacob-Christian-Munch-Andersen:
You are right, it is not 100%. But the change needs to put use in relation to get and post, I think my new edition nails it: https://github.com/visionmedia/expressjs.com/pull/100

PR #99, opened 2013-10-15
Added a comment in the Middleware section referring back to Connect middleware

mrlucmorin:


api.jade
```diff
@@ -18,20 +18,27 @@ html
...
-        h2 Response
         a(name='response')
+        h2 Response
         include en/api/res
 
-        h2 Middleware
         a(name='middleware')
+        h2 Middleware
+
+        p Express exposes its own middleware, but since Express extends
```
jonathanong, line 36: "Express exposes Connect's middleware." Express doesn't have its own middleware. I think this paragraph is a little convoluted.

api.jade
```diff
@@ -18,20 +18,27 @@ html
...
 
-        h2 Middleware
         a(name='middleware')
+        h2 Middleware
+
+        p Express exposes its own middleware, but since Express extends 
+          a(href='http://www.senchalabs.org/connect/') Connect
+          |, all the middleware exposed by Connect are also exposed by Express.
+
+        p This documentation is still under construction, so for missing middleware, please refer to the corresponding one in
```
jonathanong, line 40: please refer to Connect's

api.jade
```diff
@@ -18,20 +18,27 @@ html
       #right
         include en/api/express
 
-        h2 Application
```
jonathanong, line 21: what's with these changes? shouldn't matter but JW

api.jade
mrlucmorin, line 36: What about the router ? Isn't that a middleware exposed only by Express ?

api.jade
mrlucmorin, line 21: Had to somehow hack into the html and css to make sure the web site header gets out of the way when going to the middleware section, or else the section title gets hidden under it.

api.jade
jonathanong, line 36: hmmm true, but i think we've decided to remove that in v4.

How about just, "Express exposes Connect's middleware as well as its own. More thorough documentation can be viewed in Connect's documentation." or something. i still don't think that sounds right because we say documentation twice.

mrlucmorin, line 36: Sounds good to me. My goal was to make it clear that the Connect docs should also be read.

jonathanong:
will you be updating this PR? 4.x doesn't even include connect or the middleware, so maybe we should just close this

PR #100, opened 2013-10-15
Update app-use.jade, better explaining the order of use and VERBs

Jacob-Christian-Munch-Andersen:
See https://github.com/visionmedia/express/issues/1777

### DIFF
--- a/en/api/app-use.jade
+++ b/en/api/app-use.jade
@@ -29,6 +29,12 @@ section
     pathname.
 
   p.
+    Note that <code>use</code> declarations should usually be invoked before
+    <strong>all</strong> <code>get</code> and <code>post</code> declarations.
+    Otherwise they may only take effect after <code>get</code> and <code>post</code>
+    declarations.
+
+  p.
     Here's a concrete example, take the typical use-case of serving files in ./public
     using the <code>express.static()</code> middleware:
 


jonathanong:
related: https://github.com/visionmedia/expressjs.com/issues/32

it's not just `get` and `post`, but any `app[VERB]`

Jacob-Christian-Munch-Andersen:
I complained and someone suggested that I go write the fucking manual, so I did. But I'm here in the first place because I didn't know how to use the framework. That combination doesn't seem ideal.

It really seems like this project lacks someone dedicated to the documentation. Just like the code the documentation needs someone to take a big part of the work and responsibility. I'm not going to blame you for what you do contribute, just note that it doesn't seem to be enough to carry the project.

jonathanong:
yeah, docs suck, but that's why there are so many books and blogs on express. maintenance isn't going to get much better either since we're also working on https://github.com/koajs/koa, which is why more maintainers were added.

v4 should make things simpler. ex, this PR won't be an issue anymore. 

although I don't believe you're allowed to complain in open source, i like how you took the initiative to help instead of just troll github/twitter. sucks since this repo wasn't really active for a while

Jacob-Christian-Munch-Andersen:
> although I don't believe you're allowed to complain in open source

The problem with that stance is that there is no well defined boundaries between bug reports, comments, suggestions, criticism and complaints. In order to build a serious product you need user input. Some of it is going to be rude, some of it is going to be incoherent, but it almost always reflects an underlying issue that you ideally should solve. Real world considerations of course mean that you have to prioritise, but I think one should always strive to distinguish clearly between not wanting a change, and not wanting it bad enough to allocate the required time.

jonathanong:
> bug reports, comments, suggestions, criticism

i don't consider these complaints.

defunctzombie:
Um... I feel like this escalated for no reason. @jonathanong just made a comment that the documentation fix probably needs to include one more sentence about it being any verb. Maybe that could have been more clear, but that is what I think the docs should reflect.

@Jacob-Christian-Munch-Andersen Contributions are welcome and appreciated!

"... but it almost always reflects an underlying issue that you ideally should solve" no it doesn't. I see countless reports were someone just didn't read 5 words on a readme or just wants me to solve their problem for them without trying to think it through themselves.

As project maintainers we do what we can in the time we have allotted. The reason we push back on any PR or issue is because we want to make sure it is actually correct and then worth the time in maintenance further down the line.

Jacob-Christian-Munch-Andersen:
@defunctzombie Escalated? I may have used the wrong forum for a discussion, but it is not like I'm trying to start some form of conflict. I'm merely trying to broaden your scopes on user feedback.

As for your example, if you see one report from a user who didn't read 5 words, that is just noise. If you see countless, then it is a pattern and you should consider if something could be done to help them in general. Did some of them look other places for that information where it might be relevant to put it? I for instance just had an issue with a feature that I couldn't find documentation for. I found it in the code before I realised that it was in the Guide, but not the API Reference which I had searched incessantly. Silly me, but sometimes what seems like the most natural course of action to the designer just doesn't occur to a user.

I know that you can't fix every little thing, I know that you have to deal expediently with a lot of reports. But if you think that "dumb user" is ever a sole cause, you may be missing some valuable feedback.

If you disagree with me there is nothing I can do about it, but I hope that you will at least reflect on the matter.

As for the PR, I'll get back to it when/if I feel like I truly understand the system.

PR #105, opened 2013-11-18
Update style.css

ghost:
Fixes alignment issue.

### DIFF
--- a/style.css
+++ b/style.css
@@ -296,6 +296,7 @@ html[xmlns] .clearfix {
   width: 180px;
   margin: 0 25px;
   float: left;
+  min-height: 200px;
 }
 
 #boxes h3 {


defunctzombie:
What does this fix?

jonathanong:
ohhh this fixes #111. it's a hacky fix though - i'd rather redo it correctly. also, you got to unset it in the mobile media query - this makes it look weird in mobile.

ghost:
@jonathanong I agree, it's not the best solution. It was simply the fastest without rewriting the base HTML. Maybe the best approach would be as simple as:

``` CSS
#boxes section {
    display: inline;
}
```

Which produces this look [http://cl.ly/image/2N1u3s2t3L2O](http://f.cl.ly/items/0u1z190e260R170k3a0r/Screen%20Shot%202013-12-11%20at%201.27.02%20PM.png) and looks good on mobile browsers as well.

jonathanong:
No, you can unset the min height in that media query. Or say f u to old browsers and only do the min height in a media query

jonathanong:
fixed in https://github.com/visionmedia/expressjs.com/commit/fff79cca429002407ca63fed5c97f03eba0acf29

Issue #54, opened 2013-02-22
Showcase Cozy Cloud, a private personal cloud built on top of Express


frankrousseau:
Hi all,

  First, thank you for Express which is really an awesome framework!

I would like to submit [Cozy Cloud](https://cozycloud.cc/) to you for your showcase page. Cozy Cloud, is a young start-up that develops the first private personal cloud. It's an [open source](https://github.com/mycozycloud/) initiative with an emphasize on user empowerment and respect of privacy. We love the internet and cloud computing, but we want to create a new paradigm more respectful of the users. 
All the Cozy Cloud apps are built on top of [CompoundJS](http://compoundjs.com/) that is an augmented version of Express. We enjoy using Express, so we will be happy to see Cozy on your website!

Description: Cozy Cloud is a whole new paradigm for web services: it is your own private personal cloud. Cozy gives you a personal server, a place you can trust to store all your personal data and host your web apps. You can hack it, you can host it and you can delete it. You are in control and your privacy is respected. 

Website: https://cozycloud.cc/
code: https://github.com/mycozycloud/
Screenshots:

![home](https://f.cloud.github.com/assets/1336623/186036/d75e95b2-7d1b-11e2-8c92-77f6b4ff2428.png)

jonathanong:
hi, if you'd like your app to be showcased on this site, please open a PR: https://github.com/visionmedia/expressjs.com#showcasing

PR #114, opened 2013-12-11
add cozy to the applications section

frankrousseau:
Follow-up of the #54 issue.

Thank you Jonathan.


defunctzombie:
Please remove the built html file and just leave the jade and image changes.

frankrousseau:
Sorry for that. Now my commit should be clean. Tell if I need to change anything else.

defunctzombie:
LGTM @jonathanong @visionmedia any objections?

tj:
LGTM

tj:
would be nice if we documented the dimensions for people haha that's the main reason I haven't added too many, just a PITA to get that size correct

defunctzombie:
@visionmedia what are the dimensions? I can add it to the requirements section.

tj:
no clue :D

defunctzombie:
(╯°□°）╯︵ ┻━┻

tj:
¯_(ツ)_/¯

frankrousseau:
Hi all, thank you for the merge! By the way, if you have 5 minutes, have a look at Cozy, it encourages people to build personal web apps with Express. What is really good is that apps can collaborate together to provide new usages (communication is based on Axon).

jonathanong:
> Please remove the built html file and just leave the jade and image changes.

should we do this from now on? i find it difficult because i have to rebuild everything to see any changes (unless i'm doing it wrong).

defunctzombie:
@jonathanong I think change commits should only have the changes they introduced and we make build commits separately. I think that will be less error prone cause it is much easier to miss some build error in the diff noise.

jonathanong:
okay sounds fine to me. just a PITA. we should add that to the readme though.

PR #118, opened 2013-12-13
Yet another example app: FlyLatex

alabid:
FlyLatex is a real-time collaborative environment for LaTeX built on expressjs.


tj:
thanks!

Issue #116, opened 2013-12-12
fix mobile 


jonathanong:
too much absolute positioning and fixed widths

PR #124, opened 2014-01-27
Fix for issue #116 (replaced build files).

ghost:
Okay, put the build files back in and squashed with my changes, let's hope I did it right this time.

Apologies for all the branches and pull requests now :(

### DIFF
--- a/api.jade
+++ b/api.jade
@@ -15,7 +15,7 @@ html
 
       include en/api/menu
 
-      #right
+      #main
         include en/api/express
 
         h2 Application
--- a/applications.jade
+++ b/applications.jade
@@ -10,5 +10,6 @@ html
         include includes/logo
         active = './applications.html'
         include includes/menu
-      include en/applications
+      #main
+        include en/applications
     include includes/footer
--- a/en/community.jade
+++ b/en/community.jade
@@ -1,4 +1,4 @@
-#boxes.clearfix
+#main
   section#mailing-list
     h3 Mailing List
     p.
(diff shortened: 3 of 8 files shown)


defunctzombie:
I still see `api.html` as being removed in your commit.

No apologies needed! Some of the stuff can be confusing :)

ghost:
Oh man, can't believe I missed one. Prepare yourself for yet another branch lol.

defunctzombie:
@austindavies btw, you don't have to branch. You can just fix the change in your branch and force push.

ghost:
Okay I think that did it.

Thanks for the tip!

defunctzombie:
@austindavies the reason we do it like this is to allow for manual review of the built files ourselves before pushing built changes. It also makes the commits that actually fix things less noisy.

ghost:
@defunctzombie Makes perfect sense, thanks for being patient with me.

ghost:
@defunctzombie Something isn't right with the homepage. It looks completely broken to me.

defunctzombie:
@austindavies the homepage should not have changed yet since it has not been built.

tj:
i think this buggered up the api page a bit, there's an hr at the top now, and the padding seems weird now, if you have a thin-ish page the menu overlaps now

ghost:
@visionmedia Just noticed, and you're right, it's the reason I chose to hide the menu on smaller displays. I'll fix the padding so content adapts to the menu so they never overlap. I'll remove the hr too, not sure why I kept it in the first place, looks neater without.

tj:
rad thanks man!

tj:
oh yeah you're right, home page is broken haha.. we should stop accepting CSS patches after this, too much to review

defunctzombie:
web dev is hard :'(

PR #126, opened 2014-02-03
Mention that type is aliased as contentType

stuartpb:


jonathanong:
no longer mergable :(

stuartpb:
Rebased!

PR #129, opened 2014-02-13
Update to latest jade/deps

knownasilya:
The purpose of this is to make it super easy to contribute to the docs, which will only promote express all the more.

I've moved over from `make` to `gulp`, which will work on any os. I've also update to the latest jade, to be current to the docs. Also updated the `readme`, which includes one less for building. 

Note: In the future I'll probably setup a watch/reload on template change.

### DIFF
--- a/api.jade
+++ b/api.jade
@@ -1,4 +1,4 @@
-!!! 5
+doctype html5
 html
   head
     title Express - api reference
@@ -8,7 +8,7 @@ html
     section#content
       header
         include includes/logo
-        active = './api.html'
+        - active = './api.html'
         include includes/menu
 
       include includes/mixins
--- a/applications.jade
+++ b/applications.jade
@@ -1,4 +1,4 @@
-!!! 5
+doctype html5
 html
   head
     title Express - applications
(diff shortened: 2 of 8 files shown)


tj:
make works on any os as well. the jade defined in ./package.json should work just fine you just need to `npm install`. I don't want to use any of these js build tools, there's a new one every week, it'll just keep changing

defunctzombie:
-1000 for gulp. I am all for updating to latest jade but gulp and grunt are silly tools that just incubate millions of plugins for no reason.

Can we get an updated PR with just updated deps and none of the other stuff.

knownasilya:
I gotta say though, `make` didn't work for me, but only gave me trouble.. It wouldn't build if there were no changes at the top level jade files.

The purpose isn't to use a new build system, but to make it easy to contribute. Everyone knows javascript, and jade uses javascript, so does express. It is assumed that anyone that is going to contribute will know javascript. If the build file doesn't work, then they have to fiddle with `jade` directly or learn `make`.. and it doesn't work properly.

tj:
ah yea that's just us setting up the make deps incorrectly, easy fix

knownasilya:
Ok, I'll just remove `gulp` and keep the updates. I hope you update the Readme/makefile for building.

jonathanong:
btw this isn't mergable

PR #132, opened 2014-03-05
Add to the guide an example of csrf use

Dupey-Gordien:


defunctzombie:
This looks like code for express 3, would prefer to focus on 4 compatible code for new PRs.

jonathanong:
this is also too long for an example. and tab delimiters ew

PR #134, opened 2014-03-16
Fix typo in "req.params" documentation

alexanderheldt:
req.params is an object, not an array.


defunctzombie:
In express 3 it resembled array like features iirc. I think express 3 docs could remain as is.

alexanderheldt:
You are correct. This fix now only applies to the 4.x documentation.

PR #141, opened 2014-04-06
fix variable reference

FND:


defunctzombie:
The code is correct as is

PR #145, opened 2014-04-24
Update guide.html

neilhem:
Add compass to the supported engines list in options example


defunctzombie:
https://github.com/visionmedia/expressjs.com/blob/gh-pages/Contributing.md

PR #153, opened 2014-05-09
Fixes a typo: res.headerSent => res.headersSent

AMKohn:


dougwilson:
See https://github.com/visionmedia/express/issues/2109 for back story.

PR #160, opened 2014-05-26
Gives an example of next('route')

dresende:
This change explains and adds an example of calling `next('route')` inside a `Route.param()` as described in strongloop/express#2142

Feel free to correct my English, I'll change if needed.


hacksparrow:
Already documented:

http://expressjs.com/4x/api.html#app.VERB
http://expressjs.com/4x/api.html#middleware.application

PR #161, opened 2014-05-26
A minor grammatical correction

adrichman:
...for the correct usage of "myriad."


Zearin:
Both versions of “myriad” [are correct](http://www.merriam-webster.com/dictionary/myriad); it can be used as either a noun or an adjective.

PR #163, opened 2014-06-12
Fix grammar

Zearin:


ritch:
:+1:  Do you mind squashing the commits into a single commit / rebasing onto gh-pages?

crandmck:
Wow--quite a few great copy edits, there!   Mostly good stuff: fixing comma splices, adding commas where needed, and the like.  I might quibble with over-liberal use of parentheses in preference to commas in a few cases, but it's just a matter of style.  Overall LGTM @ritch

PR #165, opened 2014-06-19
Clarify modification of req.params

seth-admittedly:
This clarifies the usage of req.params, including an anti-pattern example that will cause unexpected behavior.

### DIFF
--- a/4x/en/api/app-param.jade
+++ b/4x/en/api/app-param.jade
@@ -29,7 +29,7 @@ section
     });
 
   p.
-    The <code>req.param</code> object is generally not safely writable. Only keys which are original to the object can be modified.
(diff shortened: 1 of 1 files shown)


4x/en/api/app-param.jade
```diff
@@ -29,6 +29,27 @@ section
     });
 
   p.
+    The <code>req.params</code> object is generally not safely writable. Only keys which are original to the object can be modified.
```
ritch, line 32: This applies more to `req.params` than `app.param()`, not sure this belongs in this section.

IMO it is sufficient to mention in the `req.params` section that the object is read only.

hacksparrow:
Does not belong to `app.param`. Also, there are many other objects, a developer might  try to set properties on. Cannot add them all, unless a decision is made about such possibilities.

PR #174, opened 2014-07-31
Update app settings doc to include subdomain offset option

jeremylevy:
The `subdomain offset` option is useful, especially during dev when you have host like `en-us.localhost:5200`.


raymondfeng:
LGTM

PR #175, opened 2014-07-31
Rename `res.sendfile` to `res.sendFile`

ForbesLindesay:
[see https://github.com/strongloop/express/pull/2266]


ritch:
Closing for now: see https://github.com/strongloop/express/pull/2266#issuecomment-50902190

nw:
This is lame. 

The best responses for closing this issue was a rationalization of node following the unix naming convention.

I think we should use node.js as the authority, not unix. 

How do you explain `fs.readFile` ?

tj:
not that node core is a good place to draw conventions from ;D

PR #194, opened 2014-08-13
module installation command fixed

hacksparrow:
`--save` was missing one `-`, and `<module-name>` was being interpreted as literal HTML and not showing up on the page.


crandmck:
Thanks for the correction.  To make this change, though, you need to actually need to modify `migrating-4.jade` not `migrating-4.html` (which the Jade template generates).

PR #206, opened 2014-08-29
fixed examples containing deprecated res.send(status, body)

jordonias:


rlidwka:
it's `res.status(406).send('Not Acceptable')` imho

jordonias:
I agree. It seems like that is the way it is done with the other examples.
I also noticed a `re.send` I made by mistake.

PR #228, opened 2014-10-07
Fixed a lapse in style convention

alwaysautumn:
Hey, just noticed that you all typically had white spaces between the closing parentheses and the opening curly braces in functions, but that wasn't the case in the example for express() in the api.

Express rocks! :smile_cat:


crandmck:
Thanks!

PR #234, opened 2014-10-12
remove duplicate app.use example

jarstingstall:
It looks like the first app.use() "Path Pattern" example was duplicated. This PR just removes the duplicate.


crandmck:
Thanks!

PR #240, opened 2014-10-14
Update basic-routing.md

slamDuncan52:
put and delete requests not actually being routed to /user, being routed to home, '/' .


crandmck:
Thanks!  How did that slip through?!

BTW, I re-created the HTML using `make` to publish this fix --  37dcd453eed628298c7b5d3aaff210589a16c817.

PR #241, opened 2014-10-15
Add 'Mastering Express' book to resources

alessioalex:


alessioalex:
Never mind, I need to put it first because of the launch date. I will rectify this now.

PR #242, opened 2014-10-15
Add 'Mastering Express' book

alessioalex:
I have added it at the top, as per the launch date. Let me know if that's correct. Thanks!


crandmck:
Thanks. I'll regenerate the HTML file.

PR #247, opened 2014-10-16
basic-routing.md apostrophe and run-on phrase update

germ13:
Added apostrophe.  The sentence defining routing seems too long.  Separated the definition of _endpoint_ into another sentence.


crandmck:
Actually, the sentence as written is grammatical.  Perhaps it is too long, but that is a matter of judgement.  The apostrophe is also not required.

The second sentence in the PR is in fact not proper sentence (it's just a phrase).  It also contains a spelling error ("enpoint"):

> An enpoint being an URI (or _path_) along with a specific HTTP request method (GET, POST, and so on).

I will, however, take it under advisement to try to shorten the sentence in question.  Thanks for your input.

Issue #250, opened 2014-10-18
API Reference Side Menu of 4.x


alanjames1987:
When clicking a side menu section link of the 4.x API reference page the wrong section gets shown.

For example: if I click "Request" in the side menu the "Application" section gets shown.

I've determined this is from a different structure of the `api.jade` file in the 3.x folder from the 4.x folder.

I will be submitting a pull request to fix this soon.

alanjames1987:
I didn't recompile the `api.jade` file to the `api.html` file when I made my changes because it was a small change. If you want me to I can compile and commit again.

hacksparrow:
The menu seems to be working fine for me, and the structure is intentional.

![image001](https://cloud.githubusercontent.com/assets/950112/4691107/cd125a96-5706-11e4-8bf2-304875817ef6.png)

Am I missing something?

alanjames1987:
It's the side menu that I'm talking about. If you click on a side menu section it will display the wrong section listing in the side menu.

![express_4 x_side_menu](https://cloud.githubusercontent.com/assets/765326/4691139/8413d8a4-5708-11e4-8d97-5be47e926c1d.png)

jordonias:
I'm having the same issue as @alanjames1987 describes. Google Chrome on Ubuntu.

hacksparrow:
Fixed - 684897ce31f60f20b1373432fb01f550c5390f52.

alanjames1987:
Awesome. Thank you for accepting the pull request.

hacksparrow:
👍

PR #251, opened 2014-10-18
Fixes Issue #250

alanjames1987:
There was an inconsistent structure between the 3.x documentation and the 4.x documentation on the `api.jade` page. This caused the side menu to show the incorrect side menu section when clicking on a sidebar section link.


hacksparrow:
Thanks!

PR #262, opened 2014-10-27
Document `stat` arg to `setHeaders` option in express.static

amasad:
See https://github.com/expressjs/serve-static/issues/21


hacksparrow:
All doc edits have to be made in their corresponding markdown files, not html files. The doc has been updated with your input, thanks for bringing up the issue.

PR #267, opened 2014-11-04
Clarification that cookieParser and CookieParser(secret) are actually required

karaokeresearch:
I and a few other folks over at https://github.com/strongloop/express/issues/1859 expressed our concern about the documentation for cookieParser() and cookieParser(secret) not making sense. As it currently reads, it implies that when cookieParser() _isn't_ used, that's when you get cookie information, whereas the opposite is actually true. 

I hope you like our proposed solution.

### DIFF
--- a/3x/en/api/req-cookies.jade
+++ b/3x/en/api/req-cookies.jade
@@ -2,9 +2,9 @@ section
   h3(id='req.cookies') req.cookies
 
   p.
-    When the <code>cookieParser()</code> middleware is used this object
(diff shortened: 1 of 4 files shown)


thefourtheye:
:+1: Thanks for updating @karaokeresearch :-)

crandmck:
Thanks for your contribution!  I'll update the HTML files (or Jade, for 3.x), since updating the markdown is not sufficient, as noted in the [README](https://github.com/strongloop/expressjs.com#contributing).

karaokeresearch:
Ah, I understand now, thanks.

PR #269, opened 2014-11-05
Change tree files `express myapp` on generator.md

isRuslan:
Shellfy docs :v:

Change skeleton image to skeleton files tree.


crandmck:
@isRuslan Could you explain your rationale for this change?  The content appears to be the same, but you've just replaced the image with text.  Why is this preferable?

isRuslan:
@crandmck Yes, practically the same. It is a matter of yours taste, but the reasons are:
- Common content style on a page (text + shell snippets);
- Shell `tree` file output looks more common for everybody than Finder window. Not-OS-X users will not be frustrated about what this window is;
- count of directories | files (not so important, but pleasant).

Just one opinion.

hacksparrow:
Makes sense.

crandmck:
WFM.

PR #282, opened 2014-11-27
Changed router examples so express variable is defined

philspitler:


Fishrock123:
@philspitler sorry, but this is actually how it works. :)

Router is declared on the object the require('express') returns.

Besides, `var express = express()` doesn't make much sense given that variable is already defined? :)

Perhaps you are thinking of [`app.route()`](http://expressjs.com/4x/api.html#app.route)?

guide/using-middleware.html
```diff
@@ -184,4 +184,4 @@ <h3 id='middleware.third-party'>Third-party middleware</h3>
...
\ No newline at end of file
+</script><a id="top" href="#"><img src="/images/arrow.png"></a><footer><div id="footer-content"><div id="github"><iframe src="http://ghbtns.com/github-btn.html?user=strongloop&amp;repo=express&amp;type=watch&amp;count=true" allowtransparency="true" frameborder="0" scrolling="0" width="110" height="20"></iframe></div><div id="sponsor">The Express project is sponsored by <a href="http://strongloop.com/">StrongLoop</a>.</div><div id="fork"><a href="https://github.com/strongloop/express?_ga=1.39756451.1408343404.1407943788">Fork the website on GitHub</a></div></div></footer></body></html>
```
Fishrock123, line 187: @hacksparrow can we make it generate end of file newlines?

guide/using-middleware.md
```diff
@@ -106,7 +106,7 @@ Router level middleware are loaded using `router.use()` and `router.VERB()`.
 The middleware system created at the application level in the example above, can be replicated at the router level using the following code.
 
 ```js
-var app = express();
+var express = express();
```
dougwilson, line 109: This change will not function at all.

philspitler:
Derp.  I clearly wasn't thinking on that one.  What I was really assuming it would be is more like:

``` javascript
var app = express();
var router = app.router();
```

After looking at the API docs, I see this isn't the case.  Sorry for wasting anyone's time and thank you for responding so fast.

PR #284, opened 2014-11-28
Fix typo in app-use.md

trizko:
changes line 198

from:
Serve static content for the apfrom the "public" directory in the application directory.

to:
Serve static content for the app from the "public" directory in the application directory.


aredridel:
:+1:

PR #289, opened 2014-12-11
Fix typo

peteut:


crandmck:
@dougwilson - could you also review this change please?  It seems puzzling to me since it's in section on `app.param()` but this changes example code to use `router.param()` ...

dougwilson:
This change is legit; just viewing the diff doesn't provide enough context :(

crandmck:
OK, thanks, I'll go ahead and merge it.

PR #290, opened 2014-12-13
Update router-route.md

ifraixedes:
Although with the comment into `.all(function (req, res, next)` is clear that it is a middleware and `next` must be called, in other examples always appear the `next` call, so I think that it is more consistent with the rest of the docs.

### DIFF
--- a/4x/en/api/router-route.md
+++ b/4x/en/api/router-route.md
@@ -18,7 +18,6 @@ router.route('/users/:user_id')
 .all(function(req, res, next) {
   // runs for all HTTP verbs first
   // think of it as route specific middleware!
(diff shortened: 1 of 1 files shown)


crandmck:
@dougwilson - Could you please review this addition.  (@hacksparrow is out for essentially rest of the month).

dougwilson:
Just the `next()` is really necessary; the other part of yelling at the doc readers in caps and printing to console is not necessary.

ifraixedes:
I've updated the PR, removing `console.log`

dougwilson:
LGTM

crandmck:
Thanks @dougwilson and @ifraixedes.

Issue #291, opened 2014-12-13
etag setting doc in 4.x is inconsistent with the implementation


ifraixedes:
I've realised that the 4.x documentation says that the default option for `etag` using a boolean true which is the default behaviour, enable **strong ETag**, however it's the opposite of what it is implemented.

When `etag` is not specified **weak Etag** is applied https://github.com/strongloop/express/blob/master/lib/application.js#L54

When the `etag` setting is set, then it calls `utils.compileETag` (https://github.com/strongloop/express/blob/master/lib/application.js#L321) which resolve a true boolean in `weak` than `strong` (https://github.com/strongloop/express/blob/master/lib/utils.js#L170)

dougwilson:
Yep, `true` = `'weak'` for the ETag, with the default being `true`/`'weak'` and in the source it sets to `'weak'` so it's clearer on what it means, with the Booleans being backwards compatible for `.enable` and `.disable`. But yes, the 4.x docs need to be fixed :)

crandmck:
Fixed in PR #292. I updated HTML in 347d0d25b7431517af8e13acfcf038302d77c0e5

PR #292, opened 2014-12-14
Update app-settings.md `etag` true value

ifraixedes:
PR for #291


crandmck:
Thanks for the correction.

Issue #297, opened 2014-12-26
Case sensitivity in documentation files


joepie91:
While setting up the repository locally in order to contribute a fix, I ran across the following issue:

```
sven@linux-etoq:~/projects/expressjs.com> make
./node_modules/.bin/jade --path 3x/api.jade < 3x/api.jade > 3x/api.html
./node_modules/.bin/jade --path 4x/api.jade < 4x/api.jade > 4x/api.html

/home/sven/projects/expressjs.com/node_modules/jade/lib/runtime.js:202
  throw err;
        ^
Error: 4x/en/api/app.jade:43
    41| section
    42|   h3(id='app.METHOD') app.METHOD(path, [callback...], callback)
  > 43|   include:md ./app-method.md
    44| 
    45| section
    46|   h3(id='app.all') app.all(path, [callback...], callback)

ENOENT, no such file or directory '4x/en/api/app-method.md'
    at Object.fs.openSync (fs.js:438:18)
    at Object.fs.readFileSync (fs.js:289:15)
    at Parser.parseInclude (/home/sven/projects/expressjs.com/node_modules/jade/lib/parser.js:567:20)
    at Parser.parseExpr (/home/sven/projects/expressjs.com/node_modules/jade/lib/parser.js:223:21)
    at Parser.block (/home/sven/projects/expressjs.com/node_modules/jade/lib/parser.js:707:25)
    at Parser.tag (/home/sven/projects/expressjs.com/node_modules/jade/lib/parser.js:816:24)
    at Parser.parseTag (/home/sven/projects/expressjs.com/node_modules/jade/lib/parser.js:737:17)
    at Parser.parseExpr (/home/sven/projects/expressjs.com/node_modules/jade/lib/parser.js:211:21)
    at Parser.parse (/home/sven/projects/expressjs.com/node_modules/jade/lib/parser.js:122:25)
    at Parser.parseInclude (/home/sven/projects/expressjs.com/node_modules/jade/lib/parser.js:594:22)
make: *** [4x/api.html] Error 8
```

While `4x/en/api/app-method.md` does not exist, `4x/en/api/app-METHOD.md` does. It would seem that case sensitivity is not taken into account by the maintainer of the website - using a Windows system, perhaps?

To make the build system function correctly on all systems, either case sensitivity should be taken into account throughout the build files, or all filenames should be left as lowercase.

dougwilson:
[...]

PR #298, opened 2014-12-26
Update router-Router documentation

joepie91:
After trying to work out for 30 minutes why my parameters defined in middleware were not being carried over, I eventually found a thread on GitHub that indicated the parameter only existed since `4.5.0` (strongloop/express#2257). As I was using `4.2.0`, this parameter didn't work for me and it would quietly fail.

This fix changes the documentation for the `mergeParams` parameter to clearly indicate that it has only existed since `4.5.0`.

Submitting as a markdown-only GitHub edit rather than a regular pull request, because of local build issues (#297).

### DIFF
--- a/4x/en/api/router-Router.md
+++ b/4x/en/api/router-Router.md
@@ -12,7 +12,7 @@ Options is an optional object to alter the behavior of the router.
 
 * `caseSensitive` Enable case sensitivity, disabled by default, treating "/Foo" and "/foo" as the same
 * `strict`  Enable strict routing, by default "/foo" and "/foo/" are treated the same by the router
-* `mergeParams` Ensure the `req.params` values from the parent router are preserved. If the parent and the child have conflicting param names, the child's value take precedence. Defaults to `false`.
+* `mergeParams` *Since Express.js v4.5.0.* Ensure the `req.params` values from the parent router are preserved. If the parent and the child have conflicting param names, the child's value take precedence. Defaults to `false`.
 
 The router can have middleware and http VERB routes added just like an application.
 


dougwilson:
To add this change, I think everything needs to be annotated as what version each feature was added, otherwise this isn't very valuable, as lots of features don't exist in 4.0.0 and were simply added later (it's even more pronounced in 3.x).

joepie91:
That may not be a bad idea. That said, this particular case is rather tricky - if it fails because you're using a version that is too old, it _silently_ fails. You don't really get any indication that the option is not yet supported, other than that it simply... does nothing.

dougwilson:
> this particular case is rather tricky - if it fails because you're using a version that is too old, it silently fails

Yes, I understand that :) There are a lot of things that do that, especially in the 3.x series, because of the standard way options objects are implemented on JS projects where it's not an error to pass in unknown properties.

Currently which version features were added is only tracked in the `History.md` file in the Express code repo (and on your hard drive in `node_modules/express/History.md`), but I think we really need to add a _introduced in version x.x.x_ to every single thing in our documentation, rather than just single one single one out is all I'm saying :)

Basically, to add this one, we need to add it _everywhere_ or _nowhere_--an all or nothing project to add this information to our documentation, otherwise it's really just _more_ confusing, since if you see this note in the docs, you'll figure that everything that doesn't have that note doesn't have a minimum version (i.e. it would be out of place).

I'll see what @hacksparrow and @crandmck think about it :)

dougwilson:
@joepie91 TL;DR I agree this information is valuable to have in the docs.

hacksparrow:
@joepie91 submitting changes to .md files is the recommended way to contribute to the docs.

Your point is valid, at the same time, it might not be a good idea to log feature history in the docs - the _everywhere_ or _nowhere_ thing, as mentioned by @dougwilson. Also, a property or method's behavior might change over time - we will need to log their history as well, if we are to go that route.

It might be best to have separate docs for each release versions like how http://apidocs.strongloop.com/loopback/, http://docs.mongodb.org/manual/ and others are doing.

crandmck:
I agree, if/when we do this, we should note "version added" consistently across the board.  As @hacksparrow notes, many projects provide fully-versioned API docs, but that's difficult given that our API docs are hand-written (vs. generated).  In a perfect world, we would add complete version support doc, but we have limited resources so we need to be a bit more circumspect.  To be clear, let's just focus on 4.x; We're not really expending effort to improve 3.x docs.

It seems to me that there are (at least) two levels to the issue: documenting option properties and more extensive version support documentation such as adding/deprecating entire functions, changing signatures, etc..  In practice it seems like the former may be the most useful, as noted in the above discussion, due to the "silent failure" nature of options handling.  Re the latter, I don't even know how much of that there is, so perhaps its not an issue...?  

I suggest a compromise where we annotate version support for options properties only.   That would address the specific issue raised in this PR, and improve the usefulness of the docs without undue effort (I think).

A cursory search shows only a few functions with options arguments in addition to the one in this PR:
- http://expressjs.com/4x/api.html#res.sendFile

The following functions list an options arg but do not document the options properties:
- http://expressjs.com/4x/api.html#res.cookie 
- http://expressjs.com/4x/api.html#app.render
- http://expressjs.com/4x/api.html#res.clearCookie

I presume we should add docs for those options?  Is that just an oversight, or am I missing something?
So, it looks like adding "supported in version 4.x.x +" wouldn't be a huge effort.  

Anyway, does that sound like a reasonable compromise @dougwilson @hacksparrow ?

dougwilson:
> Anyway, does that sound like a reasonable compromise 

Indeed it does, @crandmck :)

> The following functions list an options arg but do not document the options properties [...] Is that just an oversight, or am I missing something?

It's an oversight.

hacksparrow:
Addressed the original PR (with a table column) - http://expressjs.com/4x/api.html#router, as well as 
- http://expressjs.com/4x/api.html#res.sendFile
- http://expressjs.com/4x/api.html#res.cookie
- http://expressjs.com/4x/api.html#app.render
- http://expressjs.com/4x/api.html#res.clearCookie

Make sure to hard refresh to load the updated CSS file.

@crandmck If it looks good, you can you close the PR.

crandmck:
Great, thanks....this PR was addressed in bdaeb9689556a4f85faf8bc8e94dcb7405af112a.

However, @hacksparrow http://expressjs.com/4x/api.html#res.clearCookie has a small typo that breaks the link to http://expressjs.com/4x/api.html#res.cookie.  Need to fix that, then we can close.

dougwilson:
> http://expressjs.com/4x/api.html#res.sendFile

you can add that the `lastModified` option was added in 4.9.0

> http://expressjs.com/4x/api.html#app.render

`res.render` oddly takes `options` and `locals` merged together--stuff like the `cache` option can be handed to the second argument of `res.render` to override caching, for example.

hacksparrow:
Typo fixed, `lastModified` info added, and published.

Yes, I remember the `res.render` object taking some config options along with local variables; `template` used to be one of them. Any suggestions for the name of the object? `options` and `locals` don't sound intuitive enough, because of the nature of the object.

dougwilson:
> Typo fixed, lastModified info added, and published.

:+1: 

> Any suggestions for the name of the object? options and locals don't sound intuitive enough, because of the nature of the object.

Yea, I'm not sure. The code itself just uses `options`, but I agree, that it then doesn't describe that it's also the `locals`. I would almost say to leave the name as `locals` and just say that there are "special locals" like `cache`. In fact, you can `res.locals.cache = false` as well, so that kind of description probably makes the most sense.

joepie91:
@crandmck That seems like a reasonable compromise to me, at least for a short-term solution.

That said, in the long term, perhaps there is some way to define 'fenced sections' in the source `.md` files that use a version spec to indicate what versions the section applies to? If the parser is extensible enough, it could be possible to generate documentation for all versions from a single set of source files (per major version; ie. 3.x would still have different documentation sources from 4.x).

If the parser is not extensible, this could perhaps still be accomplished with some pre-processing magic that, for each version, does a parse of the documentation sources with all the non-relevant sections stripped out.

I think that the maintainability wouldn't suffer (much) from this, as the changes between minor (and lower) releases don't appear to be very significant from an end-user API point of view. One concern could be the generation time required.

hacksparrow:
@dougwilson are you aware about any other property apart from `cache` that is used in the view rendering process? Looking through the code, I could not find anything else. I will add a note about `cache`.

dougwilson:
Yea, Express itself only looks at `cache`.

hacksparrow:
Added this note on `app.render()` and `res.render()`:

> The local variable `cache` is reserved for enabling view cache. Set it to `true`, if you want to cache view during development; view caching is enabled in production by default.

hacksparrow:
@joepie91 yes, technically it is very much possible to generate API docs individually for each version release, it is just that we have limited resources currently. If someone can work on a version-based doc generator for Express, you are most welcome.

hacksparrow:
@crandmck @dougwilson can we close this now?

dougwilson:
I don't have anything left to add.

crandmck:
LGTM.  Closing.

PR #300, opened 2014-12-29
Changed index to index.jade

Nemonocerater:
You need to have `app.set('view engine', 'jade')` or specify the filetype for render to work properly.


hacksparrow:
Thanks for bringing it to notice, the doc has been updated to reflect fact. Let's not make it `res.render('index.jade' ...`, though.

PR #309, opened 2015-01-11
Update books-blogs.md

dileephell:
Added node-tricks-express category blog.


crandmck:
This change won't show up until we regenerate the HTML, which will probably happen soon.

crandmck:
Updated http://expressjs.com/resources/books-blogs.html

PR #310, opened 2015-01-12
Grammar Changes to the Docs

peerwit:
Hopefully this will improve the readability of the doc :)

Change on line 12; added a "the" to reference "middleware system".
Change on line 137; added an "and" after the penultimate value in the list.


crandmck:
These changes look fine.  Would you mind also changing the corresponding markdown source file, `using-middleware.md` and adding that change to the PR?   To make a change, you really need to change both, per the README.  Thanks!

crandmck:
Thank you very much!

PR #317, opened 2015-01-22
Adding DB2 for i instructions

aaronbartell:


crandmck:
Thanks for the PR.

@dougwilson @hacksparrow can you speak to the non-standard nature of installing DB2 for i?  Does this require any additional explanation?

Other than that, @aaronbartell would you mind also committing the generated HTML file (per the README) and moving the DB2 for i entry to the top (since the entries are alphabetized).  With the appropriate anchor.  Thanks!

dougwilson:
So it looks like it's a non-standard install because it's not on `npm` and it looks like you have to have some kind of license from IBM. i.m.o we should not add anything on here you cannot install from `npm`.

crandmck:
> we should not add anything on here you cannot install from npm

I tend to agree.  @altsang would you like to weigh in here?

aaronbartell:
I agree it is unfortunate this isn't an npm (yet).  For anyone that's worked with IBM you know they have their own rules to abide to (aka a city-o-lawyers) and it would be a bummer to have this technicality hinder the IBM i `(n1)` community from adopting the combination of Node.js + Express.  FWIW, the IBM licensed product that includes Node.js and npm is free, it just uses IBM's means to download and install currently.

`n1`: The IBM i operating system (f.k.a. AS/400 and iSeries) is notoriously known to run the backbone of many banks, auto manufacturers, insurance companies, pharmaceutical, etc.

hacksparrow:
> we should not add anything on here you cannot install from npm

+1

crandmck:
Until there's a standard npm package, perhaps we could just link to the [IBM documentation](https://www.ibm.com/developerworks/community/wikis/home?lang=en#!/wiki/IBM%20i%20Technology%20Updates/page/DB2%20Access%20from%20Node.JS).  That way, anyone interested can find the info, but we're not including non-standard stuff.

Perhaps add it to a new "Other databases" section in "Database integration"...?

aaronbartell:
Thinking about this some more... In reality if an IBM i developer is on Express.js doc site then it can be somewhat assumed they are already doing Node.js development and Node.js is already installed, which subsequently means the Node.js module for DB2 already exists on their machine.

It would be a bummer if this got side-lined into a "Other databases" section because the base delivery wasn't the same.  A simple link to the [IBM documentation](https://www.ibm.com/developerworks/community/wikis/home?lang=en#!/wiki/IBM%20i%20Technology%20Updates/page/DB2%20Access%20from%20Node.JS) in place of `npm install ...` should seemingly suffice, right?

crandmck:
We haven't forgotten about this, but are in the middle of a bunch of other work on the docs.  Will try to circle back ASAP.

aaronbartell:
Thanks for the update.

altsang:
IBM i OS is a specialized runtime that has an extension for Node.js, so this actually goes beyond DB2.

There are potentially other such Node.js implementations that have similar special instructions for install/running other npm modules.  It seems strange to me to call this one out.  

It would make more sense to me if IBM had exception installations centralized in their Node.js for i documentation that included an Express excerpt along with every other exception.  

In the case of IBM i, I would prefer to have the onus of how it is supported through specialized exceptions and commands maintained by IBM developerWorks as these things can change.

crandmck:
Given the comment above, I'm going to close this.  If/when this package is published to npm, we will happily add it to the docs.

aaronbartell:
FWIW, being purist is sometimes good but I don't think it is serving the community in this case.  If idiosyncrasies are documented here then the community can readily change them vs. if they are on locked-down developerWorks pages.

There are already specific instructions for other databases on this page.  Why? - to save from having to venture elsewhere or figure it out by ones self.  IBM's agenda may never venture into adopting/documenting Waterline.

PR #325, opened 2015-01-30
Adds API docs for req.app and res.app in 4.x API reference

### LINKED ISSUE #323: Please document the app property in the request object.
I think It would be useful to know by reading the [API docs](http://expressjs.com/4x/api.html), that there's an `app` property in the request object on every middleware.

oskosk:
This pull requests adds the .jade, .md files and the final api.html rendered after running `make` . Closes #323


dougwilson:
Looks fine to me.

PR #327, opened 2015-02-02
Typo fixed

ben-lerner:


hacksparrow:
Thanks for pointing it out. The correct fix would be "They can be ...". Fixed.

PR #329, opened 2015-02-03
Update error-handling.md for v4

aenario:
Error-handling middleware for v4 should be placed after all app.use and routes calls.

Should we make this part of the doc more explicit ?


hacksparrow:
Thanks!

PR #347, opened 2015-03-05
Clarifying template documentation

sunsean:
In the "developing template engines" example, adjusting code to return template errors to the callback. This clarifies the first argument is used for errors (which is standard notation, but non-obvious in the example).


hacksparrow:
Great. Arigatou gozaimasu!

PR #353, opened 2015-03-13
add note about supported HTTP Method

ahmadnassri:
a little more descriptive in what methods are actually supported.

### DIFF
--- a/guide/routing.md
+++ b/guide/routing.md
@@ -51,7 +51,7 @@ To route methods which translate to invalid JavaScript variable names, use the b
 
 There is a special routing method, `app.all()`, which is not derived from any HTTP method. It is used for loading middleware at a path for all request methods.
 
-In the following example, the handler will be executed for requests to "/secret" whether using GET, POST, PUT, DELETE, or any other HTTP request method.
+In the following example, the handler will be executed for requests to "/secret" whether using GET, POST, PUT, DELETE, or any other HTTP request method supported in the [http module](https://nodejs.org/api/http.html#http_http_methods).
 
 ~~~js
 app.all('/secret', function (req, res, next) {


aredridel:
:+1: This surprised me as I was learning. I'd have found this documentation.

hacksparrow:
Good catch!

dougwilson:
We should add a note that doing `app.route(path).all(fn)` does not have this limitation to that a well, for people seeking an alternative.

ahmadnassri:
@dougwilson are you certain it does _not_? I user routes almost exclusively and it also has the same limitation ....

``` js
var express = require('../');
var app = express();

app.route('/hi').all(function(req, res){
  res.end(req.method);
});


app.listen(8888);
```

ahmadnassri:
unless you mean it _"also"_ has this limitation, in which case, I agree, should indicate that in docs as well :)

dougwilson:
I'm 1000% certain it does not. We're talking about _Express_ here, not Node.js core. Express does not require that you use the `http` module that's part of Node.js core--you can use `spdy` or other similar modules that don't limit the verbs. BUT if you do use that, `app.all()` is still limited while `app.route(path).all` is _not limited_.

dougwilson:
Our `app.all` literally does a white list. So if you so `req.method = 'TEST'` in your code above `app.all`, you'll see we still restrict `app.all` only to the Node.js core methods. But if you do `req.method = 'TEST'` in your code above `app.route(path).all`, then you will see your route is _actually run_. This means that for people who are using different HTTP servers (Express is _not_ an HTTP server), then they need to know what supports what.

ahmadnassri:
@dougwilson ah! yes, you did explain that before! you are correct.

I think then, to be decisive in providing a clear documentation on this, the [app.listen](http://expressjs.com/api.html#app.listen) section needs to be updated, as it currently only references http vs. https usage, expanding on it by adding the info about usage of different modules (spdy, http2, etc ...)

followed up with info about `app.METHOD`, `app.all` and `app.route().all` and how they behave in abstract in context of the http module used.

I might take a stab at drafting that (after I do some more research on other possible http modules) but I don't expect to be making a PR suggestion this week...

PR #356, opened 2015-03-18
Fix for clicking section header in API shows correct sub-sections

blainsmith:
Currently if you visit http://expressjs.com/4x/api.html and click the "Request" menu item it expands the "Application" section since the last item in "Application" is `closest()` to the one you clicked. Same goes for other sections as well.

I unified all the IDs and included the `.h2` anchors in the list of headings so it can find the correct closest one.

### DIFF
--- a/4x/api.md
+++ b/4x/api.md
@@ -11,13 +11,13 @@ lang: en
   <a id='express' class='h2'></a>
   {% include api/{{ page.lang }}/4x/express.md %}
 
-  <a id='application' class='h2'></a>
+  <a id='app' class='h2'></a>
   {% include api/{{ page.lang }}/4x/app.md %}
 
-  <a id='request' class='h2'></a>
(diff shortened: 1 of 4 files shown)


hacksparrow:
Thanks. This can be merged, if you can fix its effect on 3.x docs.

We are working on a new menu, [by the way](https://github.com/strongloop/expressjs.com/issues/351).

blainsmith:
Actually 3.x docs never had this issue that those work just fine still.

Nice, I noticed that scrolling issue too once I got home and used my laptop. I could take a stab at fixing that as well. A quick fix would be to not have the menu as a fixed position so you can scroll down and click Response. I will comment over there to keep the discussions intact.

blainsmith:
This should be a fix for the overflowing issue on smaller screens unless you had a better idea for a menu.

hacksparrow:
Thanks!

PR #358, opened 2015-03-23
(cleanup) Update error-handling.md

Rudimental:
Minor grammar change from a semi-colon to a period (after the word 'calls').
Now = 'You define error-handling middleware last, after other `app.use()` and routes calls. For example:'
Was = 'You define error-handling middleware last, after other `app.use()` and routes calls; For example:'

### DIFF
--- a/guide/error-handling.md
+++ b/guide/error-handling.md
@@ -18,7 +18,7 @@ app.use(function(err, req, res, next) {
 });
 ~~~
 
-You define error-handling middleware last, after other `app.use()` and routes calls;
+You define error-handling middleware last, after other `app.use()` and routes calls.
 For example:
 
 ~~~js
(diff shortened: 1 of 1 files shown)


crandmck:
Thanks for the suggestion.

In this case, a semicolon is appropriate since the two clauses are closely related.  Either semicolon or period is correct, but I believe a semicolon is better in this case.  For general guidelines on use of semicolons, see (for example):
- https://writing.wisc.edu/Handbook/Semicolons.html
- http://grammar.ccc.commnet.edu/grammar/marks/semicolon.htm
- http://theoatmeal.com/comics/semicolon

However, the "F" should not be capitalized.  If you will change period back to semicolon and lowercase the "f", I will merge the PR.

Rudimental:
You're welcome! Sure, I can do that.

crandmck:
Thanks again!

PR #359, opened 2015-03-27
Update faq.md

brendanlong:
Use `request.status(status).send(msg)` instead of `request.send(status, msg)` in FAQ.

I copied the code from the FAQ and got the warning:

> express deprecated res.send(status, body): Use res.status(status).send(body) instead


hacksparrow:
Thanks!

PR #360, opened 2015-03-27
Updating routing guide for route tester

dheniges:
In the linked route testing tool, routes like `/content/*` do not correctly translate to `(.*)` style regular expressions. Instead the wildcard character is escaped and it performs a literal match.


crandmck:
This is a reasonable addition.  LGTM.  @hacksparrow pls take a look and merge if you agree.

PR #363, opened 2015-03-29
Add join-io

coderaiser:
Added [join-io](https://github.com/coderaiser/join-io) express middleware.

### DIFF
--- a/resources/middleware.md
+++ b/resources/middleware.md
@@ -27,6 +27,7 @@ Here are some Express middleware modules:
   - [express-simple-cdn](https://github.com/jamiesteven/express-simple-cdn): easily use a CDN for your static assets, with multiple host support (ex. cdn1.host.com, cdn2.host.com).
   - [express-slash](https://github.com/ericf/express-slash): Express middleware for people who are strict about trailing slashes.
   - [express-uncapitalize](https://github.com/jamiesteven/express-uncapitalize): redirect HTTP requests containing uppercase to a canonical lowercase form.
+  - [join-io](https://github.com/coderaiser/join-io "join-io"): join files on a fly to reduce requests count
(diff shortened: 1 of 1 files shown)


hacksparrow:
If I understand the use-case of this middleware, it is better accomplished with the help of task runners like `gulp`, `grunt` etc.

coderaiser:
No, you could not achieve the same results with help of `gulp`, `grunt` etc.
`join-io` it's middle ware that joins files on a fly. No need to store joined file like task runners does this,
because `join-io` do it with help of [streams](https://github.com/substack/stream-handbook): it's open files and pipe them to `response` one-by-one, when everything is done, stream is closing.
`join-io` based on `readPipe` function of [files-io](https://github.com/coderaiser/files-io).

After prefix`/join:` names should be written of files that need to be combined .
This way:

``` html
<link rel="/join:/css/normilize.css:/css/style.css">
```

Minification could be enabled too. No need in intermediate step.
Everything is simple and transparent.

By the way [jsdelivr](http://www.jsdelivr.com/) works in similar way: [load Multiple files with a single HTTP request](https://www.maxcdn.com/blog/jsdelivr-better/).

dougwilson:
I don't see why we can't just add it to the list (except for the fact that the Markdown looks wrong in the PR).

coderaiser:
@dougwilson, what is wrong with markdown?

jordonias:
@coderaiser I think he means the `"join-io"` after the url. Maybe a period at the end as well to be consistent.

dougwilson:
Nevermind. I see the double quotes in there adds a title to the link. I've never seen that before, i.e. I was expecting it to look like the "Links" section in https://guides.github.com/features/mastering-markdown/

coderaiser:
@jordonias @dougwilson  so I should remove double quotes and add period?

PR #368, opened 2015-04-09
Fix code error in routing.md

zxhmike:
Handlers should be chained, so semicolons are removed.


wcp1231:
Fix #369

dougwilson:
Fully fixed in #370

PR #362, opened 2015-03-28
Add semicolons

YorkshireKev:
Just added semicolons to statements in some of the examples to make the documentation more consistent as some code had them while others didn't.

PR #370, opened 2015-04-10
Remove rogue semicolons from app.route book example

YorkshireKev:
Fixed syntax error in chained routing book example caused by additional semicolons added in change #362. Thanks to @dougwilson for spotting it.


hacksparrow:
Thanks!

PR #373, opened 2015-04-20
Changed with to will

ketanbhatt:
Corrected the comment in the example for route paths based on strings


hacksparrow:
Dhanyawad!

PR #375, opened 2015-04-24
Correct spelling of "Packt Publishing"

benburwell:
The company name is "Packt Publishing" rather than "Pack Publishing" (see https://www.packtpub.com/books/info/packt/about).


crandmck:
Thanks for the correction!

PR #378, opened 2015-05-02
<pre>  block have border

kuldipem:
code block should be visibly separate from doc


crandmck:
I think this is an improvement.  It makes the code blocks stand out more clearly.

kuldipem:
thanks @crandmck

Issue #377, opened 2015-04-30
Add new article to "Advanced Topics" : "Deploying Express in production"


crandmck:
Or maybe just "Deploying to production" (?)  Describe the basic considerations when moving to production, and using tools such as StrongLoop Process Manager, pm2, and Forever.

altsang:
Proposed Outline:

# Running Express in Production - header

(background paragraph - why you want to consider this)

# Popular/Leading Choices - header

## Strong-PM

(summary paragraph - with links)

## PM2

(summary paragraph - with links)

## Forever

(summary paragraph - with links)

call to action for additional submissions with PRs

crandmck:
@altsang Thanks, I moved the draft to the wiki for ease of use: https://github.com/strongloop/expressjs.com/wiki/Working-draft:-%22Running-Express-in-Production%22

crandmck:
@hacksparrow Can you add something to the wiki above on how to use pm2 and forever?

crandmck:
OK, the new page is added on the `process-mgrs` branch: https://github.com/strongloop/expressjs.com/tree/process-mgrs.

@altsang @ijroth PTAL

@hacksparrow There are several small technical/layout issues for which I need your help:
- The menu item "Using process managers" wraps in the dropdown menu - can you please make the menu a bit wider so it doesn't wrap? 
- To get the in-page anchors (e.g. http://localhost:4000/advanced/pm.html#pm2) to work, I had to put the anchor tag a few lines above where it should really be so the title would show up below the header.  Is there a better way to do that?
- I didn't add the pages for localized sites (es, ja, etc.).  Will you please do that?

hacksparrow:
Addressed all of them. The "in-page anchors" thing was a sad hack 😄

crandmck:
Thanks @hacksparrow 

So I am just waiting for review by @ijroth.  I think @altsang has taken a look already.

hacksparrow:
Alright.

crandmck:
Added in PR #390.

PR #390, opened 2015-06-03
Adding a page to "Advanced topics" menu for process managers

crandmck:
Comparing PM2, Forever, and StrongLoop PM per #377.
@altsang


maxx0r:
The PM2 part says you need to start with
`pm2 app.js`
which should be 
`pm2 bin/www`

per express4.

crandmck:
Actually, it says `pm2 start app.js` but your suggestion might be preferable.  

But I think that command presumes that you've created the app with the standard layout as the express-generator.  

As I look at the other sections, they are not consistent.  Thoughts @dougwilson ?

maxx0r:
Can't quite remember how i started it, but i think it was through the generator. Maybe it's a good idea to put that as a hint there, so people can at least check it when the regular command fails (pm2 start app.js just fails with sigint)

crandmck:
I agree adding some more context in that page makes sense.  

I just tested it, and`pm2 start app.js` does work if you create the app manually following instructions in [Installing](http://expressjs.com/en/starter/installing.html) and [Hello World](http://expressjs.com/en/starter/hello-world.html).  Admittedly, that might not be "best practice" but it does work.

Also `pm2 start app.js` is the command shown in the pm2 instructions at https://github.com/Unitech/pm2.

dougwilson:
Sorry, forgot to reply here :) The issue woupd be how detailed this should be, which is a balancing act. Both are correct: bin/www is for the generator and app.js for thr single file hello world.

crandmck:
Thanks... My preference would be to just add a short note in each section stating that the example command (e.g. `pm2 start app.js`) is for a simple app where`app.js` is the main script and if the app is creatd w generator, you would use a different command.

Created #852 to track this.

maxx0r:
Thanks for picking this up guys, this is encouraging for users like me  to mention fixes like these instead of keeping the fix for themselfs and let others suffer :)

PR #393, opened 2015-06-08
small update to glossary.md

chuckhousley:
missing "/id" in the example URL under "route"


crandmck:
Thanks for the correction!

PR #396, opened 2015-06-10
Change req.accepts() examples using csv-string into string arrays

mogelbrod:
See [4.x req.accepts() docs wrong - csv-string unsupported after switch to accepts module](https://github.com/strongloop/express/issues/2673).


crandmck:
LGTM. Thanks for catching!

PR #397, opened 2015-06-10
Clarify next method.

davidyang:
Add some clarifying information about the next(err) method.

### DIFF
--- a/guide/error-handling.md
+++ b/guide/error-handling.md
@@ -83,8 +83,6 @@ function errorHandler(err, req, res, next) {
 }
 ~~~
 
-# The `next(err)` method
-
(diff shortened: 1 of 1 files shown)


crandmck:
Looks OK, but I would move the new section to the bottom of the existing section (and remove the added sub-header `## Handling errors`).

@hacksparrow What do you think?

davidyang:
Hi @crandmck - have updated the pull request to reflect your comments.

crandmck:
Thanks @davidyang.  Before merging, I'd just like to get input from @hacksparrow ...

hacksparrow:
Good topic and good writeup. @davidyang can you remove the sub-heading? It breaks the context of the `next` object. Might sound obvious to us, but for those who are not familiar with Express, it can be very confusing.

Ready to merge after that.

davidyang:
Hey @hacksparrow - removed the subheading - is that what you meant?

-# The `next(err)` method

hacksparrow:
Yes @davidyang

PR #399, opened 2015-06-12
Remove clickjacking

nkorth:
Has anyone tried middle-clicking (or ctrl/cmd-clicking) on the links in the navigation dropdowns? The clicked link opens in a new tab as requested... and also in the current tab, which is very frustrating! This patch removes the offending code.

It's a particularly puzzling bit of code because the `<a>` elements in question are already styled with enough padding that they'll be easily clickable on mobile without this javascript shim.

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -194,12 +194,6 @@ $(function(){
 
   })
 
-  // sub menu navigation
-  $('.dropit-submenu li').click(function () {
-    var path = $(this).find('a').attr('href')
-    document.location = path
-  })
-
   // i18n notice
   if (readCookie('i18nClose')) {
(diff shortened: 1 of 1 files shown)


crandmck:
I don't see the behavior you describe when I right click on a menu link and choose "Open Link in New Tab" but I _do_ see it when I cmd-click on a menu link (Chrome/MacOS).  I saw the same on Chrome/Windows... I presume the behavior is platform- and browser-dependent, but didn't try any other browser/OS combinations.

In any case, I tried your branch, and it does fix the cmd-click behavior for me.  So, I'm fine with the change.  

@hacksparrow PTAL.

hacksparrow:
Valid point, merged.

PR #404, opened 2015-06-22
"node app.js" doesn't do anything by default

alvarotrigo:
It seems it needs to be `node start` or `node bin/www` as suggested [here](http://stackoverflow.com/questions/23178363/unable-to-run-node-app-js-file).

Way to reproduce it:

```
> express demo
> cd demo
> npm install
> node app.js
```

### DIFF
--- a/starter/hello-world.md
+++ b/starter/hello-world.md
@@ -37,7 +37,7 @@ The app starts a server and listens on port 3000 for connection. It will respond
 Save the code in a file named `app.js` and run it with the following command.
 
 ~~~ sh
-$ node bin/www  
+$ npm start
 ~~~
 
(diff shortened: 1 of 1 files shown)


crandmck:
@alvarotrigo Thanks for catching this.  I'm surprised no one reported this sooner!

You're right, `node app.js` doesn't work, however `node start` doesn't either (at least, not for me when I  followed steps above).  However either of the following work:
- `npm start`
- `node bin/www`

They are equivalent, since `npm start` just runs the `start` script defined in `package.json`.  I don't know if one or the other is preferred as a "best practice" for new users.
@hacksparrow @dougwilson Do you have any preference or input?

dougwilson:
It should be documented as "npm start" :)

crandmck:
OK! Thanks @alvarotrigo

PR #408, opened 2015-07-07
Move the explanation of middlewares to the top

ervinb:
As a newcomer to the framework, I feel that moving the explanation about middlewares and their role in Express to the top of the page would provide a better context for the reader.

### DIFF
--- a/guide/using-middleware.md
+++ b/guide/using-middleware.md
@@ -7,7 +7,7 @@ lang: en
 
 # Using middleware
 
-An Express application is essentially a series of middleware calls.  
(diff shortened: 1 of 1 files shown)


crandmck:
@ervinb Thanks for your PR.  I agree the old explanation needed improvement, but I don't think the suggested reorganization was ideal.  As I was reviewing that article, I found a number of things that needed improvement.  So I reorganized and reworded that page substantially, and I hope that addresses your concern.  Please have a look at the result: http://expressjs.com/guide/using-middleware.html.

Since my changes were rather sweeping, I'm going to close this PR.  If you have further suggestions, please open a new one.  Thanks again.

ervinb:
Thank you very much! It's way better now.

PR #409, opened 2015-07-07
Update hello-world.md

sdbarron:
Highlighted command to run sample was unexpectedly "npm start".  "node app.js" seems to be more correct...but, hey, I'm just "Getting Started", what do I know?

### DIFF
--- a/starter/hello-world.md
+++ b/starter/hello-world.md
@@ -37,7 +37,7 @@ The app starts a server and listens on port 3000 for connection. It will respond
 Save the code in a file named `app.js` and run it with the following command.
 
 ~~~ sh
-$ npm start
+$ node app.js
 ~~~
 
 Then, load [http://localhost:3000/](http://localhost:3000/) in a browser to see the output.


crandmck:
Per discussion in #404, `node app.js` doesn't actually work here.  Rather than the suggested change, perhaps an explanation of why we're using `npm start` instead?

dougwilson:
Looking at this page, taken by itself, it does not reference using the Express generator at all and instead simply displays the contents of a file and tell you to save it as `app.js`. This page would indeed need `node app.js` to function as the page describes.

crandmck:
OK.  @dougwilson I guess I was confused by [this comment](https://github.com/strongloop/expressjs.com/pull/404#issuecomment-114287083).  

It seems there is some confusion in general caused by a "Getting Started" flow that starts (after installation) with [Express generator](http://expressjs.com/starter/generator.html) and then goes to a [Hello world](http://expressjs.com/starter/hello-world.html) that _doesn't_ use the generator.  Perhaps we should consider making these two consistent...?  Or at least adding a note to [Hello world](http://expressjs.com/starter/hello-world.html) that this is a stripped-down minimal example NOT created with the generator, which is fine for a "Hello World" example.

dougwilson:
Exactly, the difference is confusing. My comment was because I was under the assumption we were talking about generator code, but this hello world page is not generator code, apparently.

crandmck:
OK, I added an explanatory note to [Hello world](http://expressjs.com/starter/hello-world.html) and I reordered the articles, so the flow is:
1. Installing
2. Hello world
3. Express generator

Hopefully, that will make it a bit clearer.

PR #415, opened 2015-07-15
Minor doc changes

didoarellano:


starter/faq.md
```diff
@@ -17,8 +17,8 @@ Routes and other application-specific logic may live in as many files
 as you wish, in any directory structure you prefer. View the following
 examples for inspiration:
 
-* [Route listings](https://github.com/strongloop/express/blob/master/examples/route-separation/index.js#L19)
-* [Route map](https://github.com/strongloop/express/blob/master/examples/route-map/index.js#L47)
+* [Route listings](https://github.com/strongloop/express/blob/2ac25098548f739c4f2b526b2a00aa60a74c8e75/examples/route-separation/index.js#L32-L47)
```
dougwilson: Oh, these links are good to change, since they could stale at any moment. Could you perhaps change them to be based off a git tag rather than a tree hash? Then it would be easy to identify it needs to be updated.

didoarellano: Do you mean like this instead:

> https://github.com/strongloop/express/blob/master/examples/route-separation/index.js#L32-47 ?

You want to always link to the latest version of the file and just keep updating the links when they change?

crandmck:
Is that what you had in mind @dougwilson? Everything else LGTM.

dougwilson:
No, it is not. The links are back to using "master" branch. I want them to use tags. I'm not at a computer, so cannot provide examples at this time, but what I can say is the link should not contain the word "master" nor should it contain a hash.

starter/faq.md
dougwilson: No, that's what it was before. I mean that it should use a tag, not a branch.

didoarellano:
https://github.com/strongloop/express/blob/4.13.1/examples/route-map/index.js#L52-L66

Like that?

dougwilson:
Yes, that's it!

didoarellano:
Got it. :+1:

crandmck:
Thanks @didoarellano

didoarellano:
Yay, thanks.

PR #421, opened 2015-08-10
Update app-locals.md

mmmveggies:
"cannot use in middleware" is very confusing. Are you proposing it is an anti-pattern to do so?


hacksparrow:
Note really. That was just a remnant of the old documentation. We are improving it whenever we find the opportunity to do so. Thanks for the contribution!

Issue #425, opened 2015-08-25
router.param doc states that it can accept an array as name, but it cannot 


vehsakul:
see also https://github.com/strongloop/express/issues/2742#issuecomment-134736492

hacksparrow:
Thanks for reporting. Updated.

PR #426, opened 2015-08-25
Change router.param first arg to string not array

crandmck:
Per #425.  In https://github.com/strongloop/express/issues/2742#issuecomment-134736492 Doug clearly states that router.param never accepted an array as the first arg.  I don't doubt him, but it seems a bit odd that the docs had as much detail as they did on that and an example...

@hacksparrow  PTAL.


hacksparrow:
Yes, `app.param()` accepts an array, `router.param()` does not.

dougwilson:
I think it was just an accidental confusion, in think that `router.param` just worked exactly like `app.param` :( In 4.0, when `router.param` was introduced, we decided to simplify it, rather than keep with the more complex `app.param`.

hacksparrow:
Agreed @dougwilson.

PR #428, opened 2015-08-29
Fix inconsistency in coding style

saadq:


crandmck:
Thanks @saadq I know there are more of these inconsistencies around, but we might as well fix them as we find 'em.

PR #430, opened 2015-08-29
Update index.md

yhagio:
Re-translated unclear translations to more readable, and to match the source's meaning.


hacksparrow:
Arigatou @yhagio san!

PR #436, opened 2015-09-09
Fix GitHub repository URL when JavaScript disabled

madarche:
that is when `./js/app.js` is not run on the client-side and `$('#fork').html(editLink);` is thus not run.


hacksparrow:
Good catch. Thanks!

PR #441, opened 2015-09-11
Feat improve trust proxy doc

madarche:
This PR contains global improvements to all the doc related to the `trust proxy` setting.

The only true change in the doc is that the addition that `req.hostname` is modified by trust proxy cf. [code of `request.js`](https://github.com/strongloop/express/blob/master/lib/request.js#L400)

The other changes are just systematizing links and formulations.


hacksparrow:
Thanks!

PR #445, opened 2015-09-12
Recommend using absolute path for `express.static`. Fix #443.

### LINKED ISSUE #443: Improvement for `express.static` doc
As for this page: http://expressjs.com/starter/static-files.html

I think it would be better to use `express.static(__dirname + '/public')` in at least one of the examples and explain it a bit.

I was making an Express app, and the main app requires a sub app. The files look like:

```
app.js
  /sub
    sub.js
    /public
      style.css
```

`sub.js` is required by `app.js`. When I use `express.static('public')` in `sub.js`, it refers to `/public`, not `/sub/public`, even if the code is in `sub.js`. I was confused and spent quite some time on this behavior, so I hope you can give some explanation to this behavior on the doc, so in the future people won't get confused.

octref:
Somehow the code has a badly-styled line-number, but I don't know how to fix it:

screen shot 2015-09-11 at 9 23 11 pm

### DIFF
--- a/starter/static-files.md
+++ b/starter/static-files.md
@@ -53,3 +53,9 @@ http://localhost:3000/static/js/app.js
 http://localhost:3000/static/images/bg.png
 http://localhost:3000/static/hello.html
 ~~~
+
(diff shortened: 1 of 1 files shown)


hacksparrow:
Thanks!

PR #451, opened 2015-09-26
Fix word in sentence

zmateusz:
"Not" just doesn't fit there, right?


hacksparrow:
👍 that's a more positive way to put it. Thanks!

PR #468, opened 2015-10-16
Fix misused 'which'.

TOGoS:
To make the document about "use" a bit easier to understand.


crandmck:
Thanks, we got sloppy there.

PR #477, opened 2015-10-27
Update router-param.md

j0h4nn3s:
This seems to be a Copy&Paste error from the standard app.param() documentation. As documented "Param callback functions are local to the router on which they are defined.", the router param callback would't be triggered when requesting a route in "app".
The following section (which was the param section with an array from app.param()), can be ignored, since "Unlike app.param(), router.param() does not accept an array of route parameters."


hacksparrow:
Thanks @j0h4nn3s!

Issue #478, opened 2015-10-28
next() function not documented


stratboy:
The next() method is everywhere in your docs, it's quite clear that passes control to the next 'action' but it's definitely not documented itself. Where does it come from?

hacksparrow:
@stratboy `next` in described on http://expressjs.com/guide/using-middleware.html.

@crandmck do you think we should create a page dedicated to explaining `next`?

crandmck:
@stratboy makes a good point.

Although that article does explain how to use `next()` and the requirement to use it to pass control to the next middleware, it doesn't really explain "where it comes from", and a new user might assume that it's actually part of the API, either of Express or Node.

However, IIUC, it's merely a naming convention based on the arguments to the callback function.  In theory it could have any name, but almost universal convention is to call it "next"... right?

This seems to be a common question, based on these SO topics:
- http://stackoverflow.com/questions/5384526/javascript-node-js-next
- http://stackoverflow.com/questions/13133071/express-next-function-what-is-it-really-for
- http://stackoverflow.com/questions/10695629/node-js-express-next
- etc. etc.

I think it would be helpful for new developers to provide a bit more explanation and clarity.  I don't think it needs its own page; rather, I suggest that we add a section to http://expressjs.com/guide/using-middleware.html (or perhaps in http://expressjs.com/guide/writing-middleware.html) and then cross-reference it whenever we mention using `next()` (e.g. http://expressjs.com/guide/error-handling.html, http://expressjs.com/guide/routing.html, etc.)

Thoughts?

stratboy:
Yes, I'm with you, only add the section, and one line to explain just what you explained below, it's enough.

Thank you, regards,

Luca

Rand McKinney wrote:

stratboy:
Ok, then I think it should be mentioned and briefly explained even on 
the API reference page, maybe somewhere in the app section, where you 
mainly write about middlewares. Maybe a title like 'about the next() 
callback' or 'Passing control with next()' or something.

Thank you, regards

Luca

Hage Yaapa ha scritto:

crandmck:
I believe this has been addressed, so I will close.  
@stratboy Thanks for bringing it up.

stratboy:
:) thank you

PR #480, opened 2015-10-30
Add "Elements of Middleware"

nathan-boyd:
Document the elements of an express middleware, in response to #478


crandmck:
This looks OK to me as a start.  What do you think @hacksparrow ?

But I think we need to do a bit more to address the concern raised in #478

hacksparrow:
Looks good for a start. @crandmck, we'll have to decide where the details about `next` should go.

Issue #484, opened 2015-11-03
Add text link to express repo on the footer


ricardograca:
This was already handled in #254, but it has disappeared again. Would be nice to get the link back.

crandmck:
The link is still there... Is it not showing up for you?

![repo link](https://cloud.githubusercontent.com/assets/2925364/10917060/a36974ac-8213-11e5-85d2-3335f311c7e3.png)

crandmck:
The link to the repo is where it's been for the last year or so:
![repo link](https://cloud.githubusercontent.com/assets/2925364/10917195/5c1c7d0a-8214-11e5-80e9-22a5a04a0f17.png)

I agree it's not very obvious (as noted in #254).  What about adding another link in the footer like this:

> The [Express project](https://github.com/strongloop/express/) is sponsored by [StrongLoop](http://www.strongloop.com).

ricardograca:
Yes, I remember when the "Star" button was there, but now it doesn't appear for whatever reason:

![captura de ecra de 2015-11-03 18 36 35](https://cloud.githubusercontent.com/assets/1457035/10917697/03c8fb4c-825a-11e5-8b12-383b33594064.png)

I see a 405 error on this url: `http://expressjs.com/cdn-cgi/pe/bag2?r[]=http%3A%2F%2Fwww.googletagmanager.com%2Fgtm.js%3Fid%3DGTM-5XL76H`, everything else is either 304 or 200.

There is also a warning in the console with no message string pointing back to `cloudfare.min.js`.

**Update**: The GitHub button is there on the same machine when using Chromium. Not visible on Firefox 41. Using the inspector I can see that the iframe has an empty document with empty `head` and `body`.

**Update 2**: It was the "privacy badger" extension blocking it. So, that's a vote from me for adding a text link in the footer.

crandmck:
OK, I think it makes sense.  Will open a PR.

PR #486, opened 2015-11-04
Add text link to express repo per #484

crandmck:
Added link to express repo in footer for all langs per #484 
@hacksparrow PTAL.


hacksparrow:
@crandmck makes sense. Merged.

PR #487, opened 2015-11-04
Clarify next and middleware per #478

crandmck:
I added an image instead of the ASCII art, which can have alignment issues and doesn't zoom well.
Also added some clarification of `next` per #478 and a few other tweaks.


hacksparrow:
👍 made a small change to the image size, so the font size in the image does not overwhelm the doc text.

dougwilson:
Looks like the image has a cursor in the screenshot, if you want to remove it (I highlighted it in yellow):
cursor

crandmck:
> Looks like the image has a cursor in the screenshot

Good catch @dougwilson.  I removed the cursor and updated the image.

PR #498, opened 2015-11-12
use official mongodb driver in database guide

vkarpov15:
Heya,

As the current maintainer of mongoskin, I really don't think the express docs should reference mongoskin as their mongodb driver of choice. The official driver's a better choice IMO.

### DIFF
--- a/guide/database-integration.md
+++ b/guide/database-integration.md
@@ -142,21 +142,28 @@ connection.end();
 
 ## MongoDB
 
-**Module**: [mongoskin](https://github.com/kissjs/node-mongoskin)  
+**Module**: [mongodb](https://github.com/mongodb/node-mongodb-native)  
 **Installation**
 
 ~~~sh
-$ npm install mongoskin
+$ npm install mongodb
 ~~~
 
 **Example**
 
 ~~~js
-var db = require('mongoskin').db('mongodb://localhost:27017/animals');
+var MongoClient = require('mongodb').MongoClient;
 
-db.collection('mamals').find().toArray(function(err, result) {
-  if (err) throw err;
-  console.log(result);
+MongoClient.connect('mongodb://localhost:27017/animals', function(err, db) {
+  if (err) {
+    throw err;
+  }
+  db.collection('mammals').find().toArray(function(err, result) {
+    if (err) {
+      throw err;
+    }
+    console.log(result);
+  });
 });
 ~~~
 


hacksparrow:
👍

vkarpov15:
Thanks @hacksparrow, you the man. Best GitHub user name of all time :beers:

TinChak:
Thank you for your help ! @vkarpov15  However, I doubt that whether Express is maintained. I really want to know the reason why the official recommand mongoskin as the first mongodb driver. Do you use mongoskin?

vkarpov15:
Express is not maintained? News to me. Looks like the commit activity has fallen significantly of late...

@cuitianze: Mongoskin's core benefits IMO are: 1) the `.bind()` helper for collections, so you don't have to do extra async `.collection()` calls, 2) a more convenient way of creating ObjectIDs, and 3) the `findById`, `updateById`, etc. helpers. These were a big improvement over the mongodb driver in 2012, but the mongodb driver has improved its DX significantly since then - `.collection()` is no longer async, and creating ObjectIDs is as simple as `new mongodb.ObjectId()`, no more typing out `mongodb.bson.createObjectIdFromHexString()` or whatever it used to be. The `findById` methods still don't have analogs in the mongodb driver, but IMO that's not sufficient to recommend mongoskin over the official driver.

crandmck:
>  I doubt that whether Express is maintained.

@cuitianze It certainly is!  The Express community is actively working on it, along with StrongLoop and IBM.   I can't speak to the pace of commits, but if (for example), you look at [issues](https://github.com/strongloop/express/issues) you'll see lots of activity.  N.B. it is a fairly mature and stable framework, so you might not see the kind of activity you would on something that is rapidly evolving.  There is also planning going on for future releases 5.x +.

Also, based on requests from the community, we've been improving the documentation a lot recently, which you can see by activity in _this_ repo. 

> I really want to know the reason why the official recommand mongoskin as the first mongodb driver.

I believe it was just an honest mistake/oversight.  If you are really interested, you can drill down into the history, but I don't think it's terribly productive at this point.  IAC, it has been corrected now (thanks, @vkarpov15 ).

dougwilson:
I think that maintained question was probably a mis-translation (I would hope so). A lot of work has been spread out across many sub modules, especially the `router` module, and tasks for Express 5.0, like the promise router, are large tasks and are mostly taking place off line. These two combined together of course results in a drop in commit activity on the main repository.

hacksparrow:
@cuitianze the interface of the official driver used to be pretty "unwieldy". mongoskin provided a better interface along with other advantages as listed by @vkarpov15. That's why mongoskin was used for demonstrating connection to MongoDB.

With the current state of the official drive and the recommendation from @vkarpov15 himself, it makes sense to update to the example to use the official driver.

vkarpov15:
Thanks for the clarification, and thanks for the merge :+1:

TinChak:
Thanks for all of your answer. @hacksparrow @vkarpov15  @dougwilson @calvintwr  
I just lost my way when l learned mongodb driver for Express.

PR #505, opened 2015-11-23
Standardize English docs to use the /en/ dir

hacksparrow:


crandmck:
I'd like to first merge the changes in `linkfixes` into `i18n`.  Then continue making changes in the `i18n` branch that we'll use to handoff for translation.  Once that's done, we can make any necessary CSS tweaks, etc, and then merge `i18n` into `gh-pages`.

PR #508, opened 2015-11-24
Auth docs

rdegges:
Adding a guide on authentication.


crandmck:
Wow, thanks @rdegges !  I haven't had a chance to read this thoroughly, but it looks great...
I'll read it ASAP.
@hacksparrow @dougwilson PTAL.

dougwilson:
The documentation seems fine, and I don't mind linking to the service, but I think that having the only article describing how to do authentication using a third-party data storage service is not very inviting. Think about it: a user wants to learn Express and of course they want to do user management, but then the only documentation Express provides on this is to use some third party service, hosting all your user credentials, it seems weird. I wouldn't mind adding this guide if there was also a complete guide on how to do this all on your local server.

rdegges:
Thanks! My idea was to get an authentication page structured similarly to how the database integration guide is: where there can be multiple sections describing how to do basic authentication with different tools / services. Would it be cool if I wrote a local strategy passport thingy as well and then included that? oO

dougwilson:
Yea, adding a guide for people to use passport would work. Basically, my line of thought is that even if this is a AaaS, I don't mind adding it, but really as long as it doesn't make it seem like the _only option_, which with the amount of new people we get, they'll assume it could be :) You don't have to go too deep into passport, but probably like at least something that's like a "hit the ground running" kind of thing, similar to this. You can even include in the Stormpath section advantages to using it over passport, of course.

crandmck:
@dougwilson I agree... We definitely don't want to present this as the only option.  Adding passport sounds like a good idea, and perhaps link to other alternatives as well.

rdegges:
Cool, will get this out today =) Thanks <3

rdegges:
Heyo, so I'm working on this now. It's actually quite complicated to reproduce this for passport in a reasonable way for a new user to follow. EG: Lots of steps: hashing passwords, setting up mongo / some db, creating templates / routes / etc. Would be quiteeee long to show the local strategy stuff.

Any ideas for shortening it? Open to suggestions =)

dougwilson:
Is there a way to provide very simple defaults in some way? For example, use some kind of in-memory-only store or a store that only requires a directory path? I don't know enough about Passport besides it seems popular; if there is a simpler option to demo some simple way to parity what's in this guide, that would be ideal. It doesn't have to be Passport.

crandmck:
@rdegges Any update?  I think using an in-memory data store as an example would be a good option.

crandmck:
Since it seems this has stalled, what should we do?  I agree with Doug's comment:

> I think that having the only article describing how to do authentication using a third-party data storage service is not very inviting. 

Perhaps we could accept the suggested change to the FAQ, but not include the entire additional page on using StormPath...?

rdegges:
Oh, for some reason Github never emailed me these replies! I hadn't seen any of this :(

rdegges:
I'll work on something similar to what @dougwilson proposed, thanks for the bump!

crandmck:
Great, @rdegges... Then we'll wait for your next update. Thanks!

crandmck:
@rdegges Are you still working on this?

crandmck:
@rdegges Can you please let us know if you intend to complete this?  If we don't hear from you soon, I'm going to close this PR.

crandmck:
It looks like this has been dropped, so I'm going to close this.  @rdegges if you want to pursue per the above discussion, feel free to open a new PR.  Thanks!

PR #509, opened 2015-11-25
Update basic-routing.md

v-e-y:


crandmck:
Hi @EdvardV thanks for the contribution. 

I'll merge this, but see https://strongloop.com/strongblog/express-goes-global/ ... we're in the process of having the docs translated professionally, which will supersede existing community translations.  The professional translations are not yet complete, but when they are (in a few weeks) they will replace the existing community translations and we will no longer be able to accept community contributions to the translated documents, though we will be happy to accept comments that we will pass on to the translation center.

PR #512, opened 2015-11-27
translation of 'starter' tab and index page is corrected

Croolis:
I know that this site will be translated into Russian by professionals. But now it looks realy weird.


hacksparrow:
Thanks @Croolis

PR #513, opened 2015-11-30
Add 3.x deprecated notice to security page

dougwilson:
So when looking at the security page, it's not exactly clear that even using the most recent version of the 3.x line may be secure. This is basically just adding the deprecation notice from the 3.x API page to the 3.x section on the security page. Let me know if there are any desired style changes :)


hacksparrow:
@dougwilson looks good.

PR #515, opened 2015-12-03
Fix URL to 2.x docs in nav menu

JamesMGreene:


hacksparrow:
Thanks @JamesMGreene!

PR #517, opened 2015-12-04
Fix typo

kohei-takata:


hacksparrow:
Thanks for the corrections @kohei-takata!

PR #518, opened 2015-12-05
Writing correction

rafael-franca:
Only writing correction


hacksparrow:
Obrigado @rafael-franca

PR #520, opened 2015-12-08
Remove incorrect argument information for res.json

NicholasAntonov:
The res.json documentation incorrectly stated that it only took an object or array as an argument, which isn't true as internally JSON.stringify is used, and being able to send primitive types as a json response is a very desirable feature.

### DIFF
--- a/_includes/api/en/4x/res-json.md
+++ b/_includes/api/en/4x/res-json.md
@@ -1,6 +1,6 @@
 <h3 id='res.json'>res.json([body])</h3>
 
-Sends a JSON response. This method is identical to `res.send()` with an object or array as the parameter.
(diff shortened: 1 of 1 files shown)


dougwilson:
Hi! Unfortunately this PR is not worded correctly and is very misleading. It says "This method is identical to `res.send()` with an object or array as the parameter.".

You change is to make it say "This method is identical to `res.send()` with any JSON type as the parameter (object, array, string, boolean number).", which is wrong, because res.send() with a string as the parameter sends HTML, and res.send() with a number as the parameter sends the number as the status code with a text body. res.json() with a string or number behaves very differently than res.send() with those same arguments.

PR #522, opened 2015-12-12
Improving example for PostgreSQL

vitaly-t:
Replacing the core `pg` example with the modern `pg-promise`.

### DIFF
--- a/en/guide/database-integration.md
+++ b/en/guide/database-integration.md
@@ -204,33 +204,26 @@ apoc.query('match (n) return n').exec().then(
 
 ## PostgreSQL
 
-**Module**: [pg](https://github.com/brianc/node-postgres)
+**Module**: [pg-promise](https://github.com/vitaly-t/pg-promise)
 **Installation**
 
 <pre><code class="language-sh" translate="no">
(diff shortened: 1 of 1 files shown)


bajtos:
@crandmck @ritch I am not sure who is reviewing pull requests for expressjs.com. I have assigned thi PR to Ritch, feel free to reassign as appropriate.

crandmck:
@bajtos Usually either I do; for more technical items @dougwilson or @hacksparrow does.  But if @ritch can review this, that's fine too.  Note that this PR came in the context of  https://github.com/strongloop/expressjs.com/issues/479.

dougwilson:
This seems fine, especially since looking at the old example, it was terrible. This is already an improvement, regardless of the change from callbacks to promises.

vitaly-t:
> This is already an improvement, regardless of the change from callbacks to promises.

A longer example could show how nicely it works with ES6 generators also, but I tried to keep the example as simple/small as possible. I think it's modern enough :)

crandmck:
Given Doug's comment above, I merged this.  Thanks @vitaly-t!

PR #525, opened 2015-12-16
Make sure to append /{{ page.lang }} to all the links within the site

martinerko:
I have updated all the links within the site to make sure that user will be always redirected to correct language mutation


crandmck:
Hi @martinerko Thanks for the contribution.

I see this PR adds the lang code to links in the API docs and also to links in the existing translations.  The reason we did not do this already is that API docs are not being translated, and we will soon be replacing the current translations with the professionally translated content.

I think we should probably accept the change to the API docs, because links in those docs to other parts of the docs should go to the appropriate language rather than English.  Links _to_ the API docs should fall back to English where there is not translation in that lang.  Currently, the only language with translated API docs is pt-br, IIRC.

As far as the changes to the existing translations, it's somewhat moot since we'll be replacing these with the new translated content soon (and the new translations will be based on the current en site, which does include the lang code in links).

So, I'm inclined to land the PR, but I'd like a bit more time to look it over... Other thoughts, anyone?

hacksparrow:
@crandmck @martinerko I think we should wait till the translation work is over.

crandmck:
@hacksparrow Makes sense.

martinerko:
Alright guys, by when do you expect the work on new translation will be finished?

crandmck:
We should be able to open a PR and land it this week, I hope.

martinerko:
Any updates?

crandmck:
We merged the new translations last week, and I believe all the links now include `{{page.lang}}` as they should.  So I believe we can close this PR.   

@martinerko If you find any exceptions, please open a new issue or PR.  
At this point this PR has merge conflicts and I don't think we can resolve them with the new translations.  Thanks very much for your contribution, but the timing was such that we couldn't take it.

Issue #523, opened 2015-12-14
Globalization Issue: English text in JavaScript


crandmck:
Two English strings are used in the footer: https://github.com/strongloop/expressjs.com/blob/gh-pages/js/app.js#L50.

When you view a localized page, "Edit this page on GitHub." is displayed in English.
Can we externalize this string and load the appropriate language?
I'm not sure when "Fork the website on GitHub" is displayed.

@hacksparrow I would appreciate it if you could take a look at this.

crandmck:
After looking a bit more at this, I think we should consider removing this line altogether from translated pages, because the link would take the reader to the English page, and most non-English speakers won't be interested in editing the en source page; also it might confuse the reader into thinking that the link would be to the translated source, for which we won't be accepting contributions.

So, pending agreement from the globalization team, I think we should just configure it so the "Edit this page in GitHub" line appears _only_ on en pages, not translated pages.  Thoughts?

dougwilson:
> So, pending agreement from the globalization team, I think we should just configure it so the "Edit this page in GitHub" line appears only on en pages, not translated pages. Thoughts?

Seems like a reasonable compromise to me.

hacksparrow:
@crandmck @dougwilson the link can point to the corresponding lang page. Shall I do that, instead of removing it altogether?

crandmck:
> the link can point to the corresponding lang page. 

Well, the problem is that we won't be accepting PRs for the translated pages, so it's not clear what the purpose would be.   In contrast, we will still accept PRs for contributions to English content, so that does make sense.

If someone just spots an issue with a translation, they can open an issue and refer to the page in question.

hacksparrow:
@crandmck alright.

hacksparrow:
d8ab769 addressed this.

PR #526, opened 2015-12-17
Remove English string from non-en pages footer

crandmck:
This addresses #523.  Note that the IBM translations will include the footer.
@hacksparrow PTAL.

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -47,10 +47,17 @@ $(function(){
 
   var editLink;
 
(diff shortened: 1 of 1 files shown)


hacksparrow:
@crandmck there is a `console.log()` in the commit. We would need to edit the footer pages of the existing lang files, too.

I will close this PR and implement what you are trying to achieve.

hacksparrow:
Pushed at d8ab769.

crandmck:
> there is a console.log() in the commit. 

Ooops, dang!

Thanks!

PR #536, opened 2016-01-01
fix broken links inside /ja/4x/api.html

kiwamulet:
It seems that some of the links inside the menu on the page [Express 4.x - API リファレンス](http://expressjs.com/ja/4x/api.html) does not work.

For example, menu item "Application" links to [http://expressjs.com/ja/4x/api.html#application](http://expressjs.com/ja/4x/api.html#application), not to [http://expressjs.com/ja/4x/api.html#app](http://expressjs.com/ja/4x/api.html#app).

In the same manner, "Request" should link to #req, "Response" should link to #res. 
These links on ja/4x/api.html are corrupt.


hacksparrow:
Thanks @kiwamulet

PR #554, opened 2016-01-14
Updated the windows command.

ghost:


dougwilson:
A single ampersand is the correct way to perform this action. The `cmd.exe` shell treats them differently.

Can you elaborate on why you think we need to change from a single two a double and specifically if it solves a problem or what the behavior change you desire from this change?

ghost:
@dougwilson I use windows and neither cmd.exe or powershell will accept a single ampersand.

dougwilson:
I only develop on Windows, use cmd.exe and it definitely works. If you are using a shell that mimics bash, then the Windows command will not work, because it is for Windows. cmd.exe has different behavior for single vs double ampersand, and single is the correct behavior for this command.

ghost:
That's a bit strange because when I use a single ampersand it won't work. (Win10 x64)

dougwilson:
Hi @karimdag , ampersands just don't work with PowerShell in the same way; it is just for `cmd.exe`. In PowerShell, the command should use `;` instead, but even then, using `set` is not the recommended way in PowerShell. PowerShell is just a different paradigm than `cmd.exe`.

I'm' trying to find you the official documentation from Microsoft on the two, but still looking. The best page on Microsoft's site so far that explains the difference and how they both should work is https://support.microsoft.com/en-us/kb/279253

crandmck:
So @dougwilson it sounds like the doc is OK and we can close this?  

Sorry I wasn't able to help w this, but I don't have a Windows machine (any more... my old PC just finally died :-(

dougwilson:
@crandmck, looks like the PR was actually already merged, regardless if it was ultimately the right solution. I assume @hacksparrow confirmed this on Windows 10.

dougwilson:
FWIW, I tried on Windows 10 and the original command works just fine, and has the benefit of having the single ampersand semantics. I think that simply blindly accepting this pull request is doing a disservice to this user and other users. I'm sure there is some reason this user is having issues, and we should figure out the way to understand what the pros/cons of changing the documentation is, rather than making a blind change.

capture

crandmck:
> looks like the PR was actually already merged

Silly me, somehow I missed that.

@hacksparrow Why did you merge this?  It seems questionable at best.  

> I tried on Windows 10 and the original command works just fine, and has the benefit of having the single ampersand semantics. ...

I can't try it since I don't have a Windows machine, but I tend to agree with what Doug said.  Can you please revert this until we get to the bottom of it.

ghost:
@dougwilson Hey ! I'm sorry for the delay, you were right and that this pull request was in fact unnecessary though you should know that when using powershell neither of these two methods work.

dougwilson:
> though you should know that when using powershell neither of these two methods work.

Correct, I am well aware of that, as I noted above:

> Hi @karimdag , ampersands just don't work with PowerShell in the same way; it is just for cmd.exe

dougwilson:
@crandmck @hacksparrow, what is the progress of reverting the merge?

crandmck:
@hacksparrow never did so, and for some reason I was unable to revert it, so I just fixed with 0d191dd2e5c2cb4f6c8bb00ac3e38baa43ff8927.

ghost:
@dougwilson sorry

Issue #541, opened 2016-01-07
Multiple duplicates of API docs exist


crandmck:
Due to the way we restructured the site for localization, we have API docs for each language, even though the API doc was not translated (except for pt-br community translation).  So, for example, we have:
- English API docs in https://github.com/strongloop/expressjs.com/tree/gh-pages/_includes/api/en
- Duplicate  in https://github.com/strongloop/expressjs.com/tree/gh-pages/_includes/api/de (which creates http://expressjs.com/de/4x/api.html#app.listen where links from the German localized site go)
- Duplicate in https://github.com/strongloop/expressjs.com/tree/gh-pages/_includes/api/es (which creates http://expressjs.com/es/4x/api.html#app.listen where links from the Spanish localized site go)
- And so on...

This is a problem for maintenance, because it means whenever we update the English API docs, we must make the same change in all 9 language duplicates (because pt-br is translated, so not there).

It would be best to maintain only ONE English version of the API docs, while still allowing for translated API docs (currently only pt-br) and potentially for other langs in the future.  Any ideas on the best way to accomplish this? @hacksparrow

hacksparrow:
@crandmck there is a problem in redirecting the user to the ONE English version - the user is now in the en "docland". This redirection might be seen as a "bug" by many.

crandmck:
OK @hacksparrow I can see the issue.  I would really like to find a way to eliminate having 13 versions of the API docs to maintain, though.

It should be possible to maintain the localized menu, header, and footer on the API doc pages, so the only real problem I think is that links to non-API docs (i.e. "guide" topics) would always go to the English version.

One possible way to mitigate that would be to have some kind of UI widget (tab/dropdown/whatever) on EVERY page that would enable the user to easily switch language .  That way, if you're in an API topic and then click on a link to a guide topic (that shows up in en), you could easily navigate to the translation of your choice.  What do you think?

crandmck:
@hacksparrow This takes on a bit more urgency in light of [this development](https://github.com/nodejs/TSC/pull/39#issuecomment-179356711):

> In addition to the core strongloop/express repo, IBM/StrongLoop will be contributing the API documentation to this effort, most likely by moving that documentation into the main express repository for the time being while we figure out the domain question separately.

We need to figure out how we can use .md docs in a different repo (https://github.com/strongloop/express or wherever it lands in the future).  And I don't think we want to move 13 copies of the API docs there.

jasnell:
[...]

PR #555, opened 2016-01-14
Broken Links

jordonias:


crandmck:
Thanks for the PR @jordonias. 

Some of these fixes are "no-brainers" but we need to look at this in light of #541 and how we intend to address that.  In short, we won't ever have professional translation of the API docs (for a variety of reasons), but we want to allow for potential community translations (as we have now for [Portuguese](http://expressjs.com/pt-br/4x/api.html)); the same goes for the "Resources > Books and Blogs" and "Resources > Applications" pages and old (v2 and v3) API docs, although no one would ever translate those I suppose.

In any case, we certainly don't ever want a broken link, so thanks for fixing those.

@hacksparrow PTAL when you have a chance, as we discussed.

crandmck:
Actually, I'll go ahead and review this.  I hate broken links.

crandmck:
Now that I look at this in detail, none of these changes are for links to the 4.1 API docs, so I think we can safely merge this.  Thanks @jordonias

Issue #542, opened 2016-01-07
Remove the "Applications" page


crandmck:
http://expressjs.com/en/resources/applications.html is basically a hold-over from the old site.  I don't think it provides a great deal of value and I think we should consider removing it for the following reasons:
- The list of sites/apps is naturally limited, and it's not clear why these are listed vs. others.  Presumably they knew someone or asked to be on the list.  Why aren't some of the more well-known Node users (Uber, WalMart, PayPal, LinkedIn, etc.) listed?
- It's not technical documentation, but more "marketing" for Express.  Express is well-known enough now that this is not as important as it was a few years ago.  IMO this site should focus on technical doc which is its primary purpose.
- It's easy enough to find this info with Google or elsewhere.
- Maintenance -- We don't have any way of ensuring that all the sites/apps listed still use Express.  They did at some point, but we'd have to spend some time to find out if they still do.  Also, if they ever move off Express, we have no way of knowing to remove them.

hacksparrow:
+1

crandmck:
@dougwilson, @ritch, anyone else -- thoughts?

crandmck:
Unless anyone speaks up with a compelling argument, I'm going to remove this page....

jordonias:
I feel it would be better to show a list of companies that use express, rather than a list of some applications I have little interest in. Not sure if this exists somewhere already. If it does, it should be on the homepage.

crandmck:
A reasonable idea.  I'll see if I can track down a reliable source for such information.  @jordonias Do you have a specific suggestion?

The other issue could be whether companies are OK with being listed on such a page.   Not sure of the implications there....  Will look into that.

jordonias:
@crandmck  No specific suggestion. I've only heard of companies using express, can't be 100% sure of any of them myself. It may have to be opt-in.

crandmck:
I'm going to take a two-step approach:
1. Remove the existing `resources/applications` page (for all langs)
2. Open an issue to add an opt-in "Who's using Express" page.  Once we have a few companies listed there, then we can add that back in.

PR #556, opened 2016-01-15
Remove Applications page

crandmck:
Per #542


crandmck:
Note this PR also removes the spurious file `_includes/header/header-zh.html`.  There is no longer a `zh` lang directory, but rather `zh-tw` and `zh-cn`.

Note also that I did not remove the [images/apps](https://github.com/strongloop/expressjs.com/tree/gh-pages/images/apps) directory for the time being.

PR #559, opened 2016-01-18
Fix missing semicolons

martinerko:
In addition to #558 I executed bunch of grep and sed commands to fix missing semicolons in many other places. 
This PR consists from multiple commits to make it more clear and easier to merge it.
It means one commit with all the fixes per each language translation.


hacksparrow:
Thanks @martinerko

PR #565, opened 2016-01-29
node foundation proposed express WG governance

dougwilson:
Per the proposal to add express to the Node.js Foundation incubator (nodejs/TSC#39) an Express Documentation WG under the Node.js TSC would be chartered to oversee the nodejs/expressjs.com github repository. Per the TSC guidelines, every WG needs a minimum set of governance policies in place that describe how the WG will operate. This adds an initial draft that is based directly on the base policy templates provided by the TSC.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,5 @@
+### Express Documentation WG Code of Conduct
+
+The [Node.js Code of Conduct][] applies to this WG.
+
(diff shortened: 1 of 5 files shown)


crandmck:
Can this be closed now?  We have http://expressjs.com/en/resources/contributing.html on the site that explains governance for all the repos in expressjs org (including this one). 

Also, as soon as the TC gives final approval, we should be able to put this repo under CCSA license per #634, which should clear up any other questions around licensing, contribs, etc.

dougwilson:
No longer necessary.

PR #568, opened 2016-02-04
Single API doc source

hacksparrow:
The **en** API doc will be used as the source of all language docs.


crandmck:
LGTM!  
Next step is to add some kind of UI widget to guide (non-API) docs to make it easier for readers to switch to another lang.  This will be useful for non-en readers who click on a link in API doc to another page.... Will open an issue for that.

Issue #574, opened 2016-02-11
Work to move API doc source to expressjs/express repo


crandmck:
Per https://github.com/expressjs/express/issues/2887 we're going to move the API doc source from this repo to https://github.com/expressjs/express/.  

This issue will track the work to be done in this repo to make that happen.  

In addition to actually moving the .md files, we need to add a script here so we don't have to maintain two copies of the API doc.

Before we do that, we should convert ``~~~js ... ~~~`
back to
`{% highlight js %} ... {% endhighlight %}`
So the doc source files display better in GH.

Then we need to add a script here to pull down .md files from `expressjs/express/doc` to here.  Not sure if its' possible to add some kind of hook to make that happen automatically whenever one of the API doc source files is changed in that repo?

This script should also convert 
``~~~js ... ~~~`
to
`{% highlight js %} ... {% endhighlight %}`
So we get syntax highlighting in code blocks.

hacksparrow:
@crandmck AFAIK, we'll need something online if we want to use git hooks.

crandmck:
> we'll need something online if we want to use git hooks.

OK, for now it will just have to be a manual process, I guess... unless anyone has any suggestions.

My plan is to complete this task this week.  

Once I make the above-noted edits to the API .md files and check them in to the `express` repo, I'll remove them from this repo and add a script at the top level that looks something like this.  I tested this, and it works, but I'm certainly open to suggestions if there is a better way:

```
git clone https://github.com/expressjs/express.git
mkdir _includes/api/en/4x
cp ./express/doc/*.md _includes/api/en/4x
sed -i.bak 's/~~~js/{% highlight js %}/g' _includes/api/en/4x/*.md
sed -i.bak 's/~~~/{% endhighlight %}/g' _includes/api/en/4x/*.md
rm _includes/api/en/4x/*.bak
```

crandmck:
To clarify, we don't actually want to remove the `_includes/en/api/4x/*.md` files from this repo.  We just want to copy them there (see https://github.com/expressjs/express/pull/2892), then make sure that any time someone makes a change there it gets copied here.

So, all this actually requires is adding a script to sync the files from `expressjs/express`.

crandmck:
Current state at this point is that API docs are in https://github.com/expressjs/express-api-docs and we need to sort out how these will be integrated with the API docs here.

crandmck:
[...]

PR #582, opened 2016-02-18
Add script to copy api files

crandmck:
For #574


crandmck:
I think we can close this since for the time being, since it appears the docs will live in `expressjs/express-api-docs`.

PR #583, opened 2016-02-19
Update routing.md

zakizak2016:
This route path will match `abcd` and `acd`.

its better than 
This route path will match `acd` and `acd`.

Make reade have better logic

### DIFF
--- a/en/guide/routing.md
+++ b/en/guide/routing.md
@@ -103,7 +103,7 @@ app.get('/random.text', function (req, res) {
 
 Here are some examples of route paths based on string patterns.
 
-This route path will match `acd` and `abcd`.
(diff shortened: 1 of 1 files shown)


crandmck:
Currently it says

> This route path will match `acd` and `abcd`.

And you have just reversed the order so it says

> This route path will match `abcd` and `acd`.

What is the point of that?  The two say exactly the same thing.  I don't see any difference in readability.

crandmck:
@zakizak2016 As stated above, this PR doesn't seem to make much sense.  Since you haven't responded --- I'm going to close. If you can make a good argument for it, then we can reconsider, but closing this until then.

PR #588, opened 2016-02-25
fix github button

hypebeast:
- use expressjs user instead of strongloop


crandmck:
Thanks for catching that, @hypebeast !

PR #595, opened 2016-03-03
Makes it more clear that the user is installing express-generator.

gh0st:


crandmck:
Thanks, that is definitely a good clarification / improvement!

PR #615, opened 2016-04-13
update about (not) removing uncaughtException event

martinerko:
I think the documentation is a little bit outdated about removing uncaughtException from Node.js core.

They decided to keep it as is and closed the proposal about removing it.
https://github.com/nodejs/node-v0.x-archive/issues/2582#issuecomment-9971225

The documentation was updated as well.
https://github.com/nodejs/node-v0.x-archive/commit/a2fd657b10c7609e1776004156a8cd3016e2c6c8

So we should too :)


crandmck:
Thanks for the update!  LGTM.

PR #623, opened 2016-04-21
Restored router.param callback "name" parameter

Jokero:
It was deleted from previous version of documentation

### DIFF
--- a/_includes/api/en/4x/router-param.md
+++ b/_includes/api/en/4x/router-param.md
@@ -6,6 +6,7 @@ Adds callback triggers to route parameters, where `name` is the name of the para
 - `res`, the response object
 - `next`, indicating the next middleware function
 - The value of the `name` parameter
+- The name of the parameter
 
 Although `name` is technically optional, using this method without it is deprecated starting with Express v4.11.0 (see below).
 


Jokero:
@crandmck Be careful, please, when edit documentation

crandmck:
Thanks for the PR @Jokero.
I'm trying to understand what the fourth parameter to the callback function actually is....

From the examples in http://expressjs.com/en/4x/api.html#router.param, it seems that it is in fact the _value_ of the `name` parameter, not the _name_ of the `name` parameter, which is why I made that change.  However, I'm not entirely sure.

In the first example... 

```
router.param('user', function(req, res, next, id) { ...
```

Here does  `id` have the value of `user` parameter?  Or something else?  Is it meant to be the name of the parameter, (here, `user`)?   That is what your change says.

In any case, I think this bears clarification.

Somewhat unrelated, the section that begins

> The behavior of the router.param(name, callback) method can be altered entirely by passing only a function ...

indicates that in fact, the `name` parameter is optional.  We should modify the heading/signature to indicate this, i.e. I believe the section heading should actually be:

#### router.param([name, ]callback)

Right?

dougwilson:
The actual change looks right. You can see that in Express 4.13.4, the callback is called with 5 arguments (https://github.com/expressjs/express/blob/4.13.4/lib/router/index.js#L404), where the 5th argument is the name of the parameter.

As for the `router.param(fn)` signature, this is an entirely different beast. Calling the method `param([name, ]callback)`, where it looks like `name` is optional is not very correctly, as the two different signatures operate in very different ways (the functions are not called with the same arguments or even in the same life cycle).

hacksparrow:
@crandmck as Doug mentioned `router.param(name, callback)` and `router.param(callback)` are totally different things. Also, `router.param(callback)` is deprecated now.

crandmck:
OK, thanks for clarifying.  I obviously didn't realize that the cb took 5 arguments, not 4, so that makes sense now.

However, all the examples show the callback with 4 or fewer arguments (which was partly the source of my confusion), so I think we need to clarify which are optional.  The examples have cb with at least `req` and `res` args, so are those required and the rest optional?  @hacksparrow @dougwilson

hacksparrow:
Unless the user wants to terminate the connection with a response, `next` is not optional.

crandmck:
@hacksparrow Yeah, that makes sense.  So, perhaps I should note that information for this example:

So, then are the final two args optional?  i.e. value of the name parameter and the name of the parameter ... ?

hacksparrow:
Yes, they are optional.

crandmck:
I clarified which cb args are required/optional in bd71fc13588aa3f6ce3ad135e986ba6736459063.

Jokero:
@crandmck I think it's incorrect to call callback arguments as required or optional. They all will be passed to the callback. Required/optional make sense then you want to call some function yourself:

``` js
arr.forEach(callback[, thisArg])
```

Here `thisArg` is optional. If you want you can specify this value. But it's not the same as callback arguments where you can't change anything, you just use or don't use them

crandmck:
@Jokero Maybe you can clarify for me.  I appreciate your patience.

If I understand what you're saying: the callback args are available within the callback function.  Obviously, if the callback signature can handle them, then the system passes the args to the callback.    But my question is do you _have to_ provide them when you define the callback function that you pass to `router.param` ?

We state that the callback takes 5 args: req, res, next, value of the name parameter, the name of the (name) parameter.  But then we have the example where the cb is provided with 4 args:

```
router.param('user', function(req, res, next, id) {
  ...
})
```

Looking around, I see another example where the cb is only provided with 3 args:

```
router.param('postId', function(request, response, next) {
  ...
})
```

So, unless these examples are wrong, you are not required to provide the last two arguments in the callback to `router.param()`.   Are you _required_ to provide them in your definition of the callback function in the call to `router.param`?    But it seems to me that you are not _required_ to define your callback with those arguments.

I just want to tell people how to write their code, and whether the MUST provide _all_ these args to the callback function to `router.param` or not.  

If you are always required to provide all 5 arguments when you define the callback function for `router.param`, then we should say that and change the example.  Otherwise, I think its worthwhile to note which you must always provide (i.e. are required) and which are optional.  If that's unclear, then perhaps I should add a note to clarify what we mean.

dougwilson:
We have to decide how much we expect our users to understand how JavaScript works.

crandmck:
It's a good point @dougwilson.  I would say we can assume basic JS knowledge, but we should focus on what's the "best practice".  So maybe I should just remove all the stuff I added about optional/required, and just give a best practice recommendation?  

Any suggestions along those lines?  

Is best practice to always provide all 5 arguments (except for `next`, which you omit to terminate the connection with a response)?  

Or just include the arguments that you are going to use?

dougwilson:
Is best practice to always provide all 5 arguments (except for  next , which you omit to terminate the connection with a response)? Or just include the arguments that you are going to use?

Yea, that's the biggest problem here. Documentation should call out all the arguments for completeness, but the best practice is the opposite: only specify the ones you need.

crandmck:
OK, for the time being I simply removed the required/optional notes from the callback parameters.
Thanks for noting @Jokero

Issue #597, opened 2016-03-07
Import changes to fix code blocks, etc.


crandmck:
Import the changes made in https://github.com/expressjs/express-api-docs/pull/3.  This allows us to use backticks for code blocks, and some other improvements.

crandmck:
This includes the changes above and modiying all the code blocks to use backticks. I also added the `Gemfile`, updated `config.yml` as in https://github.com/expressjs/express-api-docs/pull/3, as well as adding to it the line 

```
hard_wrap: false
```

which seems to be required to make text wrap properly.

However, in this branch there is an issue with the 4.x API doc page (http://localhost:4000/en/4x/api.html when running locally). It seems that `prism.js` is throwing an error, which really screws up the layout for that page (only)....  but I haven't been able to get to the bottom of it.

@hacksparrow Can you please have a look and see if you can sort out what's going on?

Also, I would appreciate any other eyes. Merging this branch is going to be a huge PR because it touches almost every file.

crandmck:
I figured out the problem with the 4.x API page and committed fix to the branch.

crandmck:
Fixed with https://github.com/expressjs/expressjs.com/pull/628 for English content.  
I did not do the same for translated content, so the PR wasn't as large as I originally thought.

PR #628, opened 2016-04-25
Replace tags with backticks, and other changes

crandmck:
This includes all the changes noted in #597.  I decided to make the change to code blocks only for English pages.  I left all translated (non-en) pages "as is" with `<pre><code> ...` tags instead of backticks.


hacksparrow:
Looks good. Has GitHub fixed the backtick issue?

hacksparrow:
I just checked. The backtick behavior is still in the "broken" state. If this PR is merged, we will lose the current code highlighting. We will need additional work to save the code highlighting.

crandmck:
@hacksparrow Obviously we don't want to do this if it breaks code highlighting.

I was told that it was fixed by the GitHub employee who opened the PR with these very changes against `express-api-docs` quite some time ago.  Unfortunately, I deleted that repo (since it was redundant) so I don't have the specific issue/comment.  

I thought that the changes to `_config.yml` were sufficient to fix it.  I'll investigate further.
BTW, how did you test it?

hacksparrow:
What's there in `_config.yml` that it will solve the problem?

I tested it on http://hacksparrow.github.io/git-experiments/. Look at the source code, GitHub tries to do it's own highlighting using Rouge, completely breaking how we expect the code to be.

crandmck:
> What's there in _config.yml that it will solve the problem?

I believe it was this:

```
syntax_highlighter_opts:
    disable : true
```

Should I try to create an demo repo such as yours to ensure that it will work?

hacksparrow:
Ok, I will take a look at it tomorrow, and merge this, if everything looks fine.

crandmck:
OK, thanks!

hacksparrow:
```
syntax_highlighter_opts:
    disable : true
```

make no difference. I wonder what I correct directive is. Tried searching "jekyll _config.yml disable syntax highlight" and the variants, no success.

crandmck:
@hacksparrow I created a test site, and everything (including syntax highlighting) seems to be working OK: http://crandmck.github.io/

Source repo: https://github.com/crandmck/crandmck.github.io.

NOTE: This test site does not include the translations, just to make it a lot smaller :-)

NOTE2: I'll remove this test site content once we've merged this PR and resolved this issue.  It's just temporary for testing.

Perhaps you missed something...?  Did you use the entire `_config.yml` file:

```
# Site settings

# Build settings
markdown: kramdown

gems:
  - jekyll-redirect-from

kramdown:
  input: GFM
  hard_wrap: false
  syntax_highlighter_opts:
    disable : true
```

hacksparrow:
I had `syntax_highlighter_opts` at the root level, hence it had no affect.

Confirmed to be all good, merging this.

crandmck:
Thanks @hacksparrow !
Sorry I wasn't clearer about where `syntax_highlighter_opts` went.

PR #629, opened 2016-04-26
remove unsupported methods

hacksparrow:
Updated the doc as per - https://github.com/expressjs/express/issues/2966


crandmck:
Thanks.  It's good to get that fixed.

PR #631, opened 2016-05-03
Fix HTML entities in generator options

SlashmanX:
Possibly a byproduct of the change to backticks (#628)

screen shot 2016-05-03 at 10 14 12


crandmck:
Thanks @SlashmanX, yes I think it did slip in when we changed to backticks.

PR #638, opened 2016-05-14
Remove incorrect argument information for res.json

NicholasAntonov:
The res.json documentation incorrectly stated that it only took an object or array as an argument, which isn't true as internally JSON.stringify is used, and being able to send primitive types as a json response is a very desirable feature.

Also it incorrectly stated that it was almost identical to `res.send`, which is incorrect in the case of numbers and strings and has been reworded.

### DIFF
--- a/_includes/api/en/4x/res-json.md
+++ b/_includes/api/en/4x/res-json.md
@@ -1,6 +1,6 @@
 <h3 id='res.json'>res.json([body])</h3>
 
-Sends a JSON response. This method is identical to `res.send()` with an object or array as the parameter.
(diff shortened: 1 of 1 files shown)


crandmck:
@hacksparrow @dougwilson Can you verify that this change is correct?

dougwilson:
I won't comment on the wording, but the technical changes are correct.

crandmck:
Thanks, I merged and clarified the wording in ad73b8dbe2ef4ee2654b555181459b7cd6aa6824.
Thanks for the contribution @NicholasAntonov.

PR #643, opened 2016-05-19
Note that app[method] names are lowercase

jmm:
It's mentioned in the [API docs](http://expressjs.com/en/4x/api.html#app.METHOD), but not here. It lacks clarity because the HTTP method names are uppercase, and these docs use a convention of representing placeholder values in uppercase (e.g. `app.METHOD`) .

And link to fragment about request methods.


crandmck:
Good clarification, thanks!

PR #644, opened 2016-05-19
 Update contrib links: strongloop/ => expressjs/

jmm:
I also changed:

``` diff
- Clone the Express repository
+ Clone the `expressjs.com` repository
```

Since "the Express repository" to me sounds like the repo for the framework.


crandmck:
Thanks!  We keep finding instances of these that we overlooked...

PR #645, opened 2016-05-20
Add more information to res.render

dougwilson:
This adds more information to the `res.render` API documentation, mainly documenting what the `view` argument actually is and does, and calls out that the argument is to be trusted input, rather than user-controlled input.


crandmck:
Thanks @dougwilson, this is good information to add.  I'm going to make some minor edits to the wording before merging.

PR #665, opened 2016-07-01
Adding KeystoneJS to Frameworks

JedWatson:
I've also ordered the frameworks alphabetically, this seems to make sense and avoids favouritism in the ordering.


crandmck:
Thanks.  I meant to keep them in alpha order, but they got out of order due to numerous PRs.

Issue #671, opened 2016-07-08
Add general info about TC meetings


crandmck:
As noted in https://github.com/expressjs/discussions/issues/23, we should have some some general info about TC meetings somewhere on the site, e.g. in http://expressjs.com/en/resources/community.html and/or http://expressjs.com/en/resources/contributing.html as well as links to the meeting announcements and minutes.

PR #672, opened 2016-07-13
Add info about TC, other misc cleanup

crandmck:


crandmck:
This is per #671.  It's very straightforward, so I'm going to merge it myself.

PR #674, opened 2016-07-14
Found an awkward Korean word

DaleSeo:
We Koreans don't' use the Korean word, "시간소인" to translate the English word, "timestamp".
It's more natural and acceptable to native Koreans to translate it to "타임스탬프".
Please refer to the following links.

https://zkorean.com/dictionary/search_results?word=timestamp
http://krdic.naver.com/detail.nhn?docid=39084100


crandmck:
Thank you!

PR #677, opened 2016-07-20
Mention signedCookies when reading about cookies

kevinsimper:
I could not figure out why my cookies was not in the req.cookies, but it was because I had signed it. I went to cookie-parser, but it was not mention either in that README. I created a Pull request, but I think the info could also be needed here to avoid confusion. https://github.com/expressjs/cookie-parser/pull/29


crandmck:
LGTM.  Thanks!

PR #685, opened 2016-08-11
Correct the sentence to say 'Now install Express in the myapp directory'

vedala:
Since we currently located in the myapp directory, running command `npm install express --save` will install express in the current directory.


crandmck:
Good catch, thanks!

PR #691, opened 2016-08-22
Corrected variable name in Couchbase sample code

mgroves:
Sorry, I mean to use 'bucket' instead of 'db' in the sample snippet.


crandmck:
Oops!  Thanks for fixing...

PR #692, opened 2016-08-23
Replacing code snippet which was deprecated.

vintesh:


crandmck:
@hacksparrow Would you please review this?  It's not immediately apparent to me whether it's correct.

hacksparrow:
Good update. Thanks @vintesh!

PR #694, opened 2016-08-25
Add basic print media adjustments

dougwilson:
This adds the most basic print media CSS needed to keep web browsers that understand fixed position from positioning the header at the top of every page, covering content.


hacksparrow:
Thanks @dougwilson

crandmck:
Thanks @dougwilson !

PR #697, opened 2016-08-28
Add missing path argument types

aks-:
- Adds missing path arguments type for app.put, app.post, app.get.
- documents path type for app.METHOD and app.all
- addresses issue https://github.com/expressjs/expressjs.com/issues/630

Issue #630, opened 2016-04-27
Document path array signature for app.METHOD


crandmck:
From https://github.com/expressjs/express/issues/2978:

`app.METHOD(path_array, callback [, callback ...])` signature is not documented.

Per @dougwilson:

> ... the `path` argument to `app.METHOD` can not only be a string path, but also an array of string paths, a regular expression, an array of regular expressions, or an array of mixed types.

So, we need to add that.  

We probably should add a similar note to to the [routing guide](http://expressjs.com/en/guide/routing.html) (and ideally a short example of using a path array as well) and perhaps also reference topics for `app.get`, `app.post`, and `app.put` for completeness and consistency.

BTW, is this also true for `app.all`?  It would make sense, but I'd like to confirm before adding anything there?

dougwilson:
> BTW, is this also true for  app.all ?

Yep, it also has the same ability. In fact, `app.use` also has this ability if it's not already in the documentation (did not actually look for `.use`).

crandmck:
Thanks for confirming that, @dougwilson.

We do document path arrays for `app.use`, specifically it says:

> ... path can be a string representing a path, a path pattern, a regular expression to match paths, or an array of combinations thereof.

Perhaps that could be worded better, but at least it notes that you can use an array.

dougwilson:
Cool. Yea, I would say at lest a quick fix would be to just copy that wording to the `app.METHOD` and `app.all` sections :)

hacksparrow:
@crandmck it is not just the `path` parameter details we are missing in `app.METHOD` and `app.all`, we are missing the details of the `callback` parameter, too. Ideally it would be good to have the details of both the tables for `app.METHOD` and `app.all` also, or have some kind of "smart reference".

crandmck:
Good point, @hacksparrow !  

Can you provide that information?  Are the cb params just `req`, `res`, and `next` or is there more to it?  

And, not sure what you mean by:

> ... it would be good to have the details of both the tables for app.METHOD and app.all also, or have some kind of "smart reference".

Can you pls clarify?

hacksparrow:
On http://expressjs.com/en/4x/api.html#app.use you will see two tables listing the possible types and combos for the `path` and `callback` parameters, the same table applies to `app.METHOD` and `app.all` as well.

ryhinchey:
Closing as https://github.com/expressjs/expressjs.com/pull/701 was merged

PR #701, opened 2016-09-01
Clarify arguments to routing methods

crandmck:
This addresses #630, related issues, and some other minor issues.

First, I have to apologize for leaving #630 unaddressed for so long.  Especially when the resolution was pretty straightforward and the necessary info was in the issue.  I guess it just fell off my radar.

Second, I want to thank @aks- for putting it back on my radar by opening #697.  Unfortunately, I think this PR will obviate that one, although if you still want to add some more examples of (realistic) regexes, you could probably do that.

Anyway, here's what's in this PR:
- Consolidate description of path and callback arguments into an "include" and then duplicate it everywhere appropriate (app.METHOD, app.get, app.put, app.post, and app.all) via `{% include api/en/4x/routing-args.html %}`  I also included a paragraph about providing multiple callback functions that was duplicated in a number of places.
- Added subsection headers for the examples in app.use for "Path examples" and "Middleware callback function examples" and linked to these from all the applicable sections 
- (essentially unrelated) Move note about `req.orignalUrl` from "app.use" section (where it was a bit out of place) to "req.orginalUrl" section.

One minor question: Since the method signatures (in the headings) have `callback [, callback ...])
`, I wonder if we need to specify "A series of middleware functions (separated by commas)." since that's implied by the signature? Maybe it's clear either way? Or perhaps we should just have `callback(s)` in the signature?  

Finally, I _think_ the description of the `path` & `callback` arguments is also applicable to `router.all`, `router.METHOD`, and `router.use`, but I wasn't 100% sure.  If so, I can land that in a separate PR, as this one has enough in it already.  Having the pertinent info in an include will make it very simple.

### DIFF
--- a/_includes/api/en/4x/app-METHOD.md
+++ b/_includes/api/en/4x/app-METHOD.md
@@ -2,9 +2,11 @@
 
 Routes an HTTP request, where METHOD is the HTTP method of the request, such as GET,
 PUT, POST, and so on, in lowercase. Thus, the actual methods are `app.get()`,
-`app.post()`, `app.put()`, and so on.  See below for the complete list.
+`app.post()`, `app.put()`, and so on.  See [Routing methods](#routing-methods) below for the complete list.
 
-For more information, see the [routing guide](/guide/routing.html).
+{% include api/en/4x/routing-args.html %}
+
+#### Routing methods
 
 Express supports the following routing methods corresponding to the HTTP methods of the same names:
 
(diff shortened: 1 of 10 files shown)


dougwilson:
Generally LGTM. I know you asked questions as well, but just wanted to get that feedback to you. Also, why is the arguments file HTML vs MD?

crandmck:
Thanks @dougwilson The questions are not critical at this time... Can always just iterate to address them later, but I'll wait til tomorrow to merge, in case there are any other comments.

> why is the arguments file HTML vs MD?

Table formatting in markdown is extremely limited, and I wanted to be able to do stuff like have a bullet list in table cells, specify column width, etc, which you can't easily do in markdown--at least, it's very hacky and error-prone to do so... It just ends up being easier to do it in HTML.

Issue #698, opened 2016-08-31
scroll click of button gives different behaviour


ranjan-2209:
I have hovered my mouse on html page as shown in snapshot and clicked scroll of mouse. It opens the page in same tab as well as other. This scroll click should have no effect on the current page.
![1](https://cloud.githubusercontent.com/assets/16834561/18130235/cf56785e-6fab-11e6-9062-8f5a0fc60377.png)

crandmck:
What browser and OS are you using?

ranjan-2209:
Windows 10 latest & Chrome beta latest. You also must be getting it as it is independent of OS and browser.

blakeembrey:
Thanks for the report. It seems to be caused because the page has _two_ handlers. It's a link, so the regular behaviour applies. But, for some reason there's also a manual redirect in JavaScript. @crandmck any idea why that would be there?

![image](https://cloud.githubusercontent.com/assets/1088987/18153188/2c85caa0-703f-11e6-9ee4-b010341b9332.png)

![image](https://cloud.githubusercontent.com/assets/1088987/18153197/36cb7c08-703f-11e6-983e-644dc8d6f8d8.png)

crandmck:
Yes, I do see it now (initially, apparently I was doing something different and didn't see it).  And I hadn't noticed it before.

Anyway @hacksparrow would have more insight, since he implemented the menu system.

dougwilson:
Oddly enough, it doesn't happen using Microsoft Edge. Anyhow, I understand that that JavaScript is trying to enlarge the click area, by making a click anywhere on the `<li>` trigger a navigation. If the margin is removed from the `<a>` such that is consumes the entire area of the enclosing `<li>` then that click event would definitely be redundant. I assume the intention is such that if the hovering shows the highlight, it ensures that the exact same mouse position is also clickable.

dougwilson:
I guess I assumed it was being more clever. Looking more into it, it simply seems redundant. PR #702

hacksparrow:
This behavior is a side-effect of increasing the touch area on mobile devices. It has to be made smarter, of course. I am working on it.

hacksparrow:
https://github.com/expressjs/expressjs.com/pull/706 limits the behavior of `li`s to mobile devices.

PR #702, opened 2016-09-02
Remove menu li click events

dougwilson:
The removes the menu li click events, which seem unnecessary and clicking directly on the anchor tag would be OK.

Proposed fix for #698

/cc @crandmck @hacksparrow

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -188,10 +188,6 @@ $(function(){
         $(this).addClass('active-mobile-menu')
       }
     }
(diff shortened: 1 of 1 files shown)


hacksparrow:
This breaks the "enhanced" menu on mobile devices. https://github.com/expressjs/expressjs.com/pull/706 is suggested, instead.

crandmck:
Closed in favor of #706

Issue #699, opened 2016-09-01
redirection issue


ranjan-2209:
![2](https://cloud.githubusercontent.com/assets/16834561/18162823/a88f1a1e-7055-11e6-96dd-d78d4dfdf842.png)

crandmck:
The #restart anchor was missing on the heading.  Added by 913e43791fb5b128b45c52a4dc5dc46af93a10d3 so the link to `Ensure your app automatically restarts` should work.

BUT--this link, the other two you note above, and in fact _any_ link to an anchor actually doesn't work quite  right, due to the menu bar "overlay" at the top of the page that obscures the top 30px or so of the content--you have to scroll the page up a bit to see the actual heading to which the link goes. 

This is a general problem with the site that I'd like to fix.  It shouldn't be too hard... @hacksparrow could you PTAL?

dougwilson:
The root issue is that h5 does not have the `:target` style. Also, the existing target style is ugly, creating a gap above the last clicked item.

hacksparrow:
Taking a look at it. I had implemented a solution for the problem on the API page, will implement a uniform solution with Doug's PR on mind.

hacksparrow:
https://github.com/expressjs/expressjs.com/pull/705 is the uniform implementation.

dougwilson:
Nice, @hacksparrow !

hacksparrow:
https://github.com/expressjs/expressjs.com/pull/705 fixed it.

PR #703, opened 2016-09-02
Add :target style to all elements in the API doc

dougwilson:
This adds the target padding to all the elements in the API doc area, as well as repositioning it back up to remove the created whitespace gap.

Proposed fix for #699 

/cc @crandmck @hacksparrow


crandmck:
Awesome!  LGTM.

PR #706, opened 2016-09-02
Make menu links "clickable" only on mobile devices

hacksparrow:


dougwilson:
The `'ontouchstart' in document.documentElement` is still returning true on my laptop. Probably because it has a touch screen, like many computers these days. Any other way to detect mobile?

hacksparrow:
Oh, should have thought about that! There are many ways to detect mobile, each with its pros and cons. Let me look for the best.

hacksparrow:
@dougwilson, I have added a small lib to detect mobile devices, please take a look now.

dougwilson:
LGTM

crandmck:
Thanks @hacksparrow

PR #712, opened 2016-09-19
Update package.json reporting jade being there

rauno56:
I didn't update other packages because I don't think the versions matter that much.


crandmck:
Thanks!

Issue #662, opened 2016-06-28
Update doc when Pug is added to Express generator


crandmck:
When https://github.com/expressjs/generator/pull/117 is merged, update http://expressjs.com/en/starter/generator.html accordingly (should be a trivial change).  
Also, perhaps add a note about Pug being the new name for Jade.

dougwilson:
Really, with each release of Express, we should plan to update the generator output, because there are already staged changes to the generator that will need that page updated :)

crandmck:
Good point @dougwilson.  I'll keep an eye on that repo for that reason.

dougwilson:
Hi @crandmck the generator has been updated now. As discussed, the default has not been changed from jade (yet) but there are new options, including one for pug. We'll want to update the generator docs to use an option, since it will give a warning without one.

crandmck:
Fixed by #726

Issue #609, opened 2016-03-23
Change references to "jade" to "pug"


crandmck:
Per https://github.com/expressjs/express/issues/2952, need to update http://expressjs.com/en/guide/using-template-engines.html (and search for any other instances of "jade").

Knighton910:
I'm doing a few updates to express guides, i can handle this, but it will be several days or so, do you wish to take care of this, or shall i get to it?

crandmck:
@lordKnighton Please do!  I've been meaning to get to this, but just haven't had a chance.  Thanks!!

Knighton910:
no worries

crandmck:
@lordKnighton Are you going to take another crack at this?  If not, then I'll try to do it.  No worries either way, just let me know.

Knighton910:
go for it, i got swapped. I can do it next week, if you prefer me to do it.

crandmck:
OK, PR updates the English docs, and I opened #618 to track task to update translated docs.  Also added #617 for when the generator is updated.

@lordKnighton (or anyone) pls take a quick look at the PR ^

Knighton910:
👍

crandmck:
Thanks!  
We'll address non-en docs as time permits, and generator docs when the generator is updated.

Knighton910:
+1

martinjoiner:
Hey guys.  I am a firs-timer here on this repo. I would like to contribute by replacing references to Jade with Pug in the "Express Application Generator" page http://expressjs.com/en/starter/generator.html  I recently got confused when followed this guide to bootstrap a project and then tried to find Jade documentation. Is it OK to go ahead and fork, edit and create a pull request?

crandmck:
Hi @martinjoiner thanks for offering to contribute.

This seems reasonable, and IIRC, the only reason we haven't made this change was because Pug was still in beta.  Although npm still says version if 2.0.0-beta11, it's replaced Jade now for quite a while, so I'm not sure if this is still an issue.

Even though the generator still supports Jade, I agree we should start nudging people toward Pug, since it's being maintained instead of Jade now (due to the naming issue).

However, I'll defer to @dougwilson on this.... Doug, should we wait for Pug non-beta release to make this sort of change?

dougwilson:
I don't see any reason any longer not to just update that generator page to use `--view=pug` if that reduces confusion 😄

martinjoiner:
[...]

PR #715, opened 2016-09-24
Change Jade to Pug in german docs

JonathanPrince:
Changes Jade to Pug in the german docs as outlined in #609 and #616 

Excludes changes to generator docs which are pending, see #662

Would be happy to update the other languages in the same manner, if so just let me know.


crandmck:
Danke schön!  
We would welcome corresponding changes for other languages.

PR #717, opened 2016-09-25
change Jade to Pug in multiple language guides

### LINKED ISSUE #618: Update translated docs for Jade -> Pug
Need to update all translated docs with same changes as in #616.

JonathanPrince:
change Jade to Pug in guides for remaining languages, resolves #618

### DIFF
--- a/es/guide/debugging.md
+++ b/es/guide/debugging.md
@@ -93,9 +93,9 @@ Cuando se realiza una solicitud a la aplicación, verá los registros especifica
   express:router serveStatic  : / +2ms
   express:router router  : / +2ms
   express:router dispatching GET / +1ms
-  express:view lookup "index.jade" +338ms
-  express:view stat "/projects/example/views/index.jade" +0ms
-  express:view render "/projects/example/views/index.jade" +1ms
+  express:view lookup "index.pug" +338ms
+  express:view stat "/projects/example/views/index.pug" +0ms
+  express:view render "/projects/example/views/index.pug" +1ms
 </code>
 </pre>
 
--- a/es/guide/migrating-4.md
+++ b/es/guide/migrating-4.md
@@ -320,7 +320,7 @@ var app = express();
 // all environments
 app.set('port', process.env.PORT || 3000);
 app.set('views', path.join(__dirname, 'views'));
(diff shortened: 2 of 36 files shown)


hacksparrow:
Thanks for the PR @JonathanPrince. However, we are not supporting Pug till it comes out of beta. Even then, there might be some changes in the default or no-default view engine. Just letting you know that this PR is going to sit here for while.

JonathanPrince:
Thanks for letting me know @hacksparrow. Just to avoid any confusion, similar changes are already merged for english #616 and german #715, let me know if/when I can help with anything.

hacksparrow:
I see! @crandmck should have the context of the merges. Let's wait for him to handle this as well.

crandmck:
Sorry, been slammed with other work...

Looking into this, I may have been too hasty in updating "Jade" to "Pug" everywhere... 
I somehow got the impression that the generator had been updated, but I see from https://github.com/expressjs/generator/pull/117 that this hasn't actually happened yet.
I probably assumed that when the PR with the changes was merged, they were also released to npm--which obviously is a dumb assumption :-(

Which means perhaps I shouldn't have merged those other PRs (#616 and #715 ) so quickly.  
_sigh_  If we can avoid backing those out, that would be ideal, but we do need to make sure that what we have is correct.

Until the generator is updated, it makes it difficult to document clearly--since people should start using Pug to make sure they're getting the "latest and greatest" version of Jade (Pug), but they would have to make a bunch of manual changes if they're using the generator.

I think most of the changes in http://expressjs.com/en/guide/using-template-engines.html are OK, even if the generator hasn't been updated.  EXCEPT, this is not actually true:

> The Express application generator uses Pug as its default

Anyway, let's wait to merge this PR open until generator 4.14 is released.  
I'll fix the above line in the English docs.  @hacksparrow If you see anything else that needs to change, please let me know.

sandizapar (approved):

sandizapar (approved):

crandmck:
This should have been landed a long time ago, since https://github.com/expressjs/generator/pull/117 has been in for over a year.  

Sorry for the delay.

PR #718, opened 2016-09-27
ES6 Hello World

DomVinyard:


hacksparrow:
In my opinion, it might be a little early to use ES6 syntax in the examples. People new to Node.js/JavaScript necessarily need not know ES6, but people who know ES6 already know what the ES6 equivalent of the existing code should look like.

/cc @expressjs/express-tc

DomVinyard:
OK, I fully defer. I have copy/pasted this snippet about half a dozen times today (don't ask!) and I thought it would've been quicker to convert yours once than mine 6 times. It's there when you need it.

LinusU:
Why is `app` declared with `let` and not `const`? Also, I think that we should keep the spacing, something like this:

``` js
const app = express()

app.get('/', (req, res) => {
  res.send('Hello World!')
})

app.listen(3000, () => {
  console.log('Example app listening on port 3000!')
})
```

hacksparrow:
@DomVinyard where did you have to copy/paste it? Is it an open community?

dougwilson:
As for ES6 examples, I think these days, especially since Node.js changed their docs, it's probably OK to transform it to ES6 _when it makes sense_, In this case, That said, I don't think in this case it makes sense. There are only two functions in the example and in both cases, I think the change to an arrow function is not great:
1. The GET route function is nice to be a full function, because literally except for this example, no real function is only to be just one statement. Using that one line arrow function seems like trying to optimize the example to the one statement, rather than being an example that people are expected to extend.
2. The listener one seems to just make a really long line.

If you change those two arrow functions to be blocks, at that point it doesn't seem like anything is really different from the original example, except a few characters less. But maybe that is worth it?

TL;DR I think ES6 in examples these days may be OK, especially since all the Node.js docs are in ES6. That said, this PR needs to add the vertical spacing back and put the function bodies back to blocks.

hacksparrow:
Let's discuss migrating the examples to ES6 in in the next TC call, and make it an all-or-none thing. Closing this for now.

Thanks for the suggestion, @DomVinyard. Your point is noted.

PR #720, opened 2016-10-06
CLI rename

davidgatti:
Change the command to generate the structure to the one that actually works.

### DIFF
--- a/en/starter/generator.md
+++ b/en/starter/generator.md
@@ -10,7 +10,7 @@ redirect_from: "/starter/generator.html"
 
 Use the application generator tool, `express-generator`, to quickly create an application skeleton.
 
-Install the `express` command line tool with the following example:
+The `express-generator` package installs the `express` command-line tool. Use the following command to do so:
 
 ```sh
 $ npm install express-generator -g
--- a/en/starter/generator.md
+++ b/en/starter/generator.md
@@ -10,7 +10,7 @@ redirect_from: "/starter/generator.html"
 
 Use the application generator tool, `express-generator`, to quickly create an application skeleton.
 
-Install `express` with the following command:
+Install the `express` command line tool with the following example:
 
 ```sh
 $ npm install express-generator -g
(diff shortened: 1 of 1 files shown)


crandmck:
Thanks for the PR.  

Not to make a mountain out of a molehill, but the problem is that the package name is `express-generator` and the command is `express`, so one could argue that either of the following is correct:
- "Install `express-generator` with the following command ..." (what we have now) 
- "Install `express`  with the following command ..." (what you propose)

To anyone familiar with Express, it's not confusing, but it could potentially be confusing to a newcomer.

Perhaps it would be most clear to say:

"The `express-generator` package installs the `express` command-line tool; install it by entering this command:
`$ npm install express-generator -g`

(Instead of "installs the `express` command-line tool" could also say "sets up..." or "creates..." to distinguish from an npm install).

Thoughts?

en/starter/generator.md
```diff
@@ -10,7 +10,7 @@ redirect_from: "/starter/generator.html"
 
 Use the application generator tool, `express-generator`, to quickly create an application skeleton.
 
-Install `express-generator` with the following command:
+Install `express` with the following command:
```
dougwilson: > Install `express` with the following command

Yes, definitely very poor wording here. It sounds like you're _installing express_, which to me, sounds like this module express. I understand that there are people (like me), who would think `express` means the module and other (yourself) who would think `express` means the command line program.

I'm not saying either is right or wrong, but as you can see, with both thinking those different things, I think it means the entire wording just needs to be completely redone to make the intent clear.

dougwilson (changes requested):

en/starter/generator.md
crandmck: @dougwilson What do you think of my suggestion above?

dougwilson: Seems fine.

davidgatti:
OK guys, check the following change. And let me know how you like it :)

crandmck:
Thanks @davidgatti this is close, but the wording is not quite right: "... with the following example" should be "... with the following command".

Also, I think it's helpful to preface the statement with something like this to clarify what's actually going on:

"The `express-generator` package installs the `express` command-line tool"

davidgatti:
What do you think now? :)

crandmck:
Works for me.  Thanks @davidgatti

PR #721, opened 2016-10-19
Fix an issue with start command

samunplugged:
It is not possible to run the app with the current command mentioned in this document. It needs to be updated. Hence proposing this file change.

### DIFF
--- a/en/starter/generator.md
+++ b/en/starter/generator.md
@@ -69,7 +69,7 @@ $ npm install
 On MacOS or Linux, run the app with this command:
 
 ```sh
-$ DEBUG=myapp:* npm start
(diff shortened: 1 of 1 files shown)


crandmck:
I just followed the instructions with no problem.  The app ran fine for me with 

```
DEBUG=myapp:* npm start
```

Also, the `express` generator displays instructions with the same exact command, so if there is any issue with it, we should also update the generator.  However, it seems to work fine.

@samunplugged Please provide more details about what happened for you, e.g. error messages, etc. What OS/version are you using?

samunplugged:
Ah, it works. Not sure what I did wrong the last time. Closing this. Thanks for response.

PR #724, opened 2016-10-21
listen signature with unix sockets support

magne4000:
See expressjs/express#3107

### DIFF
--- a/_includes/api/en/4x/app-listen.md
+++ b/_includes/api/en/4x/app-listen.md
@@ -1,6 +1,6 @@
 <h3 id='app.listen'>app.listen(path, [callback])</h3>
 
-Binds and listens for connections on the given path.
+Starts a UNIX socket and listens for connections on the given path.
 This method is identical to Node's [http.Server.listen()](https://nodejs.org/api/http.html#http_server_listen_path_callback).
 
 ```js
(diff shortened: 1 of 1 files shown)


dougwilson:
LGTM

crandmck:
I think the wording needs to clarify that this listens on a Unix socket, e.g. the corresponding [Node doc](https://nodejs.org/api/http.html#http_server_listen_path_callback) says:

> Start a UNIX socket server listening for connections on the given path.

So, some wording similar to that...

Also, the existing [app.listen doc](http://expressjs.com/en/4x/api.html#app.listen) says "...This method is identical to Node’s http.Server.listen()" with link to the Node doc.

@dougwilson I assume the same is true for this signature?  If so, then we we should also link to https://nodejs.org/api/http.html#http_server_listen_path_callback in this section.

dougwilson:
So I see the current wording is a copy and paste from the host + port section, simply changing out host and port to be path. @crandmck is really good with wording, so I think the ask here sounds good. Yes, this is the same as in http.Server. There is a link included that does point to the different docs, though the link's description is just http.Server.listen like the one below.

crandmck:
>  There is a link included that does point to the different docs ..

Oops, quite right! How could I have missed that?  I think I got distracted by the big outage that left me unable to access GitHub on and off all morning :-(

magne4000:
Updated

hacksparrow:
Thanks @magne4000

PR #726, opened 2016-10-31
Update generator documentation to 4.14.0

dougwilson:
This pull request simply updates all the generator documentation to reflect the changes in express-generator version 4.14.0. It does _not_ change the view engine away from jade, simply updates the commands to work the same way it does today. There is a much more comprehensive PR #717 that is changing the views from jade to pug. If that PR is rebased on top of these changes, the generator command can simply be updated from `--view=jade` to `--view=pug`.

### DIFF
--- a/de/starter/generator.md
+++ b/de/starter/generator.md
@@ -28,10 +28,12 @@ $ express -h
   Options:
 
     -h, --help          output usage information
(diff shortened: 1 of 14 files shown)


crandmck:
Thanks @dougwilson !

PR #736, opened 2016-11-20
Addition of Hydra-Express to list of frameworks

cjus:
Addition of Hydra-Express to the list of library / frameworks. Hydra-Express is a library for building microservices using Hydra-core and ExpressJS.  This is one of three Hydra projects which were open sourced at EmpireNode 2016.


crandmck:
LGTM, thanks!

PR #744, opened 2016-11-29
Fix grammar in res.sendFile documentation

OliverBalfour:


crandmck:
I don't think this will work because the string `'Sorry! You can't see that.'` is single-quoted and the apostrophe is also a single quote.  You need to either escape the apostrophe or enclose the string in double-quotes `"`.

hacksparrow:
@Tobsta can you make the change as suggested by @crandmck? I can land it, if you do.

OliverBalfour:
Just updated, sorry about that. Didn't notice at all.

hacksparrow:
Thanks @Tobsta

PR #747, opened 2016-12-05
Remove front page notice re NINA 2016

crandmck:
@hacksparrow I left the CSS files/styles in case we want to reuse them in the future.


dougwilson (approved):
LGTM

Issue #748, opened 2016-12-10
Document that GET handler also handles HEAD requests


syrnick:
The current docs don't mention that GET handlers must also accept HEAD requests:

https://github.com/expressjs/express/blob/master/lib/router/route.js#L65

This is quite unexpected and would cause bugs if something is done for GETs, but not for HEADs. E.g.
```
app.get('/', (req, res, next) => {
   assert(req.method === 'GET');
   ...
});
```
looks reasonable, but will throw for a HEAD request.

(Originally incorrectly posted to https://github.com/expressjs/express/issues/3149)

crandmck:
OK, if I understand what you're saying, we should add a note saying that `app.get()` and `router.get()` will handle HTTP HEAD requests as well as GET requests.  Is that right?

[According to the spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.4):
> ... the server MUST NOT return a message-body in the response.

So perhaps we should note that as well?

Ideally we would provide a short example of the right way to handle both GET and HEAD.  @dougwilson Is that just a matter of checking `req.method` and doing the appropriate thing for `GET` vs. `HEAD`?

dougwilson:
Express handles the details for the users. Users are expected to just do the GET action in their GET route and not look at the req.method. This is because per spec a HEAD is expected to perform the same action as a GET for the same path, but simply not return a body. Express (and Node.js as well) ensure that the body is dropped automatically.

syrnick:
It just needs a note like this: A GET handler MUST handle HEAD requests unless a HEAD handler was explicitly added. 

@dougwilson the spec says nothing about the "action" it just talks about the headers. E.g. if a GET request would do some work, warm up caches and load things into memory, HEAD request might do none of that. It could only check permissions and last modified time and be done.

I think it's a good idea to call GET handlers for HEAD requests. Developers just need a reminder that HEAD requests exist and they should expect them in GET handlers. If they call a function that is available only for the GET requests (e.g. provided by some middleware), they should check for request type first. 

The following will look crazy to many and they'd need to learn about https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.4 and  https://github.com/expressjs/express/blob/master/lib/router/route.js#L65 .

```
app.get('/', (req, res, next) => {
   if (req.method === 'GET') {
   } else {
   }
});
```

dougwilson:
[...]

PR #750, opened 2016-12-13
Add notes about automatic HEAD

dougwilson:
Proposed notices to add regarding the "automatic HEAD" feature of Express routing for #748

### DIFF
--- a/_includes/api/en/4x/app-METHOD.md
+++ b/_includes/api/en/4x/app-METHOD.md
@@ -53,6 +53,11 @@ However, the other methods listed above work in exactly the same way.
   `app['m-search']('/', function ...`.
 </div>
 
+<div class="doc-box doc-info" markdown="1">
+  The `app.get()` function is automatically called for `HEAD` method in addition to the `GET`
+  method if no `app.head()` was called for the path before `app.get()`.
+</div>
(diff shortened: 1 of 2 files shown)


syrnick:
Looks great! Can `HEAD` be a link to https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.4 ?

dougwilson:
Since RFC 2616 is obsolete and has been replaced, I would not link to that, personally.

syrnick:
sounds good

crandmck (approved):
LGTM, thanks!

renatoargh:
does any one know if there is a way to turn off this automatic head feature?

ryhinchey:
> does any one know if there is a way to turn off this automatic head feature?

@renatoargh There isn't a configuration to let you have get routes not respond to head requests.  That said, `get` routes should only respond to `head` requests if `app.head` was not called before `app.get`.  Here are the docs:

https://expressjs.com/en/api.html#app.METHOD

```
The app.get() function is automatically called for the HTTP HEAD method in addition to the GET method if app.head() was not called for the path before app.get().
```

Hope this is helpful!

PR #752, opened 2016-12-14
update lang yaml variable for pt-br

eddiemonge:


crandmck:
Yep, that's a mistake.

PR #763, opened 2017-01-10
Guide to overriding sugar methods

### LINKED ISSUE #757: Missing documentation for overriding sugar methods
This issue originated here: https://github.com/expressjs/express/issues/3151

Documentation that would allow us to override sugar methods is missing from the documentation on the site.

hacksparrow:
Guide explaining sugar methods and how to override them; resolves https://github.com/expressjs/expressjs.com/issues/757.

@dougwilson @crandmck @raijinsetsu

### DIFF
--- a/en/guide/overriding-express-api.md
+++ b/en/guide/overriding-express-api.md
@@ -1,0 +1,61 @@
+---
+layout: page
+title: Overriding the Express API
+menu: guide
+lang: en
+---
+<div id="page-doc" markdown="1">
+
+# Overriding the Express API
+
+The Express API consists of various methods and properties on the request and response objects. These are inherited by prototype. There are two extension points for the Express API:
+
+1. The global protoypes at `express.request` and `express.response`.
+2. App-specific prototypes at `app.request` and `app.response`.
+
+Altering the global prototypes will affect all loaded Express apps in the same process. If desired, alterations can be made app-specific by only altering the app-specific prototypes after creating a new app.
+
+## Methods
+
+You can override the signature and behavior of existing methods with your own, by assigning a custom function.
+
+Following is an example of overriding the behavior of [res.sendStatus](/4x/api.html#res.sendStatus).
+
+```js
+app.response.sendStatus = function (statusCode, type, message) {
+  // code is intentionally kept simple for demonstration purpose
+  return this.contentType(type)
+    .status(statusCode)
+    .send(message)
+}
+```
+
+The above implementation completely changes the original signature of `res.sendStatus`. It now accepts a status code, encoding type, and the message to be sent to the client.
+
+The overridden method may now be used this way:
+
+```js
+res.sendStatus(404, 'application/json', '{"error":"resource not found"}')
+```
+
+## Properties
+
+Properties in the Express API are either:
+
+1. Assigned properties (ex: `req.baseUrl`, `req.originalUrl`)
+2. Defined as getters (ex: `req.secure`, `req.ip`)
+
+Since properties under category 1 are dynamically assigned on the `request` and `response` objects in the context of the current request-response cycle, their behavior cannot be overriden.
+
+Properties under category 2 can be overwritten using the Express API extensions API.
+
+The following code rewrites how the value of `req.ip` is to be derived. Now, it simply returns the value of the `Client-IP` request header.
+
+```js
+Object.defineProperty(app.request, 'ip', {
+  configurable: true,
+  enumerable: true,
+  get: function () { return this.get('Client-IP') }
+})
+```
+</div>
--- a/_includes/header/header-en.html
+++ b/_includes/header/header-en.html
@@ -56,6 +56,8 @@
                           </li>
                           <li><a href="/{{ page.lang }}/guide/using-middleware.html">Using middleware</a>
                           </li>
+                          <li><a href="/{{ page.lang }}/guide/overriding-express-api.html">Overriding the Express API</a>
+                          </li>
                           <li><a href="/{{ page.lang }}/guide/using-template-engines.html">Using template engines</a>
                           </li>
                           <li><a href="/{{ page.lang }}/guide/error-handling.html">Error handling</a>


crandmck:
I may have some small wording improvements, but I'll wait first to see if there are any technical review comments.

en/guide/overriding-sugar-methods.md
```diff
@@ -0,0 +1,42 @@
...
+
+Sugar methods are methods added by Express on Node's HTTP request and response objects, which make up the bulk of Express' documented methods. Since the request and response objects are accessible from an instance of Express, you can override their signature and behavior with your own implementation by assigning a custom function.
+
+Here is a very simple example of overriding the [res.send](/{{ page.lang }}/4x/api.html#res.send) method. This will force all request to the app to be rejected with a 403 response.
+
+```js
+var express = require('express')
+var app = express()
+
+app.response.send = function () {
```
dougwilson: Example may not be useful enough to make it worth it, even though it's goal is to show overriding. I would just remove this and keep the following one as the example, unless you had something in mind for having both?

dougwilson (commented):
One technical-level comment I had :)

en/guide/overriding-sugar-methods.md
hacksparrow: I was planning on "conditioning" the reader first :), I am fine with removing it.

crandmck:
Now that I read this more closely, I have a couple of more substantive questions/comments...

> Sugar methods ... make up the bulk of Express' documented methods.

This begs the question of exactly which methods are sugar methods that you can override?   Is it actually _all_ of them?  Only a subset? ...if so, we should clearly list them. Are there any that you shouldn't (even if you technically could?)  Are there any side-effects of doing so to caveat? Rather than focus on what is technically possible, I think we should state "best practices" to the extent we know.

Also, although I think most developers are familiar with the concept of "syntactic sugar" I don't think the term "sugar methods" would be universally understood.  I suggest instead the title should just be "Overriding Express methods" and then explain the in the text what a "sugar method" is and (per above) exactly which methods they are.

Also, it would be nice to give some idea of why you would want to do this, i.e. what you can accomplish by doing so.

hacksparrow:
> This begs the question of exactly which methods are sugar methods that you can override? ...

The Express request and response objects extend the Node HTTP [request object](https://nodejs.org/api/http.html#http_class_http_incomingmessage) and [response object](https://nodejs.org/api/http.html#http_class_http_serverresponse) respectively, and add methods such as `req.accepts`, `req.param`, `res.send`, `res.status`, and so on, which make for an easy and friendly API for the users to interact with the underlying Node methods and properties.

Technically, one can override all of the Express methods. They can even override Node HTTP objects' methods.

Overriding Express or Node methods has a very open-ended outcome, so it is not possible to state "best practices" or side-effects. It can range from "same behavior" to "completely broken" and everything possible in between. Basically, if someone decides to override the methods, they are on their own, and should know what they are doing.

One tip that should be given to the "overriders" is to look the source code of the methods, if they decide to override. That will give them a clear picture of what (else) is happening in the method.

> Also, although I think most developers are familiar with the concept of "syntactic sugar"...

Yes, "Overriding Express methods" sounds better. Let's change it as suggested.

> Also, it would be nice to give some idea of why you would want to do this ...

Reason, "because Express does not do it the way I want it". Eg: https://github.com/expressjs/expressjs.com/issues/757

@crandmck I hope you will recompose the content based on my input.

hacksparrow:
My PR was based on the discussion at https://github.com/expressjs/express/issues/3151. After listening to the discussion in the recording, I think it makes sense to extend "sugar properties" as well. So the title would be now "Overriding the Express API" or something along that line.

@dougwilson `app`, the Express application instance's methods can be overriden too. Do we want to add it as well? Similarly, Router instances.

@crandmck I will add info about overriding properties and other possibilities in the next comment.

hacksparrow:
**Documented** properties in the Express API (limited to `request` and `response` in this write up) are either:

A. Inherited from the Node core

`res.headersSent`

There are many other core properties, only `res.headersSent` is documented on the website. I don't remember why it was decided to document only `headersSent`; maybe becaus of this - https://github.com/expressjs/express/issues/2109.

B. Assigned properties

`req.app`, `req.baseUrl`, `req.originalUrl`, `res.app`, etc.

These properties are assigned on to the `request` and `response` objects in the context of the urrent request-response cycle by the router.

C. Defined as getters.

`req.protocol`, `req.secure`, `req.ip`, `req.ips`, `req.subdomains`, etc.

Since A and B are dynamically added by the router for each new request-response cycle, they cannot be overriden.

Properties under C are basically helper methods defined as getters, which return the values in the context of the current request-response cycle, so their behavior can be changed by re-assigning a new getter.

The following code rewrites how the value of the `ip` is to be derived. Now, it simply returns '999.999.999.999' (an invalid IP address).

```
Object.defineProperty(app.request, 'ip', {
  configurable: true,
  enumerable: true,
  get: function () { return '999.999.999.999'; }
});
```

The assignment operator has no affect on properties defined with getters. For example:

```
app.get('/', (req, res) => {
    req.ip = 'HELLO'
    console.log(req.ip)
  }
)
```

`req.ip` won't be "HELLO" in the code above, it will still be what the getter was programmed to return.

hacksparrow:
@expressjs/express-tc do you suggest documenting the ability to extend and override the methods and properties on the express, app, and router instances as well?

crandmck:
@hacksparrow Let's try to land this!  

In TC, we discussed just adding minimal info to start, and then we can add more later if we want to.    Can you please make any final changes and then request final review?

hacksparrow:
@crandmck I have updated the PR for final review.

@expressjs/express-tc

en/guide/overriding-express-api.md
```diff
@@ -0,0 +1,69 @@
...
+```js
+res.sendStatus(404, 'application/json', '{"error":"resource not found"}')
+```
+
+## Properties
+
+Properties in the Express API are either:
+
+1. **Inherited from the Node core** - `res.headersSent` etc.
+2. **Assigned properties** - `req.app`, `req.baseUrl`, `req.originalUrl`, `res.app`, etc.
```
dougwilson: `req.app` and `res.app` are actually just part of the prototype, so could be overwritten (but why??)

en/guide/overriding-express-api.md
```diff
@@ -0,0 +1,69 @@
...
+
+```js
+Object.defineProperty(app.request, 'ip', {
+  configurable: true,
+  enumerable: true,
+  get: function () { return '999.999.999.999'; }
+});
+```
+
+The assignment operator has no affect on properties defined with getters. For example:
```
dougwilson: Since the article is "Overriding the Express API" I'm not 100% sure we even need to mention this aspect.

hacksparrow: Someone might try to overwrite the property by assigning a new value and observe that it does not work.

So maybe add it as a sidenote in an info box?

en/guide/overriding-express-api.md
hacksparrow: @dougwilson  "buy why" as in "**but why** would someone want to overwrite it"?

en/guide/overriding-express-api.md
dougwilson: I agree, but seems like something we should probably add to all the getter-only properties in the API docs :)

en/guide/overriding-express-api.md
dougwilson: Yea, sorry. I was saying that they are properties that are on the prototype, not assigned properties as written here. That means that someone actually _can_ override them, and the "buy why??" part was just wondering why someone would want to--which was unnecessary commentary :)

dougwilson (approved):
time for this to land. there was an issue recently where this was used, so good to have something to point to. rebased and fixed up for merge

PR #773, opened 2017-01-31
Fix logo anchor such that it is clickable again

aaronang:
In this patch, the logo is assigned a z-index such that it can be used again to navigate to the homepage. Prior to this patch, the `div#navbar` was overlapping the logo, only allowing to the top border of the logo to be clickable.

Note that I am no expert in CSS and therefore I am not sure if this the _best_ solution. Let me know what you think 😬

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -81,7 +81,7 @@ strong, th {
 
 #homepage-rightpane {
   min-width: 500px;
-  padding: 25px 30px 0px 0px; 
+  padding: 25px 30px 0px 0px;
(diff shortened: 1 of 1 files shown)


crandmck:
@hacksparrow PTAL

hacksparrow:
@iamropo ping!

iamropo:
LGTM, Thanks @aaronang

aaronang:
@iamropo It was my pleasure 😄

PR #776, opened 2017-02-02
Parameterize current Express version for updating

dougwilson:
In going to update the site to display 4.14.1 as the current version, I see that it's a little tedious. I'm not sure if there is some other style of parameterizing that is desired, but I broke out a really simple one for the Express version, at least, so it doesn't require updating a dozen files every time :)

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,4 +1,10 @@
 # Site settings
+defaults:
+  -
+    scope:
+      path: "" # all pages
+    values:
+      current_express_version: "4.14.0"
 
 # Build settings
 markdown: kramdown
--- a/de/index.md
+++ b/de/index.md
@@ -9,7 +9,7 @@ lang: de
     <div id="overlay"></div>
     <div id="homepage-leftpane" class="pane">
     <section id="description">
-        <div class="express"><a href="/">Express</a><a href="/{{ page.lang }}/changelog/4x.html#4.14.0" id="express-version">4.14.0</a></div>
+        <div class="express"><a href="/">Express</a><a href="/{{ page.lang }}/changelog/4x.html#{{ page.current_express_version }}" id="express-version">{{ page.current_express_version }}</a></div>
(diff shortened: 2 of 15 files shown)


crandmck:
Cool.... this is one way to do it, and it certainly works.  The `_config.yml` file is meant for defining [default values for page variables](http://jekyllrb.com/docs/configuration/#front-matter-defaults).

Probably the more standard way (cf http://jekyllrb.com/docs/datafiles/) would be (for example) to create a `_data/version.yml` file with the single key `current_express_version: 4.14.0`, then refer to that as `{{site.data.version.current_express_version}} in pages.  But I don't think it matters too much what way we do it, as long as we're consistent and understand how to update the version number.

crandmck:
Actually, I spoke a bit too soon, now that I look at my local running site...  A couple things:
- Did you mean to have the version be `4.14.0`?  I thought you wanted to link to 4.14.1?  In this PR, it's still 4.14.0 in `_config.yml` and thus on all the home pages.
- The link in the English home page works fine to http://expressjs.com/en/changelog/4x.html#4.14.0 but on non-en pages, it goes to (for example, for Spanish) http://expressjs.com/es/changelog/4x.html#4.14.0 which doesn't work since we don't translate the changelog.  You just have to remove `{{ page.lang }}` from the URL.
- I don't see the link at all in the German page, which is really weird because it's clearly there in the code.  I don't have time to figure it out now, but there must be something going on there... all other langs look OK.
- The version number is pretty darn faint and easy to overlook.  I guess I never really paid attention.... but maybe we could make stand out just a bit more if we want folks to see it?

dougwilson:
Yea, all that makes sense. Let's hold this PR for me to get some updates put together. PR #778 is the fix for the German page issue you saw.

crandmck:
Great, LMK if I can help.... Went ahead and landed #778, thx for fixing!

dougwilson:
Hey sorry I let this longer, kinda forgot about it :) Since the version number isn't even on most of the pages any longer, defiantly less of a concern anyway :)

PR #777, opened 2017-02-03
Fix error handing in a res.sendFile example

dougwilson:
This is a possible way to fix #774 . The issue is that `err.status` may be `undefined`, and in newer versions of Node.js that will cause an error to be thrown. Since the example wasn't really trying to demonstrate doing anything of value with the error, I think it's better to show it being handed to `next()` to participate in the error handling pipeline of the app.

PR #778, opened 2017-02-03
Fix Express version style on de index

dougwilson:
This fixes the issue with the Express version number on the home page for de (http://expressjs.com/de/) as noticed in #777


crandmck:
LGTM, I'll go ahead and merge.

PR #782, opened 2017-02-11
Fix html escaping on japanese behind-proxies page

kadoppe:
I found character entity references like `</li> </ul>` on the page below: 
  * http://expressjs.com/ja/guide/behind-proxies.html

To fix this, I need to move close tag `</li>` to the next line.


hacksparrow:
Thanks @kadoppe

PR #785, opened 2017-02-14
Generator page (all languages): Changed 2nd example to use Pug

martinjoiner:
I have changed the 2nd example on the Express Application Generator page to use Pug instead of Jade to help avoid confusion for new adopters. 

I have left the reference to Jade in the first section where it states that Jade is the default. As this is still true it would be unhelpful to remove it until the default becomes Pug (or whatever it may become in future).

I have also expanded the description that precedes the 2nd example in the Englsih, Spanish and French versions of the page. Translators were unavailable for the other versions.


crandmck:
Thanks!  I think this will help to avoid confusion.

PR #787, opened 2017-02-25
Add `database` field to mysql connection section.

christopher4lis:
The `mysql.createConnection()` snippet only had connection information for `host`, `user`, and `password`, but should have a `database` field as well for comprehensiveness. This way the user knows the property name without having to do an additional Google search on the mysql module.

### DIFF
--- a/en/guide/database-integration.md
+++ b/en/guide/database-integration.md
@@ -170,10 +170,11 @@ $ npm install mysql
 ```js
 var mysql = require('mysql')
 var connection = mysql.createConnection({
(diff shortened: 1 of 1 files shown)


dougwilson:
Indeed. Also, so people don't have to Google search to lean more about the module, can you add a link to the module as well?

christopher4lis:
Ah, looks like a link is already in there, didn't even notice.

crandmck:
LGTM, thanks!

PR #789, opened 2017-02-26
Fix get-readmes.sh for parity with written files

dougwilson:
So when I just did #788 I noticed that there were a couple issues I had running the bash script, and this is my attempt to fix them:

1. The `\n` was literal in the output. I'm not sure if this is a difference in bash versions / environment, but from what I can tell, they should be enclosed in `$''` syntax to work everywhere. At least, that does fix it on my machine.
2. The `<div>` wrapper was missing from the output.

/cc @crandmck

### DIFF
--- a/get-readmes.sh
+++ b/get-readmes.sh
@@ -36,14 +36,16 @@ get_release_vers() {
 # </div>
 
 writefile_mw() {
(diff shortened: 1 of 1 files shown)


crandmck:
Thanks @dougwilson I do want to rework how we get READMEs from other repos to be a bit cleaner, but in the meantime these are good improvements.

Issue #791, opened 2017-02-28
Stage 4.15 API documentation


dougwilson:
I'm not sure how we should go about this, but figured at least open a tracking issue. I assume that we can just make a 4.15 branch in here and commit / PR to it for the different things in prep for the 4.15 release.

- [x] 4.15 changelog
- [x] 4.15 announcement
- [x] API docs added for `next("router")` (implemented in https://github.com/expressjs/express/pull/3187/commits/9722202df964bfbfc0f579e4baeb5a4e1b43b344)

crandmck:
> ...just make a 4.15 branch in here and commit / PR to it ...

Yes, that's the best way to do it, afaik.

PR #792, opened 2017-03-01
Add Express 4.15 documentation

dougwilson:
This is my initial stab at adding the Express 4.15 information for #791 . The `next('router')` documentation can use some more beefing up, but at the very least it's mentioned somewhere and we can probably land as we get it better updated.

### DIFF
--- a/_includes/announcement/announcement-en.md
+++ b/_includes/announcement/announcement-en.md
@@ -1,8 +1,8 @@
 <ul>
   <li>
-    <p class="announcement-title">Express 4.14.0 is the current release.</p>
+    <p class="announcement-title">Express 4.15.0 is the current release.</p>
     <p markdown="1">
-     For more information on what was added in this release, see the [4.14.0 changelog](/{{ page.lang }}/changelog/4x.html#4.14.0).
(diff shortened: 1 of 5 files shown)


dougwilson:
Note that Express 4.15 hasn't actually been released yet, so links to the history file assume a release date of 3/1/2017 for now.

dougwilson:
Sorry I ended up merging this, since a security issue was posted on one of our dependencies today and even though we're not affected, people see it, so just pushed the release out. We can edit the site as needed from here :)

crandmck:
Thanks for updating the site @dougwilson.  

I made one small change with c8e36fc8ff2433dcebab3622f40cea48393c4225 to update the version number displayed next to the big "Express" title/logo on the home page ... it was still saying "4.14.1".  

I went ahead and implemented the idea that you suggested a while back of putting the current version in a data file.  I just called it [`express.yml`](https://github.com/expressjs/expressjs.com/blob/gh-pages/_data/express.yml) and for now it only has one key, `current_version`, which you can access in any page as `{{site.data.express.current_version}}`.

dougwilson:
Oh, haha, totally meant to update that, but forgot 😊 thanks for the fixup!!

PR #797, opened 2017-04-07
Added the wrapper to JSONP

franciscop:
This displays the use as it would work in the real world. Stays in line with the comment and documentation shown later on.


hacksparrow:
Thanks @franciscop

PR #799, opened 2017-04-18
Fix typo in app get function description

robbiejaeger:
Before, the description of the app.get() function read:
> Returns the value of name app setting, where name is one of strings in the app settings table.

Sentence was missing *the* to instead read:
> Returns the value of name app setting, where name is one of *the* strings in the app settings table.


hacksparrow:
Thanks @robbiejaeger

PR #800, opened 2017-04-21
Update using-middleware.md

zhenhuixiang:
A small modification

### DIFF
--- a/zh-cn/guide/using-middleware.md
+++ b/zh-cn/guide/using-middleware.md
@@ -142,7 +142,7 @@ var router = express.Router();
 
 <pre>
 <code class="language-javascript" translate="no">
-var app = express();
+var express = express();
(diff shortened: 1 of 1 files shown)


crandmck:
@zhenhuixiang Thanks for raising this issue.

I believe the example code is correct "as is", but for clarity, it should be
```
var express = require("express");
var router = express.Router()
...
```

Instead of what's there now:
```
var app = express()
var router = express.Router()
...
```

The first line is confusing and irrelevant to these examples.  

The same problem occurs in the next block of example code (that has the sentence "This example shows a middleware sub-stack..." preceding it).

So, if you want to change your PR accordingly, we will gladly land it.

zhenhuixiang:
@crandmck  Thank you, I understand now. there is no problem.

PR #803, opened 2017-04-25
static: Fixed typo in API.

WORMSS:
options.extension is currently listed as type "Boolean", but "true" is not a valid value.
options.extension is only able to accept an array of strings or a false.
I have changed to "Mixed" to match options.index type.

I had thought to also add the extra information of valid options below, but I wanted communities instructions to if this is necessary or not.


crandmck:
@dougwilson Any input?

dougwilson:
Makes sense.

crandmck:
Thanks!

Issue #781, opened 2017-02-06
Add availability/deprecation information on API doc


hacksparrow:
To address issues such as #779.

hacksparrow:
@iamropo this one is for you.

crandmck:
@hacksparrow is @iamropo going to address this, or should I put it in my queue?

hacksparrow:
@crandmck sorry for such a late response. @iamropo will not be able to address it, in your queue please 🙏

PR #806, opened 2017-05-06
Add notification of API update for req.hostname

piyushpatel2005:


dougwilson:
I think this is good, but should probably be part of #781 ideal, for example, rather than adding this footnote, we should actually document `req.host` and mark it as deprecated and add to the `req.hostname` docs what version it was added.

piyushpatel2005:
I understand your point of view and that is precisely correct. However, Express 4.x API documentation does not include any doc for **req.host** and hence it causes confusion among API users. I tried to redirect using req.hostname. However, I had Express v4.0 and I was getting req.hostname = undefined. I checked the documentation for 4.x, but didn't find anything relevant as it does not include any details regarding req.host.

I had to dig down into the code to figure out that `req.hostname` didn't exist for Express 4.0 which makes API documentation fail its sole purpose. Please, include a note or a doc for `req.host` with deprecated status. I am a new learner and I believe it should be part of documentation so that a user need not dig into the code.

Thanks.

dougwilson:
Right, I know that req.host is missing. That's why I'm saying that  think this is good, but should probably be part of #781 ideal, for example, rather than adding this footnote, we should actually document  req.host  and mark it as deprecated and add to the  req.hostname  docs what version it was added.

sandizapar (approved):

PR #810, opened 2017-05-18
Fix header width

jakepanitz:
Header overflowed page at certain width. Changed css to make the header more responsive.


crandmck:
I tested this change, and it is an improvement.  Thanks!

PR #811, opened 2017-05-21
Updated hello world to use const.

iddogino:
This doc is referred to as a 'boiler plate' starting place for many developers. It should use constants - by far the best practice in this scenario for over a year now.


crandmck:
This makes sense IMHO.  

My only reservation is that we should probably make a similar change throughout the docs.  But starting here is fine, I think.

PR #815, opened 2017-06-01
npm 5.0 --save and --no-save changes

shivajivarma:
npm now saves dependencies by default without needing to place --save flag. 

It requires --no-save flag to install temporarily without adding to dependencies.

### DIFF
--- a/en/starter/installing.md
+++ b/en/starter/installing.md
@@ -34,12 +34,17 @@ Enter `app.js`, or whatever you want the name of the main file to be. If you wan
 Now install Express in the `myapp` directory and save it in the dependencies list. For example:
 
 ```sh
-$ npm install express 
+$ npm install express --save
 ```
 
-To install Express temporarily and not add it to the dependencies list, add the `--no-save` option:
+To install Express temporarily and not add it to the dependencies list:
 
-```sh
+1. omit the `--save` option for older version for npm.
+  ```sh
+$ npm install express
+```
+2. add the `--no-save` option for npm 5.0.0 and above.
+  ```sh
 $ npm install express --no-save
 ```
 
--- a/en/starter/installing.md
+++ b/en/starter/installing.md
@@ -39,13 +39,8 @@ $ npm install express --save
 
 To install Express temporarily and not add it to the dependencies list:
 
-If using npm 5.0.0 or later, use the --no-save option:
-  ```sh
+```sh
 $ npm install express --no-save
-```
-If using npm version prior to 5.0.0, omit the --save option:
-  ```sh
-$ npm install express
 ```
 
 <div class="doc-box doc-info" markdown="1">
--- a/en/starter/installing.md
+++ b/en/starter/installing.md
@@ -39,17 +39,16 @@ $ npm install express --save
 
 To install Express temporarily and not add it to the dependencies list:
 
-1. omit the `--save` option for older version for npm.
-  ```sh
-$ npm install express
-```
-2. add the `--no-save` option for npm 5.0.0 and above.
+If using npm 5.0.0 or later, use the --no-save option:
   ```sh
 $ npm install express --no-save
 ```
+If using npm version prior to 5.0.0, omit the --save option:
+  ```sh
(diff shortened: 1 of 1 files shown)


dougwilson:
Thanks, but of course this will only lead to confusion with all existing and older installations. Is there a way for the instructions to accommodate both? For example, just have both commands include the option, then the instructions will automatically become valid for all versions of npm without requiring the user to determine their version.

dougwilson (changes requested):
See comment

shivajivarma:
👍  updated accordingly.

crandmck:
@shivajivarma That's an improvement, but some small comments:
- A numbered list is not appropriate here, since you're not performing the two things sequentially.  You could use a bullet list, but I think just paragraphs will suffice.
- The phrase "...for older version " is ambiguous (older than what?) since it refers forward to something the reader hasn't encountered yet.  
- The final para about `-save` is somewhat confusing with the new info, so it needs rewording.

I suggest rewording as below.
...
To install Express temporarily and not add it to the dependencies list:

If using npm 5.0.0 or later, use the `--no-save` option:
```
$ npm install express --no-save
```
If using npm version prior to 5.0.0, omit the `--save` option:
```
$ npm install express
```

By default with version npm 5.0+ `npm install` adds the module to the `dependencies` list in the `package.json` file; with earlier versions of npm, you must specify the `--save` option explicitly.   Then, afterwards, running `npm install` in the app directory will automatically install modules in the dependencies list.

dougwilson (changes requested):
In addition, I don't think it's good to require the user to determine their version of npm. Just include the command argument in all the examples, as having the command argument explicitly given works across all versions without requiring the user to understand what version they are using.

shivajivarma:
@dougwilson: that cannot be done, as there is no `--no-save` option available in npm version prior to 5.0. And default install behavior (no options) got changed from 'save' to 'no save'.

shivajivarma:
@crandmck thanks for the feedback, updated the content.

dougwilson:
Right, I know there is no `--no-save` option below 5.0, but `npm` does not error out if you give it an argument it doesn't understand, and since no save is the default, adding a `--no-save` doesn't error and performs that action even in npm versions below 5.0.
```
$ npm -v
2.15.9

$ npm install --no-save express
express@4.15.3 node_modules\express
├── escape-html@1.0.3
├── utils-merge@1.0.0
├── methods@1.1.2
├── content-type@1.0.2
├── encodeurl@1.0.1
├── range-parser@1.2.0
├── parseurl@1.3.1
├── setprototypeof@1.0.3
├── array-flatten@1.1.1
├── statuses@1.3.1
├── cookie-signature@1.0.6
├── vary@1.1.1
├── cookie@0.3.1
├── merge-descriptors@1.0.1
├── fresh@0.5.0
├── serve-static@1.12.3
├── path-to-regexp@0.1.7
├── content-disposition@0.5.2
├── etag@1.8.0
├── depd@1.1.0
├── qs@6.4.0
├── on-finished@2.3.0 (ee-first@1.1.1)
├── finalhandler@1.0.3 (unpipe@1.0.0)
├── debug@2.6.7 (ms@2.0.0)
├── proxy-addr@1.1.4 (forwarded@0.1.0, ipaddr.js@1.3.0)
├── send@0.15.3 (destroy@1.0.4, ms@2.0.0, mime@1.3.4, http-errors@1.6.1)
├── accepts@1.3.3 (negotiator@0.6.1, mime-types@2.1.15)
└── type-is@1.6.15 (media-typer@0.3.0, mime-types@2.1.15)
```

I don't think it's good to require the user to determine their version of npm. Just include the command argument in all the examples, as having the command argument explicitly given works across all versions without requiring the user to understand what version they are using.

crandmck:
> I don't think it's good to require the user to determine their version of npm. 

Agree, it's better to provide a single path that works for everyone.  

Looks like @shivajivarma made the changes you requested, so LGTM; landing.

dougwilson:
👍

PR #819, opened 2017-06-18
Fix GFM link in README.md

sehrope:
Fixes the GFM link on the main README. The title/url are reversed.


hacksparrow:
Thanks @sehrope

PR #820, opened 2017-06-18
Explain (dis)allowed modifications in app.param callbacks

sehrope:
Adds some wording and examples to document (un)acceptable modifications to `req.params` in `app.param(...)` handlers. 

Per discussion in https://github.com/expressjs/express/issues/3341

### DIFF
--- a/_includes/api/en/4x/app-param.md
+++ b/_includes/api/en/4x/app-param.md
@@ -23,6 +23,38 @@ app.param('user', function(req, res, next, id) {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


hacksparrow:
@sehrope the possibility of modification of `req.param` and its keys, and its effect is not just limited to param callback functions, so I have updated the `req.params` instead, please take a look - https://github.com/expressjs/expressjs.com/pull/821.

hacksparrow:
Closing this, as it is taken care of by https://github.com/expressjs/expressjs.com/pull/821

PR #821, opened 2017-06-19
Do not use req.params for setting state

hacksparrow:
Ref:
- https://github.com/expressjs/express/issues/3341
- https://github.com/expressjs/expressjs.com/pull/820

@sehrope based on your feedback, I have updated the `req.params` doc to include a warning:

_"Any changes made to the `req.params` object and its keys may be overwritten by subsequent middleware in the stack. Do not use them for setting any state from your application code."_

@dougwilson @crandmck

### DIFF
--- a/_includes/api/en/4x/req-params.md
+++ b/_includes/api/en/4x/req-params.md
@@ -15,3 +15,7 @@ When you use a regular expression for the route definition, capture groups are p
 req.params[0]
 // => "javascripts/jquery.js"
 ```
+
+If you need to make changes to a key in `req.params`, use the [app.param](/{{ page.lang }}/4x/api.html#app.param) handler. Changes are applicable only to [parameters](/{{ page.lang }}/guide/routing.html#route-parameters) already defined in the route path.
+
+Any changes made to the `req.params` object in a middleware or route handler will be reset.


sehrope:
Per @dougwilson in the original issue I created, it's my understanding that it's acceptable to modify the value of the param being handled, i.e.:

```js
app.param('foo', function(req, res, next, id) {
  Foo.load(id, function(err, foo) {
    if (err) {
      return next(err);
    }
    req.params.foo = foo;
    return next();
  });
});
```

That should be called out in the docs as being explicitly allowed as it's a common pattern for REST endpoints.

Specifically, it's only okay to overwrite the value of the param name being handled by the param callback (i.e. `req.params[name] = ...` is okay). It's never okay to assign values to other attributes on `req.params` (i.e. `req.params['otherName'] = ...` is never okay).

hacksparrow:
@sehrope I had missed some context; have updated the text accordingly. Please see if the updated content sums up what you had on mind and would have helped you.

sehrope:
Not sure if "trigger" is the right word there. Perhaps "handler" would be better?

Otherwise this looks good and would have answered my original question.

hacksparrow:
@sehrope indeed, it should be "handler". Thanks for the feedback!

sehrope:
Thanks for the quick fix to the docs and a more general thanks to everyone that works on Express (app, docs, and all else too). Great piece of tech and many years of pleasant experience with it.

PR #822, opened 2017-06-20
Embed a working example server on the "Hello World" tutorial

boucher:
I'd like to propose adding a runnable embedded Express server into the documentation, to make it easy for people to play around with Express and see how it works. I've implemented a fairly small set of changes to the existing "Hello World" tutorial to make this possible, which ends up looking like this:

screenshot 2017-06-20 11 44 29

Of course, I know that given this document needs to be translated, there are perhaps reasons to change the wording less, or perhaps people would prefer another approach or to discuss including on a different page.

This is powered by [RunKit](https://runkit.com) (which I helped create), which is creating a virtual server on demand behind the scenes. And if for whatever reason RunKit goes down, the example will fall back to just looking as it does now.

### DIFF
--- a/en/starter/hello-world.md
+++ b/en/starter/hello-world.md
@@ -9,14 +9,11 @@ redirect_from: "/starter/hello-world.html"
 # Hello world example
 
 <div class="doc-box doc-info" markdown="1">
-This is essentially going to be the simplest Express app you can create. It is a single file app &mdash; _not_ what you'd get if you use the [Express generator](/{{ page.lang }}/starter/generator.html), which creates the scaffolding for a full app with numerous JavaScript files, Jade templates, and sub-directories for various purposes.
+Embedded below is essentially the simplest Express app you can create. It is a single file app &mdash; _not_ what you'd get if you use the [Express generator](/{{ page.lang }}/starter/generator.html), which creates the scaffolding for a full app with numerous JavaScript files, Jade templates, and sub-directories for various purposes.
 </div>
 
-First create a directory named `myapp`, change to it and run `npm init`. Then install `express` as a dependency, as per the [installation guide](/{{ page.lang }}/starter/installing.html).
-
-In the `myapp` directory, create a file named `app.js` and add the following code:
-
-```js
+<script src="https://embed.runkit.com" data-element-id="hello-example" data-mode="endpoint" async defer></script>
+<div id="hello-example"><pre><code class="language-js">
 const express = require('express')
 const app = express()
 
@@ -27,10 +24,18 @@ app.get('/', function (req, res) {
 app.listen(3000, function () {
   console.log('Example app listening on port 3000!')
 })
-```
+</code></pre></div>
+
+The example above is actually a working server, go ahead and click on the URL. You'll get a response, with real time logs right here on the page, and any changes you make will be reflected in real time. Below, you'll find instructions for running the same app on your local machine.
 
 The app starts a server and listens on port 3000 for connections. The app responds with "Hello World!" for requests
 to the root URL (`/`) or _route_. For every other path, it will respond with a **404 Not Found**.
+
+### Running Locally
+
+First create a directory named `myapp`, change to it and run `npm init`. Then install `express` as a dependency, as per the [installation guide](/{{ page.lang }}/starter/installing.html).
+
+In the `myapp` directory, create a file named `app.js` and copy in the code from the example above.
 
 <div class="doc-box doc-notice" markdown="1">
 The `req` (request) and `res` (response) are the exact same objects that Node provides, so you can invoke
--- a/en/starter/hello-world.md
+++ b/en/starter/hello-world.md
@@ -26,10 +26,10 @@ app.listen(3000, function () {
 })
 </code></pre></div>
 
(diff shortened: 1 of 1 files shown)


crandmck:
The idea is fine in principle, and would probably be useful for readers learning Express.

My only reservation is using a commercial third-party service (runkit).  I don't think we want to imply that Express is endorsing it, even though it looks cool and useful.  If we did include this we would want a disclaimer to that effect.  Also, we might want to put it on a separate page so that it would be an optional path for Getting Started.

Thoughts @expressjs/express-tc ?

hacksparrow:
I am not in favor of it. The Express API is simple enough, and the lesser external dependencies, the better.

boucher:
I've put a version of the site online so you can play with it to get a better idea of how this works:
http://rossboucher.com/express/

@crandmck I'd be happy to add something in about not being an endorsement (or really, to rework this in whatever way you think makes the most sense for your users). 

I should also mention that our integration with [npmjs.com](https://npmjs.com) sends a number of users to the Express example hosted on [our site](https://npm.runkit.com/express), but I think it's a better experience to have the example closer to your own documentation.

@hacksparrow I do think the Express API is quite simple, especially once you've used it a bit. But I also think having a way to make changes and get immediate feedback would help some people figure it out more quickly.

And here's a (lowish quality, sorry) gif just for fun:
express3

dougwilson:
So I am neutral about it, though it is very tempting because it looks like it would work very well. And yea, all npmjs.com listings have a link to RunKit currently, you can see this on the right-hand side of https://www.npmjs.com/package/express for example.

crandmck:
Thanks @boucher for the example and your flexibility.  I think we would want some kind of disclaimer, but before word-smithing that, I'd like to see if we can come to a consensus on whether to add this at all.  So far we have one "nay" and one "neutral".  

I can't tell if @LinusU is in favor or neutral?  Anyone else care to weigh in, e.g. @blakeembrey ?

LinusU:
> I can't tell if @LinusU is in favor or neutral?

Hehe, yeah sorry about that vague 👍 😄 

I agree with everything you said in that message, I think it's a really nice feature but we should make clear that it's from a third party that we are not endorsing...

Super nice example though @boucher, great work!

jdalton:
If you all are looking for another example integration, I added runkit to [Lodash's documentation](https://lodash.com/docs/4.17.4) about a year ago. The Lodash site runs offline and toggles the repl off when in offline mode and other such things. I went all out with resource hints and all that jazz. Runkit has a preview of code style themes I still need to look into now too. I dig them for sure.

dougwilson:
Nice. I mean, if it does cause a technical issue, we can always remove it.

crandmck:
OK, so Linus is for it, and possibly Doug is coming around (?).  I'm leaning toward accepting this PR, too.  It's a neat capability.

But I do think we should add some disclaimer text like:

> RunKit provides an interactive JavaScript playground connected to a complete Node environment that runs in your web browser.  RunKit is a third-party service that is not affiliated with Express and the Express team is not responsible for any bugs or issues with it.  

We could also include something a bit more forceful like:

"It is included here for the convenience of people learning Express, and does not constitute an endorsement by the Express project."

Thoughts?

jdalton:
> "It is included here for the convenience of people learning Express, and does not constitute an endorsement by the Express project."
>
>Thoughts?

From an outsider it seems overly formal. In my experience with runkit in Lodash, which is the most directly depended upon package in the npm ecosystem with 203K+ active users visiting the site a month, I've not had any push back or support burden related to the doc repl.

crandmck:
> From an outsider it seems overly formal.

Yeah, that's probably due to my background as a tech writer working on commercial projects for over 25 years (yikes!).  For an OS project, I agree it's probably overkill.  I think the first blurb is probably sufficient, but I'd like to see what others think.

dougwilson:
I say let's go for it, given further evidence that this is a product used around in multiple places. As for the disclaimer, I do agree that it is a bit strong, though I absolutely know where @crandmck is coming from :) Being that the site is not actually run by a company, though, I'm not sure that most of the disclaimer is really necessary. I think that it being on every package screen in npm without an endorsement disclaimer already may make people who would need such a disclaimer already think that packages are "endorsing it".

Perhaps I can see a use of adding the support comment, though that just adds wording that may confuse people, so we could always add it if it does become an issue.

Anyway, just my thoughts :)

crandmck:
OK, sounds good.  I still think it makes sense just to clarify that RunKit is entirely separate from 
 Express.  Perhaps that's obvious to us, but for a random person it might not be. 

So how about just adding:

> RunKit provides an interactive JavaScript playground connected to a complete Node environment that runs in your web browser. RunKit is a third-party service not affiliated with the Express project.

That's pretty minimal and just a statement of fact to eliminate any possible confusion.

boucher:
@crandmck added

crandmck:
LGTM. I might tweak the formatting at some point, but good enough for now.

boucher:
Thanks everyone! Let me know if you run into any trouble or have any suggestions for ways we could make this better for you. 

(Quick question: should I be doing something about getting the changed language translated into the other supported languages?)

crandmck:
> should I be doing something about getting the changed language translated into the other supported languages

Sorry for the delay, I've been on vacation--we rely on the community to update translations, so it's up to speakers of each language (German, Spanish, etc.) to open PRs for each language.

PR #824, opened 2017-06-21
Update error-handling.md

Draeli:
add next() in exemple for checkIfPaidSubscriber method, without this next, req is "hang".


dougwilson:
The `next()` addition in this PR is good, but a note: let's follow up with some fixes to this example in general. For example, that comment right above is very confusingly worded.

crandmck:
OK, thanks @dougwilson. 

I opened an issue to track that, please comment there with more specifics, etc.

PR #826, opened 2017-06-27
Update link of StrongLoop Blog in Resources/Books and blogs

DRL9:
Change the link(StrongLoop Blog: Express category) into https://strongloop.com/strongblog/tag_Express.html because the old doesn't exist.


crandmck:
Thanks!!

PR #827, opened 2017-06-27
Update generator.md

ushelp:
Add a new generator tools: Express-quiecker, to quickly create an application skeleton based on Express-quickstart(EasyTemplateJS Engine).


crandmck:
Although I'm sure this is a useful tool, in general on this site we document only modules in the `expressjs`, `jshttp`, and (potentially, though in practice very little) `pillarjs` orgs.  We can't possibly document every external module or tool for Express.

So, thanks for the PR, but I'm afraid we can't accept it.

PR #828, opened 2017-07-03
Make it more clear

davidgatti:
Went from, what is going on I have no idea what I'm reading, to: OK, now it is super clear. Especially for non English speaking people :)

### DIFF
--- a/en/guide/behind-proxies.md
+++ b/en/guide/behind-proxies.md
@@ -68,7 +68,7 @@ app.set('trust proxy', function (ip) {
   </tbody>
 </table>
 
-Setting a non-`false` `trust proxy` value results in three important changes:
+Setting `trust proxy` to `true` will results in three important changes:
 
 <ul>
   <li markdown="1">The value of [req.hostname](/{{ page.lang }}/api.html#req.hostname) is derived from the value set in the `X-Forwarded-Host` header, which can be set by the client or by the proxy.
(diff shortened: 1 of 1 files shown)


dougwilson:
You can also set it to a number, a string, or an array, all of which are non-false. The value true is only one possible way to trigger the behavior. I think if you are going to change it, you need to enumerate what all the possible non-false values are.

dougwilson (changes requested):
See comment

crandmck:
Also, "...will results in..." is incorrect grammar.  I would suggest either:
- "Setting `trust proxy` to (...specify values...) results in the following changes:"
- "If `trust proxy` is not `false` then:"

davidgatti:
OK, sounds like a plan :)

crandmck:
@davidgatti ping...

davidgatti:
Pong :) 

I know, so much poop going on :( will try to make it happen this weekend.

davidgatti:
@dougwilson I'm so sorry this simple thing took so much time, but I mad the change based on your suggestion. 

@crandmck, is the grammar OK?

dougwilson:
You missed that you can also set to a number :)

dougwilson:
(A non-falsy number, so zero does not enable, but 1, 2, 3, etc. all enable it for that number of hops).

dougwilson:
Now that I just re-read that secion, we can really just remove the first sentence and be done with it, imo

davidgatti:
🤣🤣🤣

davidgatti:
There you go ;)

crandmck:
Often, the simplest solution is the best...thanks! :-)

davidgatti:
Always :D

PR #837, opened 2017-08-08
remove unsupported methods

xieranmaya:
Update the doc as per - expressjs/express#2966
The related PR didn't remove all the mistakes in the doc


crandmck:
Thanks, I'm disappointed this mistake wasn't caught earlier, but better late than never!

PR #840, opened 2017-08-11
Resolve ambiguity in app.set documentation

### LINKED ISSUE #839: Documentation for app.set is ambiguous
I _think_ the recommended usage of `app.set` includes storing your own values in there. For instance, the examples in the docs show: `app.set('title', 'My Site');`. Yet the description for this method doesn't necessarily allow for this usage:

> Assigns setting `name` to `value`, where name is one of the properties from the app settings table.

This sentence seems to be quite explicit in that `name` must be one of the app settings values in the table that appears beneath this section, yet `title` doesn't appear in that table. This leads to ambiguity in how this method should be used.

If I'm right that this is also a general-purpose way to store values, would you be up for a PR that makes this more explicit? Consider:

> Assigns setting `name` to `value`. You may store any value that you want, but certain names have a special meaning to Express. These special names are listed in the app settings table.

If this is not the case, then the examples should probably be updated, and Express should possibly throw an Error if you set an invalid value.

Thanks for reading!

---

(Originally opened as https://github.com/expressjs/express/issues/3393 )

jamesplease:
Resolves #839

---

I often use this to access these values in my middleware, via `req.app.get('hello')`. I think it could be worthwhile to mention that somewhere, but I didn't put it here. Does anyone agree that that's worth adding somewhere? If so, let me know where and I can update this PR or do a follow up PR.

### DIFF
--- a/_includes/api/en/4x/app-set.md
+++ b/_includes/api/en/4x/app-set.md
@@ -1,7 +1,8 @@
 <h3 id='app.set'>app.set(name, value)</h3>
 
-Assigns setting `name` to `value`, where `name` is one of the properties from
-the [app settings table](#app.settings.table).
+Assigns setting `name` to `value`. You may store any value that you want,
+but certain names can be used to configure the behavior of the server. These
+special names are listed in the [app settings table](#app.settings.table).
 
 Calling `app.set('foo', true)` for a Boolean property is the same as calling
 `app.enable('foo')`. Similarly, calling `app.set('foo', false)` for a Boolean


dougwilson (approved):
This is good on a technical level.

dougwilson:
Thanks @jmeas ! Usually @crandmck will take a look from a word smithing point of view, but it looks good from a "is this technically correct?".

Yea, maybe some kind of expansion on the concepts it allows you to do (probably a diff PR since it may have more feedback) would be good if you're up for it :)

crandmck:
LGTM, thanks @jmeas !

> I often use this to access these values in my middleware, via req.app.get('hello'). ...

I think that would be worthwhile to mention, but let's do it in another PR.

jamesplease:
Thanks! Do you think mentioning that belongs somewhere in the Guides, or inline here in the API docs?

crandmck:
Perhaps add a brief mention in both places, e.g. in the API ref and in [Using middleware](http://expressjs.com/en/guide/using-middleware.html).    If it's brief then we could put it in both places....

jamesplease:
Sounds good. I'll do that. Thanks @crandmck !

PR #841, opened 2017-08-16
Include Oracle driver and example into list.

gvenzl:
Hi express community!

This commit adds the open source Oracle Node.js driver into the list and provide a simple connect example.

Thx,

### DIFF
--- a/en/guide/database-integration.md
+++ b/en/guide/database-integration.md
@@ -15,9 +15,9 @@ Adding the capability to connect databases to Express apps is just a matter of l
 * [CouchDB](#couchdb)
 * [LevelDB](#leveldb)
 * [MySQL](#mysql)
-* [Oracle](#oracle)
 * [MongoDB](#mongo)
 * [Neo4j](#neo4j)
+* [Oracle](#oracle)
 * [PostgreSQL](#postgres)
 * [Redis](#redis)
 * [SQL Server](#mssql)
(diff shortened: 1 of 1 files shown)


crandmck:
I think the Installation section should at least mention that installing `node-oracledb` requires Python and C compiler tools as specified in https://github.com/oracle/node-oracledb#-installation.  I only mention that because from experience I know that trips up a lot of people, and IMHO it's better to give a "heads up" because some Node devs might not expect that.

Otherwise, LGTM.  
@dougwilson  Do you have any input?

gvenzl:
Yeah, that's fair. Do you want me to put this into the page directly or link to the other page. The latter is perhaps a better option to automatically keep it in sync if things ever change but either is fine by me. Please advise.

dougwilson:
Yea, I think linking out to the page is good. Perhaps like around the `npm install` block that page can be linked to as "prerequisites". That section on their page does also spell out the `npm install` part, so another option is to replace the npm install with a link there (linked out as installation instructions or similar), which would of course force the eyes past those prereqs. That's my thoughts if that makes sense :)

gvenzl:
Hm, I would like to avoid removing `npm install` altogether to keep it consistent with the rest of the databases, if that makes sense.
I have added the prerequisites in parenthesis instead, hope that is ok.

crandmck:
Works for me, thanks!

gvenzl:
Thanks a lot for the merge, @crandmck.

Just for my understanding: How do we also get this into the pages for other languages?

Thx,

crandmck:
You can open a PR to update a translated page, for example http://expressjs.com/de/guide/database-integration.html -- source is https://github.com/expressjs/expressjs.com/blob/gh-pages/de/guide/database-integration.md.

In general, I'd prefer one PR per language, and if possible it would be ideal if the translation includes all changes to the English page since the last update (so that we can keep track of what's been translated more easily).  Thanks!

gvenzl:
Ok, great! I'm just double checking because the [CONTRIBUTING.md](https://github.com/expressjs/expressjs.com/blob/gh-pages/CONTRIBUTING.md) says:

> IMPORTANT: We have professional translations of the Express documentation into:
> 
> German
> French
> Italian
> Spanish
> Brazilian Portuguese
> Japanese
> Simplified Chinese
> Traditional Chinese
> Russian
> Korean
> Therefore we can no longer accept community translations for these languages, except for corrections to the existing translations.

However, I'm happy to provide translations for German, French, Italian and Spanish.

crandmck:
Ah, yes....   
> except for corrections to the existing translations.

Should really say "...except for corrections _and additions_ to the existing translations."
(I'll update that to clarify).

Since this is an addition, you're welcome to submit a PR for any of those languages.  It would involve looking at the existing translation and translating any new or changed English content.

Thanks!!

gvenzl:
Cool, thanks!

PR #845, opened 2017-08-22
Add French language example for Oracle driver

gvenzl:


gvenzl:
Close PR due to mistake: This includes German and French.

PR #850, opened 2017-09-02
Explicate optional parameter for parser methods

govindrai:
Here is a snippet from the [body-parser middleware documentation](https://expressjs.com/en/resources/middleware/body-parser.html):

> #  **bodyParser.json(options)**
> 
> Returns middleware that only parses json and only looks at requests where the Content-Type header matches the type option. This parser accepts any Unicode encoding of the body and supports automatic inflation of gzip and deflate encodings.
> 
> A new body object containing the parsed data is populated on the request object after the middleware (i.e. req.body).
> 
> **Options**
> 
> The json function takes an option `options` object that may contain any of the following keys:

The `options` objects that the JSON parser takes is optional, so shouldn't the snippet actually look like this?:

> #  bodyParser.json(**[**options**]**)
> 
> Returns middleware that only parses json and only looks at requests where the Content-Type header matches the type option. This parser accepts any Unicode encoding of the body and supports automatic inflation of gzip and deflate encodings.
> 
> A new body object containing the parsed data is populated on the request object after the middleware (i.e. req.body).
> 
> **Options**
> 
> The json function takes an **optional** `options` object that may contain any of the following keys:

Here are the two changes proposed
1. Add brackets around `options` parameter in the the method signature to signify optional `options` object
2. Change option `options` object to optional `options` object (last sentence)

If so, similar changes can be applied to all four parsers (JSON, Raw, Text, URL-encoded).

### DIFF
--- a/en/resources/middleware/body-parser.md
+++ b/en/resources/middleware/body-parser.md
@@ -68,7 +68,7 @@ or an error occurred.
 The various errors returned by this module are described in the
 [errors section](#errors).
 
-### bodyParser.json(options)
+### bodyParser.json([options])
 
 Returns middleware that only parses `json` and only looks at requests where
 the `Content-Type` header matches the `type` option. This parser accepts any
@@ -80,7 +80,7 @@ object after the middleware (i.e. `req.body`).
 
 #### Options
 
-The `json` function takes an option `options` object that may contain any of
+The `json` function takes an optional `options` object that may contain any of
 the following keys:
 
 ##### inflate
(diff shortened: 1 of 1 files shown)


dougwilson:
We use a script to grab the README from the body-parser module directly. Please file an issue (or even better, a PR) to that module and it'll get picked up here :)

govindrai:
@dougwilson I'm sorry. totally misunderstood your directions. body-parser has its own repo... sorry for the troubles.

dougwilson:
It's no problem. I figured as soon as you made the issue, you started on the PR and didn't see my issue comment; happens all the time.

PR #854, opened 2017-09-09
Fix broken anchor links

niklashigi:
The `#` was missing, so the links did not jump to the respective anchor, but led to a 404 page.


crandmck:
Doh!  Thanks!

PR #855, opened 2017-09-09
Update body-parser.md

ghost:


ghost:
After creating this PR I realized that many other pages in the middleware section have this closing `</div>` tag, is it a typo or am I missing something?

dougwilson:
All these middleware files are generated with https://github.com/expressjs/expressjs.com/blob/gh-pages/get-readmes.sh so editing one will just be overwritten. Can you edit the script instead?

ghost:
I'll try my best :)
I checked the script, the issue is within the `writefile_mw` function.

The opening div tags are not written to the .md files, but the closing div tags are.
`echo '<div id="page-doc" markdown="1">' >> $FILE` this line seems to not have worked for some reason, 

I tried the script locally and both the opening and closing tags are written to the file.

Is it okay to just remove the div tag `echo` lines from the script? I still dunno why we need them. I checked and they were added last Feb. 

image

dougwilson:
It looks like commit 8a103159cb58b61f89370e5ebcf11b361fff4a8d is what caused the issue. It looks like the tags were likely being removed by hand and that one was missed. It should have also updated the script to no longer write those tags, but that was missed by accident too.

dougwilson:
Looks like we should update the script to remove the div tags to align with the changes of that commit and then run the script, because the div tags were missed on several files.

crandmck:
Hey, sorry, my bad.  I made the change in 8a103159cb58b61f89370e5ebcf11b361fff4a8d to move the enclosing `div` into the layout template for most pages, but forgot about dealing with the middleware pages.  I'll update the script--I actually have a much cleaner implementation that I've been meaning to use. 

Opened #856 to track.

PR #858, opened 2017-09-13
Fix req.is() return value

davewasmer:
The type-is docs say the return value is `string | false`. When a matching content type is found, that is the value returned, not a Boolean `true`

### DIFF
--- a/_includes/api/en/4x/req-is.md
+++ b/_includes/api/en/4x/req-is.md
@@ -6,16 +6,14 @@ Returns `false` otherwise.
 
 ```js
 // With Content-Type: text/html; charset=utf-8
-req.is('html');
-req.is('text/html');
-req.is('text/*');
-// => 'html'
+req.is('html');       // => 'html'
+req.is('text/html');  // => 'text/html'
+req.is('text/*');     // => 'text/*'
 
 // When Content-Type is application/json
-req.is('json');
-req.is('application/json');
-req.is('application/*');
-// => 'html'
+req.is('json');              // => 'json'
+req.is('application/json');  // => 'application/json'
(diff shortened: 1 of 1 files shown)


_includes/api/en/4x/req-is.md
```diff
@@ -9,13 +9,13 @@ Returns `false` otherwise.
 req.is('html');
 req.is('text/html');
 req.is('text/*');
-// => true
+// => 'html'
```
dougwilson: Not all of these return `'html'`, only the first one does. Each of them return their input string.

_includes/api/en/4x/req-is.md
```diff
@@ -9,13 +9,13 @@ Returns `false` otherwise.
...
 
 // When Content-Type is application/json
 req.is('json');
 req.is('application/json');
 req.is('application/*');
-// => true
+// => 'html'
```
dougwilson: Of course, none of these will return the string `'html'`, but rather they will all return their own input string.

dougwilson (changes requested):
You are absolutely right! Never noticed this mistake in the docs. The examples just need to be adjusted.

_includes/api/en/4x/req-is.md
davewasmer: Ah, I didn't realize that (didn't look closely enough at `type-is`)

_includes/api/en/4x/req-is.md
davewasmer: 🤦‍♂️ yea that was a fat-finger mistake ...

dougwilson (approved):
LGTM

crandmck:
Thanks!

PR #859, opened 2017-09-14
Overriding sugar methods

sandizapar:


crandmck:
This branch is very outdated and probably needs to be rebased ... in any case to land it we'd want to merge with gh-pages branch.
I don't think this is how we want to proceed.

PR #869, opened 2017-10-15
Fix sample Redis

ilyar:
ReferenceError: redis is not defined

### DIFF
--- a/en/guide/database-integration.md
+++ b/en/guide/database-integration.md
@@ -319,7 +319,8 @@ $ npm install redis
 ### Example
 
 ```js
-var client = require('redis').createClient()
+var redis = require('redis')
+var client = redis.createClient()
 
 client.on('error', function (err) {
   console.log('Error ' + err)


crandmck:
This seems like the exact same thing.  Why would it work and not the original?

ilyar:
@crandmck  ReferenceError: redis is not defined in

caiobrentano:
Hi @ilyar. I think you tagged the wrong person! :)

en/guide/database-integration.md
```diff
@@ -319,7 +319,8 @@ $ npm install redis
 ### Example
 
 ```js
-var client = require('redis').createClient()
+var redis = require("redis")
```
dougwilson: Use single quotes here for https://standardjs.com/

dougwilson (changes requested):
Yea, this is correct, since below the direct change, there is use of the variable `redis` for `redis.print`, so the whole example seems to expect a `redis` variable.

ilyar (commented):
up

crandmck:
Oh, I see.    Sorry I didn't look at the rest of the code sample.. sorry.

OK LGTM, thanks!  I'll land it since you addressed Doug's review comment.

PR #873, opened 2017-10-26
Mention that req.params are being URI-decoded

pitamar:
From looking at the [source code](https://github.com/expressjs/express/blob/9722202df964bfbfc0f579e4baeb5a4e1b43b344/lib/router/layer.js#L110) and doing some testing I saw that req.params values are being URI-decoded via `decodeURIComponent`.
I've tested for paths with regular parameters, paths containing wild card and regular expression paths.
As far as I can tell there is no case where they aren't being decoded. Am I correct?
I think this should be mentioned in the documentation since it is not necessarily obvious.
In [Play framework for example](https://www.playframework.com/documentation/2.6.x/ScalaRouting#The-URI-pattern) they are decoded in some cases but not in others and in either case it is explicitly stated.


crandmck:
LGTM, but would want @dougwilson to confirm the technical content.

dougwilson:
Technical aspect is correct 👍

crandmck:
Thanks!

PR #874, opened 2017-11-02
Update hello-world.md

cagataycali:
Looks better :+1:


crandmck:
I think this is OK, but since we don't use arrow functions anywhere else in the docs, I wonder if this could potentially cause some confusion, for example for someone not familiar with them?

Do you have an opinion or preference @dougwilson ?

dougwilson:
I'm not too concerned with thr arrow functions, but I am concerned with the line

    const app = require('express')()

The immediate invoke has in the past caused a lot of users confusion, from (1) not realizing there was a () on the end and trying to run code without it and (2) get confused as soon as they tried to use express.static or express.Router

I think if that line is back to two separate lines the arrow function changes itself is fine.

cagataycali:
I agree with you guys, seperated back in two lines.

crandmck:
Great, works for me... Thanks @cagataycali

PR #883, opened 2017-11-24
tiny change on zh-cn generator.md

nfer:
In Simplified Chinese, we use "输入" but not the "装入" for the "input" action.


hacksparrow:
Thanks @nfer

PR #885, opened 2017-11-29
Fix link to framework page in zh-cn and add frameworks page for zh-cn…

ibelem:
Fix link to framework page in zh-cn and add frameworks page for zh-cn/resources/frameworks.html


hacksparrow:
Thanks @ibelem

PR #887, opened 2017-12-09
Update routing.md

dwkrueger2:
I felt that the instructions did not provide a very good mental model of how the express application worked.  I could only figure it out by stepping through the code manually.  I think the proposed edit will help clarify the whole 'routing' concept that express utilizes and makes understanding the express program quicker and thus easier to use.


crandmck:
Thanks for the PR.  There are multiple issues with the writing (grammar, wording, formatting, etc.) but rather than try to resolve them here, I'm just going to land this PR and then make the editorial corrections directly, since that will just be much easier and quicker.

PR #902, opened 2018-01-27
docs(healthz): add advanced topic for graceful shutdown

gergelyke:


crandmck:
While the content seems valid, I don't think it belongs in the Express documentation, since it's not specific to Express _per se_.  What do you think @dougwilson ?

dougwilson:
Yea, it can be either way (included or not included), it's hard to really say. Yes, the underlying server is the actual Node.js server instance, and closing it is strictly how one would close a Node.js HTTP server, independent of Express.js. That said, talking about closing the server is useful in the general sense as an aspect of an "Express application".

I'm a "+0" as the Node.js core folk would say.

Regarding the content itself, if it were included, the first code example is going to be quite confusing, with it not having much context, and what one would assume would be the `server` variable won't even work (because the Node.js `server` object doesn't have any kind of Promise API).

_includes/header/header-en.html
```diff
@@ -1,156 +1,316 @@
...
-        -->

-      </ul>

-  </div>

-</header>

+<header>
+
```
dougwilson: Not sure what is going on in this change, adding a blank line between every single original line?

en/advanced/healthcheck-graceful-shutdown.md
```diff
@@ -0,0 +1,78 @@
...
+   healthChecks: {
+    '/healthcheck': onHealthCheck,
+  },
+  onSignal
+});
+
+server.listen(3000);
+```
+
+You can read the original announcement blogpost for terminus here: [https://nemethgergely.com/nodejs-healthcheck-graceful-shutdown/](https://nemethgergely.com/nodejs-healthcheck-graceful-shutdown/).
```
dougwilson: If this is included, may want to reword this text so it sounds less like an advertisement.

en/advanced/healthcheck-graceful-shutdown.md
```diff
@@ -0,0 +1,78 @@
...
+* liveness is used by the kubelet to know when to restart a container,
+* readiness is used by the kubelet to know when a container is ready to start accepting traffic - when a pod is not ready, it is removed from the Service load balancers.
+
+## Third-party solution: terminus
+
+[terminus](github.com/godaddy/terminus) is an open-source project, which adds health checks and graceful shutdown to your applications - to save you from the boilerplate code you would add otherwise. You only have to provide the cleanup logic for graceful shutdowns, and the health check logic for health checks, all the rest is handled by it.
+
+```js
+const http = require('http');
+const { Client } = require('pg');
```
dougwilson: I wonder if the example can just use variables and code / comments to generally speak of "closing a db connection" instead of the example being specific to `pg`.

gergelyke: sure, going to amend the PR

en/advanced/healthcheck-graceful-shutdown.md
gergelyke: I am happy to remove it

_includes/header/header-en.html
gergelyke: Hm, that's a good question - will look into this too

gergelyke:
@dougwilson @crandmck addressed your comments. The reason I've sent this PR, is that I have the feeling that most production Node.js applications are not prepared for this, and I'd like to raise awareness. Being it in the official Express docs would help a lot to make it happen.

Examples from other communities shows, that the framework docs can be a good place for it: https://github.com/gin-gonic/gin#graceful-restart-or-stop

Let me know if it looks good this way, or if it makes sense to include it. Thanks you!

crandmck:
OK @gergelyke thanks for your edits.  I don't have a serious objection, because I can see that it's important information, but I would like to see if anyone else in the @expressjs/express-tc has an opinion one way or the other.

gergelyke:
Thanks @crandmck - let me know if you need anything else from me!

gergelyke:
@crandmck any update on this?

crandmck:
@gergelyke Sorry for the delay.  I'll try to take a final look and land this in the next couple days.

crandmck:
@gergelyke Please remove the extra blank lines in `header-en.html`, as Doug noted previously.  

Otherwise, I think it looks OK; I'll probably make some minor wording/grammatical edits after landing (it's just easier that way)...

Any further comments @dougwilson ?

dougwilson:
That's it from me 👍

gergelyke:
@crandmck thanks, should be fixed now!

crandmck:
OK I landed this: http://expressjs.com/en/advanced/healthcheck-graceful-shutdown.html

@gergelyke  you'll see I edited it as I said I would... 

In addition to basic edits and clarification, I noticed that the example code has a call to `terminus(...)` but it there was no corresponding `require()` statement at the top, so I added that; specifically:
```
const terminus = require('@godaddy/terminus');
````

If that's NOT correct, please let me know and I'll fix it, but based on the example at <https://github.com/godaddy/terminus/blob/master/example/index.js> I think it's correct.

gergelyke:
thanks Rand! that's look great!

PR #903, opened 2018-02-02
Remove redirect_from in Thai translation to correct redirects back to English pages

sehrope:
The Thai translation that was added a few weeks ago includes `redirect_from: ...` clauses. These are overriding the English pages so that by default pages like http://expressjs.com/api.html are redirecting to the Thai page rather than the English one (click it to try it out). I noticed this earlier today when a link from a Google search page sent me to the Thai translation.

PR removes the `redirect_from: ...` lines as I'm assuming the English localization is meant to be the default.

### DIFF
--- a/th/3x/api.md
+++ b/th/3x/api.md
@@ -3,7 +3,6 @@ layout: 3x-api
 title: Express 3.x - API Reference
 menu: api
 lang: th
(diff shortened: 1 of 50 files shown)


crandmck:
Yikes, how did that slip through?!  Thanks very much for catching it!

PR #909, opened 2018-03-09
Don't use the term "route name".

anielsen:
The term "route name" is only used in one place so it was a bit confusing for a newbie. I propose it's simply remove to match the rest of the docs.


crandmck:
Good suggestion, thanks.

PR #920, opened 2018-04-04
Define the `wrap` function for the async example in Best Practices page

davidjb:
This example came from https://strongloop.com/strongblog/async-error-handling-expressjs-es7-promises-generators/ but the definition of `wrap()` wasn't replicated here.  This adds the definition to avoid confusion.

### DIFF
--- a/en/advanced/best-practice-performance.md
+++ b/en/advanced/best-practice-performance.md
@@ -150,6 +150,8 @@ However, there are two caveats:
 2.  Event emitters (like streams) can still cause uncaught exceptions. So make sure you are handling the error event properly; for example:
 
 ```js
+const wrap = fn => (...args) => fn(...args).catch(args[2])
+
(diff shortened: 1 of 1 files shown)


crandmck:
Thanks for the PR @davidjb 

Good suggestion, but I'd prefer to add the information outside of the code block.  Something like this right after the code example:
```
The `wrap()` function is a wrapper that catches rejected promises and calls `next()`.  
For details, see [https://strongloop.com/strongblog/async-error-handling-expressjs-es7-promises-generators/#cleaner-code-with-generators](Asynchronous 
Error Handling in Express with Promises, Generators and ES7).
```

davidjb:
@crandmck Thanks for the suggestion; updated.  I removed the repeated link to https://strongloop.com/strongblog/async-error-handling-expressjs-es7-promises-generators/ as it would have been there twice in the same text block.

Issue #915, opened 2018-03-21
Error handling guide


pluby:
While the error handling guide is great at describing how to write error handling middleware, we spend most of their time writing ordinary route handlers and middleware. However, the number one issue we find during code reviews, and when debugging, is a failure to pass asynchronous errors on to Express in the first place. The guide only covers this a little at the end.

I'd like to suggest that the error handling guide be changed to focus initially on providing error handling guidance to developers writing routing and middleware, explaining with examples how Express catches errors in the first place using the default handler. Then follow with how to write error handlers themselves (the content that is already there). It will make the page more complete and a lot more useful to everyday development.

I'm happy to make a start on this and make a pull request unless there are objections or other ideas.

Thanks

crandmck:
Sounds like a good suggestion.  Please do open a PR.

dougwilson:
PR was merged.

PR #926, opened 2018-04-13
Add information about catching errors to the error handling guide.

pluby:
Resolves issue #915.

I'm not sure if I've done enough here, or what the situation regarding translation would be, but hopefully this is a step towards helping those new to Express with their error handling.

I've added information about how to ensure Express gets hold of errors, which I think is the most basic and important information new developers need to get started when using Express. I put this first because unless new developers get this bit right, the rest of the guide is wasted. I then swapped the existing sections around, as I thought it unlikely the developers would dive in to writing their own error handlers, but they should at least be aware of the default error handler and how to configure it.

### DIFF
--- a/en/guide/error-handling.md
+++ b/en/guide/error-handling.md
@@ -5,7 +5,167 @@ menu: guide
 lang: en
 redirect_from: "/guide/error-handling.html"
 ---
-# Error handling
+# Error Handling
+
+**Error Handling** refers to how Express catches and processes errors that
+occur both synchronously and asynchronously. Express comes with a default error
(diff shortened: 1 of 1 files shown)


crandmck:
@dougwilson This is a fairly lengthy addition, and I'd like to get your eyes on it before landing, please.

dougwilson:
I read through all the new content and it looks good from the technical standpoint 👍

crandmck:
Thanks @dougwilson .  I'll land this and make any copy edits after the fact.
Thanks for the contribution, @pluby

PR #945, opened 2018-07-06
Update using-middleware.md

danplisetsky:
Substitute `res.send` for `res.render` in the last example for Application-level middleware. The examples before all use `res.send`, so `res.render` creates unnecessary confusion, since it hasn't been introduced in the documentation just yet (it will be in the next guide)


dougwilson (approved):
Makes sense to me 👍

Issue #562, opened 2016-01-21
"Using template engines" example uses jade != instead of =


tunniclm:
The 2nd code  example showing a jade template in:

http://expressjs.com/en/guide/using-template-engines.html

..uses jade unescaped string interpolation (!=) when it should probably teach (particularly for beginners) to use escaped string interpolation (=).

Jade string interpolation is documented here: http://jade-lang.com/reference/interpolation/

jakerella:
That's a pretty easy PR to submit. ;)
But nice catch by the community.

crandmck:
@tunniclm I made the suggested change.  As I was reviewing that article, I noticed it lacked some context and cross-references, so I made a few other improvements.  Would you please review the PR?

tunniclm:
@crandmck LGTM!

crandmck:
Fixed by #563, thanks for the input!

PR #957, opened 2018-07-29
Use pug = instead of != in "Using template engines" example

tyage:
I think using `!=` is not secure and it is not good to use in the example. 

#562 reported this issue and it is fixed for english documentation in #563 .
But in other languages, it is not fixed.
e.g. http://expressjs.com/ja/guide/using-template-engines.html

So I fixed this issue for other languages.


dougwilson (approved):
Thanks!

PR #964, opened 2018-08-08
suggesting es6.

adescode:
Since const and Arrow function is implemented, template literals would be nice.

### DIFF
--- a/en/starter/hello-world.md
+++ b/en/starter/hello-world.md
@@ -16,10 +16,11 @@ Embedded below is essentially the simplest Express app you can create. It is a s
 <div id="hello-example"><pre><code class="language-js">
 const express = require('express')
 const app = express()
+const port = 3000
 
 app.get('/', (req, res) => res.send('Hello World!'))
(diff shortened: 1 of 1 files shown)


crandmck:
I'm ok with this, but now that I think about it, I wonder if using ES6 syntax in "Hello World" is really a good idea.  Has it become universal knowledge?  Or will it just muddy the waters...?  I see we landed the change to ES6 syntax in this page last Nov.

dougwilson:
Since the example looks like it is already es6 i guess it's fine. Only one nit i commented on if anyone wants to adjust it.

en/starter/hello-world.md
```diff
@@ -16,10 +16,11 @@ Embedded below is essentially the simplest Express app you can create. It is a s
 <div id="hello-example"><pre><code class="language-js">
 const express = require('express')
 const app = express()
+const port = 3000;
```
dougwilson: No semicolon on this line

Issue #961, opened 2018-08-07
Right aligned text in 'API reference' menu hard to read.


craigdanj:
While it may look nice i feel that this right aligned menu would be easier to read it all the items were left aligned. I'm currently learning express and noticed that this was a problem for me. Just thought I'd mention.

Just my two cents but this would help UX in my opinion.

For eg:

It would be easier to read if it looked like this:

```
express()
Application
Request
Properties
req.app
req.baseUrl
req.body
req.cookies
req.fresh
req.hostname
req.ip
req.ips
req.method
req.originalUrl
req.params
req.path
req.protocol
req.query
req.route
req.secure
req.signedCookies
req.stale
req.subdomains
req.xhr
Methods
req.accepts()
req.acceptsCharsets()
req.acceptsEncodings()
req.acceptsLanguages()
req.get()
req.is()
req.param()
req.range()
Response
Router
```

crandmck:
It's a reasonable suggestion.  Does anyone else have an opinion...?
@craigdanj You are welcome to open a PR with a fix if you want.

dougwilson:
I don't have an opinion either way. The left hand side is standard for left to right languages like English anyway.

craigdanj:
@crandmck  I have submitted a pull request for the same. Kindly let me know if anything else needs to be done. I've tested on mobile too to make sure it works.

--------------------------------------------------
**FYI:**  I also noticed that with the previous implementation (with the menu on the right), the menu overlaps the content in some places on smaller screens (screenshot attached below). This too has been fixed in this PR of mine since they both pertain to the menu.


![screen shot 2018-08-15 at 5 30 09 pm](https://user-images.githubusercontent.com/3023019/44147672-9de4f5a4-a0b2-11e8-9ed3-babf6e09d197.png)

dougwilson:
Right align change was merged.

PR #967, opened 2018-08-15
CSS tweaked to move fixed menu from the right to the left of the page.

craigdanj:
This is in reference to bug #961. Menu moved to the right on the API reference page. This fix was tested on mobile views as well and works.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -139,7 +139,7 @@ strong, th {
 .content {
   margin: 90px 3% 7%;
   max-width: 1090px;
-  padding-right: 140px;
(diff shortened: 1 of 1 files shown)


crandmck:
@craigdanj Thanks for the PR.  This looks good to me, but I'd like to see if anyone else has an opinion on the final result.... e.g. @dougwilson. 

Here is what it looks like (for example):

crandmck:
OK since no one has commented, I'm going to assume there are no objections.  I like the way this change looks.   We can always go back if there are meaningful objections.

Issue #966, opened 2018-08-15
Request to add an express middleware


ameerthehacker:
I have developed a middleware for express which would remove all unused css styles from the template before rendering which would greatly increase the performance of websites written in express.

I would love to add this to expressjs.com [middleware page](https://expressjs.com/en/resources/middleware.html). If this is okay shall I create a PR corresponding to this?

Please check out my middleware here
https://github.com/ameerthehacker/express-slick-css

crandmck:
Hi @ameerthehacker The third-party middleware listed on http://expressjs.com/en/resources/middleware.html are mostly packages that have been around for a while and have lots of weekly downloads on npm. While we don't have any explicit requirements to be listed there, yours is very new--so my two cents would be to wait a little while.  That being said, if you want to open a PR, we can see what other project members think.

ameerthehacker:
@crandmck thanks for the reply. The main reason I want it to be listed on the site is to take this to a huge audience and I believe my middleware can be of use to many. Since mine is new I'm looking for ways to take this to the right people who might be interested in it. Thanks!

ryhinchey:
closing this issue as the PR was closed

PR #968, opened 2018-08-17
Add express-slick-css to middleware.md

ameerthehacker:
I have developed a middleware called [https://github.com/ameerthehacker/express-slick-css](express-slick-css). It removes all the unused CSS rules before sending the response as a page. It is relatively new yet I believe this is one way I can take this to a huge audience of expressjs If you people feel my middleware is up to it!

### DIFF
--- a/en/resources/middleware.md
+++ b/en/resources/middleware.md
@@ -52,6 +52,6 @@ These are some additional popular middleware modules.
 | [static-expiry](https://github.com/paulwalker/connect-static-expiry) | Fingerprint URLs or caching headers for static assets.|
 | [view-helpers](https://github.com/madhums/node-view-helpers) | Common helper methods for views.|
 | [sriracha-admin](https://github.com/hdngr/siracha) | Dynamically generate an admin site for Mongoose. |
-| [express-slick-css](https://github.com/ameerthehacker/express-slick-css) | Remove unwanted css rules before sending response |
+| [express-slick-css](https://github.com/ameerthehacker/express-slick-css) | Remove unused css rules before sending response |
 
 For more middleware modules, see [http-framework](https://github.com/Raynos/http-framework#modules).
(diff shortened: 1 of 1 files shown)


crandmck:
I have no specific objections to this addition, but see discussion in #966. 
My main concern is that we can't possibly list _all_ the middleware that exists, so where do we draw the line?

@wesleytodd @dougwilson PTAL

dougwilson:
Without there being a formal definition of what should be in there and what should not, I have no objection to adding it in.

ameerthehacker:
Thanks @dougwilson

ameerthehacker:
@crandmck any updates on this PR?

wesleytodd:
I would like to think on this for a bit.  Adding to the website is an explicit recommendation to the community, and I think we should be judicious about that.  As it stands, this page lists some middleware which should be audited IMO.

crandmck:
I agree with @wesleytodd.  I think we should err on the side of caution... I don't recall where that set of middleware came from (probably some README), but I think it might be worth reviewing it again, and we could include this PR as a potential addition.

ameerthehacker:
@crandmck If my understanding is right my middleware will go through a reviewing process and then this PR will be considered for merging. If so how the reviewing process is going to be?

crandmck:
>  If so how the reviewing process is going to be?

@ameerthehacker It's a fair question.  We haven't established one yet.  

If and when we do establish this process, we should review all the "Additional middleware" (i.e. non-expressjs-maintained modules) listed in https://expressjs.com/en/resources/middleware.html so it's an even playing field.
 
I suggest we keep the process simple and lightweight.  Perhaps simply a discussion issue? @wesleytodd Thoughts?

ameerthehacker:
Hey @crandmck @wesleytodd any updates on this discussion?

crandmck:
Sorry for the delay.   We should probably have some kind of criteria or process for inclusion on the site...  Unfortunately, establishing it will take some time.  We're going to get more requests like this, and in fact we already did -- see #981.  

I started #988 to discuss this issue more generally. 

@ameerthehacker I'm afraid resolution of this particular PR will have to wait until the more general issue is resolved.  Thanks for your patience.

dougwilson:
So with the discussion here being controversial, and I guess this section is noted "These are some additional popular middleware modules" while this module as (currently) zero weekly downloads according to npm, I guess I won't land it 😢 . It does seem well made, though.

PR #969, opened 2018-08-17
Update best-practice-security.md

animir:
fix broken strong loop link. replace outdated `express-limiter` with race conditions and no support to `rate-limiter-flexible`


crandmck:
Thanks!

PR #971, opened 2018-08-23
Fixed: ES2015 compatible Sample Code

silverskyvicto:
Hi.

Sample code of 'Hello world' page is written in ES2015. So I fixed it.


dougwilson:
var is es2015

Issue #986, opened 2018-10-11
Image for Facebook share


gonenduk:
I share links to express.js websites often in my facebook groups, and it's a shame there is no image associated with the shared link. It onlt shows text without any image.
Even adding the logo would be enough and make the share much more noticeable.
Attaching a sample share I just did today for the new released version.
![share sample](https://user-images.githubusercontent.com/1597854/46794214-0e8d0200-cd50-11e8-9efb-c32e4e984c63.png)

dougwilson:
Yea, that sounds like a good idea to me. Any idea what changes would need to be made on the site to get an image to show up? If not, we can research it, but just figured you may happen to know.

gonenduk:
Sure, it's quite easy. In each HTML page, you should first define the og (open graph) alias in the html tag:
`<html prefix="og: http://ogp.me/ns#">`
and then add in the head a meta data link to the image (best if 360 on 360 pixels):
`<meta name="og:image" content="https://expressjs.com/path/to/logo.png">`
I would do the change myself but I do not see any image file with a logo of the project.
After you make the changes, you can check them using the facebook sharing debugger in this link:
[Facebook Sharing Debugger](https://developers.facebook.com/tools/debug/sharing/?q=expressjs.com)

dougwilson:
That's awesome and super helpful! Express does not have a logo, which is why there is no logo image in this repo.

gonenduk:
Then maybe using just the 'ex' image as you use on GitHub, or the complete express work, or even the image I attach or something similar.

<details>
<summary>logo example</summary>

![express_js](https://user-images.githubusercontent.com/1597854/46839643-96fab980-cdc7-11e8-9507-1b218d6507f2.png)

</details>

dougwilson:
Well, who owns the trademark on that js logo? We cannot of course violate any outstanding US trademarks.

gonenduk:
I'm not sure you cannot use the js logo, but what about the Express word alone as an image?
I can do a pull request with this image and it can always be replaced later.
![express](https://user-images.githubusercontent.com/1597854/46839911-03c28380-cdc9-11e8-858f-8220f2474fe4.png)

dougwilson:
> I'm not sure you cannot use the js logo

Well, without explicit permission, the risk is a cease and dissect order, but can be much harsher than that at least in the US (which is where I reside).

> what about the Express word alone as an image?

Seems fine to me, at least.

gonenduk:
[...]

PR #987, opened 2018-10-12
Facebook share image

gonenduk:
As discussed in issue #986, adding an image to facebook share.
This will make sharing pages of the website on facebook much more visible on the feed.
The image used is the word "Express" as used on the homepage of the website.


crandmck:
LGTM.  Thanks @gonenduk

PR #998, opened 2018-11-05
Update writing-middleware.md

shivgodi7:
My observation, in case I am not correct please ignore, thanks.
When you call req.requestTime, you are not calling the middleware function requestTime, rather you are calling Date.now() which is assigned to req.requestTime.
Here the property added to req happens to be with the same name as the middleware function requestTime.


dougwilson (approved):

crandmck:
Thanks!

PR #1001, opened 2018-11-09
State functionality of 'next(err)' upfront.

bvaisakh:
Functionality of next(err) is stated towards of the end of the document. I felt it should be stated upfront. And where it is stated, there is a mention of some error handling middleware code snippet which do not exists.

If you pass anything to the next() function (except the string 'route'), Express regards the current request as being an error and will skip any remaining non-error handling routing and middleware functions. If you want to handle that error, you’ll have to create an error-handling route **as described in the next section.**

### DIFF
--- a/en/guide/error-handling.md
+++ b/en/guide/error-handling.md
@@ -43,6 +43,8 @@ app.get("/", function (req, res, next) {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


crandmck:
I think it makes sense to move this up, since it's important to understand up front.  This whole page could probably be improved further, but this is a step in the right direction.  Thanks!

PR #1007, opened 2018-12-12
Update using-middleware.md

### LINKED ISSUE #951: Update to documentation with examples of using middleware as second param on the router
In digging into the documentation a bit, i noticed there is no examples of using middleware as a second param on the router.

https://expressjs.com/en/4x/api.html#router.METHOD

For example no examples with something like this:

```js
router.get('/foo', someMiddlewareFn, callbackFn)
```

Can you please update the docs with some examples like this?

toymachiner62:
Fixes #951

### DIFF
--- a/en/guide/using-middleware.md
+++ b/en/guide/using-middleware.md
@@ -214,25 +214,6 @@ app.use('/admin', router, function (req, res) {
 })
 ```
 
-This example shows an array with a middleware sub-stack that handles GET requests to the `/user/:id` path
-
-```js
-function logOriginalUrl (req, res, next) {
-  console.log('Request URL:', req.originalUrl)
-  next()
-} 
-
-function logMethod(req, res, next) {
-  console.log('Request Type:', req.method)
-  next()
-}
-
-var logStuff = [logOriginalUrl, logMethod]
-router.get('/user/:id', logStuff, function (req, res, next) {
-  res.send('User Info')
-})
-```
-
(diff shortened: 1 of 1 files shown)


wesleytodd:
Is the redundancy in the example intentional?  As I read that page, I am unsure repeating the same example is necessary.  Maybe it would be sufficient to just mention that you can also use arrays of middleware in Router level middleware?

toymachiner62:
The redundancy was simply me trying to follow what was already there. I thought that much of the application vs router was already redundant so i was trying to follow the existing convention.

wesleytodd:
Yeah, I agree what is there repeats itself a bit and it makes sense why you would follow.  If you are interested in cleaning that up as well I think it would be fine to do here.  If not, maybe just break from the existing stuff for this and we can clean up the rest later?

crandmck:
> maybe just break from the existing stuff for this and we can clean up the rest later?

Makes sense to me...

toymachiner62:
I'm not following what you mean by:
> maybe just break from the existing stuff for this and we can clean up the rest later?

Can you elaborate?

crandmck:
I don't want to speak for @wesleytodd but I interpreted it to mean: eliminate the redundancy in the example you added, but leave the rest (i.e. the previously existing material) "as is" and we can eliminate the redundancies there later, as needed....

toymachiner62:
Ok i updated it

dougwilson (approved):

crandmck:
Thanks @toymachiner62

Issue #1011, opened 2019-01-12
express.json([options]) documentation needs example


toymachiner62:
I was looking at the express docs again the other day and noticed express.json([options]) on this page https://expressjs.com/en/4x/api.html and I see it says that this is based on body-parser. I've used body-parser many times before, but curious if express.json() is a built in replacement for an external dependency?

In either case there is no example on how to use it. Can you please add one? Is it used the exact same way as body-parser?

wesleytodd:
Hey @toymachiner62, it is in fact just exposing `body-parser` directly (as in `express.json = bodyParser.json` [see here](https://github.com/expressjs/express/blob/master/lib/express.js#L78)).  I think the goal is to not duplicate any documentation so that we don't have to maintain it in multiple places.  That being said, there are already a bunch of doc's there coupled to the interface of `body-parser`, so if you wanted to submit a PR with an example, I think it would be helpful to have.

toymachiner62:
Struggling to find the file in the repo that corresponds to this page. help.

wesleytodd:
https://github.com/expressjs/expressjs.com/blob/gh-pages/_includes/api/en/4x/express.json.md

ryhinchey:
Closing this issue as it looks like this content has been added to the site

PR #1012, opened 2019-01-14
Update express.json.md

toymachiner62:


toymachiner62:
Fixes #1011

dougwilson (changes requested):
The wording that was changed at the top here is the same wording used for all the build-in middlewares: express.json, express.urlencoded, and express.static. They should all be using the same wording / linkage so they are not all slightly different when they should have the same underlying meaning.

I'm not sure if this is the right wording here, but at the very least, this should keep the wording for these three pieces in sync in the PR.

toymachiner62:
While i don't necessarily agree with the current language (mainly since `express.json` is not based on `body-parser`, but rather IS `body-parser`, I respect your wishes and changed it back.

wesleytodd:
I think you both are agreeing with subtle variations.  I think @dougwilson intended to say that if you want to improve the wording (as you say, it IS `body-parser`, which IMO is a good change), then you should do it in all three places `.json`, `.urlencoded` and `.static`.  I think a reasonable PR would be to change all three in one go.

dougwilson:
Exactly.

PR #1014, opened 2019-01-18
Update SQL server database integration example

joshgrib:
This has fallen out of sync with the [tedious documentation](http://tediousjs.github.io/tedious/getting-started.html).

Running it with the current setup results in the warning:
```
tedious deprecated The "config.userName" property is deprecated and future tedious versions will no longer support it. Please switch to using the new "config.authentication" property instead. sql.js:11:18
tedious deprecated The "config.password" property is deprecated and future tedious versions will no longer support it. Please switch to using the new "config.authentication" property instead. sql.js:11:18
```

### DIFF
--- a/en/guide/database-integration.md
+++ b/en/guide/database-integration.md
@@ -377,9 +377,14 @@ var Connection = require('tedious').Connection
 var Request = require('tedious').Request
 
 var config = {
-  userName: 'your_username', // update me
-  password: 'your_password', // update me
-  server: 'localhost'
+  server: 'localhost',
+  authentication: {
+    type: 'default',
+    options: {
+      userName: 'your_username', // update me
+      password: 'your_password' // update me
+    }
+  }
 }
 
 var connection = new Connection(config)


wesleytodd:
Hi @joshgrib, thanks for the update.  While I do think that if we plan to keep this documentation we should update it to match, I **don't** think we should keep it.  Express should not be in the business of providing documentation for other packages.  My preference would be to remove it, not update it.  Would be interested to hear others opinions.

joshgrib:
That definitely makes sense,

joshgrib:
Sorry I just pressed the wrong button!

That makes sense, it was a valuable page for me, but there's also plenty of info online. Thanks!

wesleytodd:
No worries, happens all the time.

I think a reasonable change to start with would to replace that section with a link out to their docs.  What do you think about that?

joshgrib:
I think that'd be good. Then this would be more of a page of suggested packages for connecting to various databases, and these are already listed under the database name on this page. This could be shown succintly as a table as follows:

| Database | Suggested package |
| :-- | :-- |
| Redis | [node_redis](https://github.com/NodeRedis/node_redis) |
| SQL Server | [tedious](https://github.com/tediousjs/tedious) |
| ... | ... |

Thoughts?

crandmck:
In general I agree we shouldn't document other packages here.  I believe @hacksparrow originally added this page.... The thinking was that this was something that many people would find useful.  However, paring it down to something like @joshgrib mentions probably makes sense in the long run.

If we do this, we might also consider moving this page to the **Resources** menu, where we have info on other related/3rd-party packages.

Is there any useful general advice we can give re connecting to DBs that is specific to Express?  Or maybe an example?  That would serve to make this more useful that just a list of packages.  If not, it's probably fine, but just putting it out there.

wesleytodd:
@crandmck there are probably patterns we could illustrate on ways you can integrate *any* DB, but only a small portion would be Express specific.  I would rather start clean and add modern/up-to-date documentation where it would be very helpful than to worry about it in the current state.  Unless someone has time to really dig into this now (which I doubt with the goal of 5.0 doc being worked on soon).

crandmck:
Fair enough. So, IIUC the plan is to replace what's there now with just table or list of links out to the relevant  DB packages....  @joshgrib Would you be willing and able to take a stab at this? :-)

dougwilson (approved):
I think there is a great discussion here that, if someone wants to follow up on, please open an issue / PR. I am at least merging this fix, as it's better to at least not have out-of-date docs on the site

PR #1015, opened 2019-01-27
Update generator.md

Downes:
"set DEBUG=myapp:* & npm start" generates an error on Windows 10 ("The ampersand (&) character is not allowed. ") - "set DEBUG=myapp:* | npm start" is the correct syntax.

### DIFF
--- a/en/starter/generator.md
+++ b/en/starter/generator.md
@@ -81,6 +81,11 @@ On Windows, use this command:
 > set DEBUG=myapp:* & npm start
 ```
 
+On Windows 10, use this command:
+```sh
+> set DEBUG=myapp:* | npm start
+```
+
 Then load `http://localhost:3000/` in your browser to access the app.
 
 The generated app has the following directory structure:


dougwilson:
This is not correct. The command there works just as expected on Windows 10:

image

Downes:
Yes it is correct.

image

I don't care whether you merge it or not, but there is an error in the documentation there.

Downes:
p.s. It may be unique to PowerShell 

See eg. 
https://stackoverflow.com/questions/42095967/how-to-run-an-express-app-with-powershell
https://www.reddit.com/r/node/comments/2fpuyh/cant_get_npm_start_command_to_work/

crandmck:
I don't use Windows any more, so I can't check myself, but it does look like this is an issue with PowerShell. 

@Downes Perhaps you could change your PR to keep the existing command, but state that it is for the Command window (or whatever it's properly called) and add your suggested command for PowerShell in addition.  I know lots of people use PowerShell, so it might be worthwhile to have that as well as what we have now.

dougwilson:
Sorry, I didn't see the response. Even your Powershell screenshot is not working correctly; it should be displaying the debug message of the port listening. It's possible the generator itself needs to support powershell first before adding that here to prevent confusion, as that powershell command is missing the output it should be printing.

Downes:
> it should be displaying the debug message of the port listening. 

No, the command fails, so the app never starts, so there is no port listening. The error displays in the only place it can, in the shell where the command was issued.

dougwilson:
Hi @Downes I'm referring to the second command in your screenshot: the command is supposed to produce the output "myapp:server Listening on port 3000 +0ms" but that is missing from your command output for the command you're saying is "working".

Downes:
Understood, looking at it now...

Downes:
OK, it was because I had clipped out the 'listening' message on the screen shot. Please see this one:

image

dougwilson:
Are you sure whatever that is really is in PowerShell? Can you try a different machine? Here is the command on my Windows 10:

image

Perhaps we need another user who has Windows 10 to try it as well to verify which one of our machines is what users will actually encounter.

dougwilson:
Actually, I just realized, that output you're showing is wrong, for some reason. It should not be "Example app listening on port 3000!" it is supposed to be "myapp:server Listening on port 3000 +0ms".

dougwilson:
Can you make a new, blank project and follow this page you're editing step by step to get your app to the exact same state someone would have it when following the instructions on this page?

Downes:
Yres, good point, it's pulling the error message from an app.js Hello World example from a higher level directory... let me delete that and try again...

Downes:
> Can you make a new, blank project
Will do

Downes:
OK, I've duplicated your results. Note that Express is in fact listening on 3000.

I tried running the two commands separately, but it still doesn't produce a 'listening' message. I'm not sure where in Express this message would be generated.

image

dougwilson:
Right, so that brings it back around to https://github.com/expressjs/expressjs.com/pull/1015#issuecomment-458372158 , in that it's possible that either (a) the command you have is not actually the correct command to use with PowerShell at all or (b) the generator itself needs to support powershell first before adding that here to prevent confusion, as that powershell command is missing the output it should be printing.

Downes:
OK, with reference to https://github.com/visionmedia/debug#windows-note

Here is what works: $env:DEBUG='myapp:*';npm start

image

PR #1016, opened 2019-01-29
Update generator.md

Downes:
Corrected to say PowerShell

### DIFF
--- a/en/starter/generator.md
+++ b/en/starter/generator.md
@@ -81,6 +81,13 @@ On Windows, use this command:
 > set DEBUG=myapp:* & npm start
 ```
 
+Or in Windows PowerShell, use this command:
+
+```sh
+> set DEBUG=myapp:* | npm start
+```
+
+
 Then load `http://localhost:3000/` in your browser to access the app.
 
 The generated app has the following directory structure:


en/starter/generator.md
```diff
@@ -81,6 +81,13 @@ On Windows, use this command:
 > set DEBUG=myapp:* & npm start
 ```
 
+Or in Windows PowerShell, use this command:
+
+```sh
+> set DEBUG=myapp:* | npm start
```
dougwilson, line 87: So this command doesn't produce the expected debug output when run. It should show the line of what port it's listening on. This makes me believe that, even though now it doesn't _error_, it's still not the correct command to be running in PowerShell.

dougwilson (changes requested):

en/starter/generator.md
dougwilson: The command is expected to produce the output "myapp:server Listening on port 3000 +0ms" on screen. When I run with Powershell, no output is produced, so there is probably a different PowerShell command that should be here instead.

Here is a side-by-side screenshot of the cmd.exe that produces the expected output, and the PowerShell command here which is missing the expected output.

image

PR #1020, opened 2019-02-15
Fix req.protocol hash link

KubaJastrz:
When using with `/4x/api.html` it skips language check and breaks the hash-link.

Check it out: http://expressjs.com/en/4x/api.html#req.protocol


dougwilson (approved):
Indeed, the current redirect strips the hash, but it looks like the other links are just the bare hash, and this is inline with those.

PR #1021, opened 2019-02-18
Remove --save option and comment

stevemoser:
As npm 5.0+ has been out for awhile.


dougwilson:
let's wait until at least it's an unsupported version of npm. For example, Node.js 6.x is still in Maintenance LTS status (https://github.com/nodejs/Release) and the latest version of 6.x bundles npm 3.10.10 (https://github.com/nodejs/node/blob/v6.16.0/deps/npm/package.json#L2).

PR #1022, opened 2019-02-20
Update express-generator docs to use npx

pspi:
The docs for express-generator could use `npx` these days.

The user would be left with an unpolluted global dependencies directory and if they choose to re-run it later, they'd always run the latest version.

Npx was introduced in Node.js 8.2.0 / npm 5.2.0 (July 2017).

### DIFF
--- a/en/starter/generator.md
+++ b/en/starter/generator.md
@@ -10,10 +10,17 @@ redirect_from: "/starter/generator.html"
 
 Use the application generator tool, `express-generator`, to quickly create an application skeleton.
 
-The `express-generator` package installs the `express` command-line tool. Use the following command to do so:
+You can run the application generator with the `npx` command (available in Node.js 8.2.0).
 
 ```sh
-$ npm install express-generator -g
+$ npx express-generator
+```
+
+For earlier Node versions, install the application generator as a global npm package and then launch it.
+
+```sh
+$ npm install -g express-generator
+$ express
 ```
 
 Display the command options with the `-h` option:


wesleytodd:
Could you add the `npx` example as an option but keep the old way?  Until express does not support older npm versions I would prefer we keep the docs with the existing ways.  I think it is fine to present the new way first but specify its only for `>8.2`.

pspi:
Sure. How about this?

crandmck:
LGTM. @wesleytodd any input?

wesleytodd:
Can you keep the old formatting with the backticks and the leading `$`.  Those are there because it allows styling code in the readme with syntax highlighting (although not used here, but best to keep consistency) and the `$` is a nice indicator that it is a terminal command.

Sorry for the delayed response. For some reason Github is delivering random old notifications to me now.  I wonder if my few month hiatus backed them all up or something?

crandmck:
@pspi I think we can land this if you make the changes suggested by @wesleytodd ...

pspi:
Consistency is key! Here's with backticks and sh specifier.

wesleytodd:
Hey @pspi, thanks for the updates, they look good!  One bit of advice for your future contributions  (which you hopefully make!).

Squash your changes down to one clean commit on top of the target branch.  It helps reduce the work the merging maintainer has to do to keep things clean. In case you are unsure how to get this done, you can do it a few ways:

- Amend the original commit when changes are made: `git commit --amend`
- Rebase squash once you are complete: `git rebase -i <original commit hash>`

Sometimes it is not necessary, but even a `git merge --squash` done by the maintainer often adds an unneeded merge commit.  Not all maintainers care, but saving time for maintainers of larger projects means they have more time for other more important things than cleaning up git history.

Thanks for listening to my Saturday thoughts 🍺 😁

pspi:
Squashed and rebased to top of gh-pages.

crandmck:
Thanks for the contribution @pspi !

PR #1024, opened 2019-03-03
Added links to next chapter at the end of the content in guides

a-deeb:


crandmck:
OK, I tested this and it works.  Thanks for the PR @a-deeb.  

At some point, it might be nice to give the next/previous links a little slicker appearance, but that can come later.  They are functional, and I do think it is a nice convenience.

PR #1027, opened 2019-03-24
Update req.query example for collection values of same key

jenilChristo:
- This PR will add a missing example to `req.query` documentation.
- To show case how express.js parses the query string when the url contains multiple values or a collection for the same key.
- This is internally done by the `parse` method in `qs` module used by express.js 

```
http://example.com/users?id[]=1&id[]=2&id[]=3

req.query.id => [1, 2, 3];
```


crandmck:
LGTM, thanks!

PR #1030, opened 2019-04-03
docs(frameworks.md) - Add NestJS (https://github.com/nestjs/nest)

johnbiundo:
Nest seems to fit in with this list.  The description is pulled from the header at https://github.com/nestjs/nest, but is a bit long.  Happy to edit it down if that is recommended.

As I understand the guidelines it's appropriate to submit just the en version and professional translators will update other language versions.   Please advise to the contrary.


dougwilson (approved):
LGTM

PR #1042, opened 2019-05-11
translation update.

silverskyvicto:
Hi.

I corrected the translation because there was a difference with the English version of the official document.

How is it?


crandmck:
Thanks!

Issue #1044, opened 2019-05-20
Search bar and menu overlapping


madflow:
* !!! DAMN YOU CSS !!!

![grafik](https://user-images.githubusercontent.com/183248/58049931-4f52c200-7b4e-11e9-81dc-9146b678be2c.png)

* Firefox 66.0.5 
* Arch Linux
* 1440 x 900

wesleytodd:
Hm, I tested in FF 66 on mac and am not able to reproduce.  If you have a fix PR that would be great!

reticulan:
![image](https://user-images.githubusercontent.com/7167113/58183775-16dfef00-7cda-11e9-807c-87c8673e972b.png)

same issue chrome 74 1366 x 768

expressjs.com/ru:
![image](https://user-images.githubusercontent.com/7167113/58183971-72aa7800-7cda-11e9-8edc-5698e4f393c1.png)

PR #1045, opened 2019-05-22
quick fix for #1044

madflow:
This is a quick fix for #1044 .

I am not quite sure what `min-width` and `max-width` is doing good here. By removing the `max-width` the navbar looks okayish in Firefox again.

hth

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -802,7 +802,6 @@ h2 a {
 
 .algolia-autocomplete {
   min-width: 12em;
-  max-width: 12em;
   top: -0.2em;
 }
 


wesleytodd:
If I remove that in my FF it has no effect.  This is the first time I have dug into the markup/css of the website, and weow it could use some love :)

This looks like a good fix to me if it fixes your issue!

madflow:
I actually tested with Browserstack and Firefox on Win10 and Mac - the issue was not reproducible. This seems to be a FF on Linux Problem. 

But yeah - it fixes the issue I am seeing.

dougwilson (approved):
Not sure what the purpose of the max width was for, but removing it doesn't seem to break anything, and if it fixes your display, all the better

bhumijgupta:
The changes does not appear on the site, as I am still seeing the overlapped search bar and navigation.
Here's a screenshot: 
Screenshot from 2019-05-27 17-16-55

* Chrome Version 74.0.3729.157
* Ubuntu 18.04

PR #1047, opened 2019-05-29
Fix for navbar and logo position

### LINKED ISSUE #1046: Oddly placed navigation bar and logo in header
The navigation bar for the website is oddly placed. According to the general rules of UX, the margin of navigation content should be same from both the sides of the page to make it look consistent and maintain flow of data.
Also the logo is not placed at centre of the navigation bar, which makes it look off centered.

Current view:
current
Notice the distance of Express logo from left margin and distance of 'Resources' from right margin.

Proposed view:
proposed

As viewed on:
* Google Chrome Version 74.0.3729.157
* Ubuntu 18.04

bhumijgupta:
Closes #1046 
* The navbar has been positioned to the right
* The position of logo has been centered in navbar

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -222,7 +222,6 @@ a {
 
 #logo {
   position: fixed;
-  top: 2px;
   margin: 0;
   padding: 10px 25px;
   width: auto;
@@ -671,8 +670,6 @@ footer {
 /* navigation */
 
 #navbar {
-  padding-left: 5%;
-  max-width: 1000px;
   height: 100%;
   position: relative;
   display: flex;
@@ -701,7 +698,7 @@ footer {
   margin: 0;
   right: 0px;
   z-index: 1000;
-  margin-right: 15px;
+  margin-right: 25px;
 }
 
 #navmenu > li {


wesleytodd:
Maybe could you provide a screenshot of the changes? Just so we are all clear on what it should look like.

bhumijgupta:
Sure @wesleytodd 
### Current appearance of site:
original
### Appearance of site after changes:
changes

Also, I have checked that the proposed changes does not break the responsiveness of the the website.

dougwilson:
It does feel a bit harder to use, at least for me:

Before:
image

After:
image

The navigation now ends up way on the far right, but the content of the site does not move to be in the middle or anything.

bhumijgupta:
@dougwilson This does seem to be an UX issue for screens with width higher than 2080px (2K and 4K displays). The website seems to have been made initially for screen width 1080px as after 1080px the white space on website increases drastically and the content does not fill it.
I can suggest two solutions:-
1. Let the changes be same. It might require extra effort for users with displays >= 2K, but the number of these users are still less and most of the users still use 1366p or 1080p displays ([source](http://gs.statcounter.com/screen-resolution-stats/desktop/worldwide)). But it will enable greater feasibility and UI advantage for users with less than 2K displays. The website behaviour will be similar to that of Gmail in 4K displays with logout button in far top right


2. We can center the whole content of the website when the screen dimension is greater than 1440px.
Example:
4k
This is the approach followed by most of the websites but would require major changes to the stylesheet.

dougwilson (changes requested):
Yea, looking into this more, the bar is right aligned to the edge of the content area, not the display window. This works with displays of all sizes.

This change breaks for displays over 1000px wide by right aligning past the edge of the content area.

I would say either reject this PR and keep the site as it is or overhaul the site such that the content area will correctly center or stretch itself so the bar being on the far right makes sense.

bhumijgupta:
@dougwilson I will work on overhauling the design to center the content after 1024px.

PR #1048, opened 2019-06-05
Document other shell variable for debuging express

zombieleet:
Specified `DEBUG_FD` and `DEBUG_COLORS` shell variables to handle redirection of logs to a file and enabling/disabling color output respectively

### DIFF
--- a/en/guide/debugging.md
+++ b/en/guide/debugging.md
@@ -113,4 +113,24 @@ You can specify more than one debug namespace by assigning a comma-separated lis
 $ DEBUG=http,mail,express:* node index.js
 ```
 
+## Advanced options
+
+When running through Node.js, you can set a few environment variables that will change the behavior of the debug logging:
+
+| Name      | Purpose                                         |
+|-----------|-------------------------------------------------|
+| `DEBUG`   | Enables/disables specific debugging namespaces. |
+| `DEBUG_COLORS`| Whether or not to use colors in the debug output. |
+| `DEBUG_DEPTH` | Object inspection depth. |
+| `DEBUG_FD`    | File descriptor to write debug output to. |
+| `DEBUG_SHOW_HIDDEN` | Shows hidden properties on inspected objects. |
+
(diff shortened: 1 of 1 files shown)


wesleytodd:
Hi @zombieleet, thanks for the addition!  Can you talk about why it requires re-asigning the FD to put debug out into a file?  This will redirect *only* debug output, but wont most people be alright with just normal piping to file `$ DEBUG=express:* node index.js > log.txt`?  Just thinking about keeping the docs simple :)

zombieleet:
@wesleytodd `$ DEBUG=express:* node index.js > log.txt` does not send the `DEBUG` output to log.txt

wesleytodd:
Oh, the debug module outputs on `stderr`?  Sorry I am just not a user of that module so am not familiar with the way it works.  Even so, you don't need to change the `fd` I think.  I tried doing:

```
$ cat index.js
'use strict'
const debug = require('debug')('foo')
debug('testing')

$ DEBUG=foo node index.js 2> log.txt

$ cat log.txt
2019-06-07T16:09:52.972Z foo testing
```

Seems to work with the normal `stderr` fd, right?

zombieleet:
Yes, that is right. But I see no reason why the default stream should be `stderr`. But the default stream can be changed to `stdout` [output streams](https://www.npmjs.com/package/debug#output-streams) . I think this should be in the docs, most express.js users might be interested in sending the logs to a file rather than the terminal

wesleytodd:
> I see no reason why the default stream should be stderr

I think this is a common best practice for *nix systems, a program's debug output being on `stderr`.  While I understand your viewpoint, I see no reason we should recommend something different from the norm.

> I think this should be in the docs, most express.js users might be interested in sending the logs to a file rather than the terminal

This I completely agree with.  And I would like to merge this, but I think adding the more complex example might make it harder for beginners to know what is going on.  That is why I was hoping to simplify the example to just `DEBUG=foo node index.js 2> log.txt`.

dougwilson (approved):

PR #1051, opened 2019-06-30
Update index.md

hyeonjae:


wesleytodd:
Hi @hyeonjae, I think I understand this change correctly, buy can you confirm that this makes the Korean docs link to the correct language docs on other pages?  Was this not working before?  Can you describe the behavior before and after?

crandmck:
OK since @hyeonjae hasn't responded (perhaps language issues?) ... this PR tries to fix the broken link on the Korean home page http://expressjs.com/ko/ to http://expressjs.com/ko/ko/changelog/4x.html#4.17.1 but we don't translate the changelog, so it should be to http://expressjs.com/en/changelog/4x.html#4.17.1.

Unfortunately, this PR doesn't fix it properly, as it changes the link to http://expressjs.com/changelog/4x.html#4.17.1 which doesn't currently work, either. 

However, this is easily fixed, which I will do, since (as most other pages on the site do), we should fallback from that URL to http://expressjs.com/en/4x/api.html.

It also fixes a link to the non-existent Korean "Frameworks" page (which was also a more recent edition) to be to the en page.

crandmck:
n.b. Only the Turkish translation has a similar (broken) link to the changelog.  None of the other translations links to the changelog (which as I recall we added some time after the original IBM translations).

crandmck:
I'm going to go ahead and land this since Doug +1 ed #1055 (thanks @dougwilson!)

PR #1052, opened 2019-06-30
Replace jade with pug

profiluefter:


crandmck:
Except the Express generator still actually puts Jade in `package.json`:
```
...
"jade": "~1.11.0",
...
```
And when you run `npm install` you get this warning:
```
npm WARN deprecated jade@1.11.0: Jade has been renamed to pug, please install the latest version of pug instead of jade
```

So not sure it's entirely accurate to say it uses Pug...

wesleytodd:
@profiluefter, maybe you could submit a pr to update the [generator](https://github.com/expressjs/generator)?  If we have that we could for sure update the docs.

profiluefter:
@wesleytodd I'm not very experienced with node or express so I think it would be better if someone else would do this. In fact I was just getting started with node in general and was confused with the deprecated package on the page.

wesleytodd:
It is never to soon to start @profiluefter :)

In this case the change would be almost just a search and replace [in this file](https://github.com/expressjs/generator/blob/master/bin/express-cli.js).  Specifically it would be these lines:

https://github.com/expressjs/generator/blob/master/bin/express-cli.js#L463-L467
https://github.com/expressjs/generator/blob/master/bin/express-cli.js#L229-L230
https://github.com/expressjs/generator/blob/master/bin/express-cli.js#L301-L303
https://github.com/expressjs/generator/blob/master/bin/express-cli.js#L463-L467

As far as simple changes go, this is a great first contribution!

PR #1056, opened 2019-07-16
Fix unnecessary ',' in docs.

sanair96:


wesleytodd:
Hi @sanair96, thanks for the PR.  I am sure there could be a debate about the oxford comma, but personally I think it belongs.  I am not sure if we have a docs style guide, or if there would be value in creating one, but I would vote in favor of the oxford comma if I could.

sanair96:
Yeah didn't cross my mind it could be an Oxford comma.

PR #1057, opened 2019-07-19
Update code samples in 5.x docs to use const

RamirezAlex:
This update the code samples in 5.x alpha documentation to use the modern keyword `const` instead of `var`

Please review @crandmck @dougwilson :)

### DIFF
--- a/_includes/api/en/5x/app-listen.md
+++ b/_includes/api/en/5x/app-listen.md
@@ -4,8 +4,8 @@ Starts a UNIX socket and listens for connections on the given path.
 This method is identical to Node's [http.Server.listen()](https://nodejs.org/api/http.html#http_server_listen).
 
 ```js
-var express = require('express')
-var app = express()
+const express = require('express')
+const app = express()
 app.listen('/tmp/sock')
 ```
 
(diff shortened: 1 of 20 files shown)


crandmck:
Seems reasonable to me.  
@dougwilson @wesleytodd Do you have an opinion?

_includes/api/en/5x/app-listen.md
```diff
@@ -43,7 +43,7 @@ The `app.listen()` method returns an [http.Server](https://nodejs.org/api/http.h
 
 ```js
 app.listen = function () {
-  var server = http.createServer(this)
+  const server = http.createServer(this)
```
dougwilson: This is meant to be the exact code that is in the Express source tree: https://github.com/expressjs/express/blob/5f0c829d7ca7da746ee859f13a54631000f8a9b5/lib/application.js#L602-L605

dougwilson (changes requested):
I'm fine with it, though found at least one issue. There is a lot to go through, so I may have missed something. I'm not sure if this was just done with only as find and replace, or the examples were all tested to validate they are working or not still after the changes.

wesleytodd (approved):
Other than @dougwilson's comment, LGTM!

crandmck:
Since he made Doug's suggested change, I'm going to merge.
@RamirezAlex sorry for the delay.

PR #1059, opened 2019-07-28
Fixes ISRG page link

bomoko:
Current link gives a 404. Should point to https://www.abetterinternet.org/about/


wesleytodd:
Is there a reason to link to the about page?  I think we should just link to their homepage, since this link is more of a general info link, not to some specific topic.

bomoko:
Fair enough! More than happy to change.

PR #1063, opened 2019-08-17
Update res.cookie for setting multiple cookies in a single response

hematy61:
I have updated the doc to show how multiple key values for a cookie can be set.

### DIFF
--- a/_includes/api/en/4x/res-cookie.md
+++ b/_includes/api/en/4x/res-cookie.md
@@ -28,6 +28,20 @@ res.cookie('name', 'tobi', { domain: '.example.com', path: '/admin', secure: tru
 res.cookie('rememberme', '1', { expires: new Date(Date.now() + 900000), httpOnly: true })
 ```
 
+And if you want to set multiple key values, simply call `cookie` more than once.
+
+For example:
+
+```js
+res
+.status(201)
+.cookie("access_token", "Bearer " + token, {
+  expires: new Date(Date.now() + 8 * 3600000) // cookie will be removed after 8 hours
+})
+.cookie("test", "test")
+.redirect(301, "/admin");
+```
+
 The `encode` option allows you to choose the function used for cookie value encoding.
 Does not support asynchronous functions.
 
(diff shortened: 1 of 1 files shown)


wesleytodd:
Hey @hematy61, I am interested in what this adds which is not covered or clear with the existing docs?  Was it unclear from the existing docs how to do this?

hematy61:
Hey @wesleytodd. thanks for asking. I didn't find anything mentioning how to add more than one key value to a cookie. is there any? if there is, shouldn't it be mentioned here as well?

wesleytodd:
> add more than one key value to a cookie

Not sure what you mean by this.  A cookie can have only one value.  In your example you are just adding two separate cookies, which is basically the same code as the example above:

```
res.cookie('name', 'tobi', { domain: '.example.com', path: '/admin', secure: true })
res.cookie('rememberme', '1', { expires: new Date(Date.now() + 900000), httpOnly: true })
```

I was just wondering if that code snippet is not enough to illustrate the point.  If not, I could see maybe using the same code block and adding to the paragraph above it a point about "this example sets two cookies in one response" or something to that effect.  Make sense?

hematy61:
I see what you mean. In the beginning, I was confused about how I can set 2 or more key values in one request. technically, 
```js
res.cookie( 'key1', 'value1')
res.cookie( 'key2', 'value2')
res.send()
```

is the same as chaining them together like:

```js
res.cookie( 'key1', 'value1')
      .cookie( 'key2', 'value2')
      .send()
```

but for someone who doesn't know chaining is working here, it's gonna be confusing (as it was for me). I had to search to see how I can add multiple key values in one response. I hope my point makes sense now.

wesleytodd:
That makes way more sense to me now.  I think a great addition would be to show that you can do either of those with a single line of "You can set multiple cookies in a single response by calling `res.cookie` multiple times, for example:"...

hematy61:
Right on. that's it. And I have modified it with what you have suggested.

wesleytodd:
Looks fine to me now.  Just needs formatting cleanup, use two space indentation, single quotes, and no semicolons (sorry for the nitpicks, but gotta keep things consistent with the formatting guidelines moving forward!).

@crandmck do you have opinions on this addition?

hematy61:
No worries. it should be good now.

hematy61:
shouldn't we close it?

crandmck:
LGTM... Thanks for the contribution!

PR #1065, opened 2019-08-26
Update healthcheck-graceful-shutdown.md

manjufy:
Based on the current documentation, for the following 
 
```
terminus(server, {
	signal: 'SIGINT',
	healthChecks: { '/healthcheck': onHealthCheck },
	onSignal
});
```

Throws an error `terminus is not a function`.

Updating the documentation to the latest implementation of terminus.


crandmck:
Based on https://github.com/godaddy/terminus this change looks correct.

Issue #1066, opened 2019-09-01
static-files.md (Case sensitivity of path)


bochap:
Hi,

I encountered a case that when serving files with MacOS as host for express.static, it will fail to find the file if the path requested is of a different case as the file. I look debugged into the code and discovered that `express.static` is actually using the npm package `serve-static` and `serve-static` uses the npm package send to return the file.

And send actually uses Node.js's `fs.stat` to check that the file exists before returning the file. And the case sensitivity of that call depends on the host OS of Express.js. For host OSes that are case sensitive the call will return an error of the file not being found.

```
  fs.stat(path, function onstat (err, stat) {
    if (err && err.code === 'ENOENT' && !extname(path) && path[path.length - 1] !== sep) {
      // not found, check extensions
      return next(err)
    }
    if (err) return self.onStatError(err)
    if (stat.isDirectory()) return self.redirect(path)
    self.emit('file', path, stat)
    self.send(path, stat)
  })
```

Would it be a good idea to add a note that this middleware could be case-sensitive depending on the host of Express? Since it could cause behaviors that are not easily traceable if the developer most an application built on a case insensitive OS and deployed it on a case sensitive OS

wesleytodd:
Hey @seetd, I am fine adding a comment to this effect.  PR's welcome!

bochap:
Hi @wesleytodd ,

Thanks I will look into adding in the comment.

bochap:
Hi @wesleytodd,
 
As an update I am still working on this. I had to switch to a new Mac so it took sometime to get everything up again. PR should be coming soon

bochap:
@dougwilson Sorry for not understanding the structure of the project. Should I add the note to the Readme in https://github.com/expressjs/serve-static for this?

dougwilson:
It depends on where exactly you're looking to put it on the website. If you want to put it in that exact location, then it would need to be incorporated into that module's readme. If you wanted to put it in express.static docs, then no.

But that brings up a good point: perhaps we should come up with some wording we want to use here and where it should go in the docs? For example I assume that would also apply to res.sendFile API as well.

dougwilson:
[...]

PR #1067, opened 2019-09-15
Add note for case sensitivity of path for serve-static #1066

bochap:
This adds a note that explains that the case sensitivity for serve-static depends on the underlying filesystem. It uses the format similar to body-parser.md that provides a short description of the module followed by the `note`. I will be happy to modify it if that is not the correct way to add the note. Thanks for your review time


dougwilson:
All the files in `_includes/readmes` come directly from the relevant module's readme. If we accepted this change, it would just get overwritten back to the readme that is in `serve-static` module.

PR #1071, opened 2019-09-25
Update hello-world.md

hyeonjae:
update old sample code

### DIFF
--- a/ko/starter/hello-world.md
+++ b/ko/starter/hello-world.md
@@ -15,8 +15,8 @@ lang: ko
 
 `myapp` 디렉토리에 `app.js`라는 이름의 파일을 작성한 후 다음과 같은 코드를 추가하십시오.
 
-<script src="https://embed.runkit.com" data-element-id="hello-example" data-mode="endpoint" async defer></script>
-<div id="hello-example"><pre><code class="language-js">
+<pre>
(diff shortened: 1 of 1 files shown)


wesleytodd:
Hey @hyeonjae, it looks like there are two things here:

1. updates to the example to make it use arrow functions
2. adding a script from runkit

So I am not opposed to number 1, even though I think it is unnecessary.  But Number 2 is something I think we should avoid.  If we are going to do code highlighting (which is my assumption about what this does) I really dont want to do it with a third party script.

crandmck:
@hyeonjae removed the Runkit stuff, so what's left is just changing `var` to `const` and using arrow functions.  In some ways, I think using arrow functions may obfuscate the sample code, even though it's more "modern" JS.  Can we safely assume that all readers will know arrow functions?
@wesleytodd @dougwilson Do you want to weigh in?

wesleytodd:
>  Can we safely assume that all readers will know arrow functions?

I think we can.  If nothing else it will not be a put off to the more "modern" folks out there.  My concern is that with our delay in 5.0 it might increase the concern about the project momentum being low if we also are not using modern and common syntax.  So in this case I think it is good to update the docs.

dougwilson:
I'm not opposed to it. This section is not even express version-specific.

PR #1073, opened 2019-10-07
Removed Broken Link

realabbas:
# Removed Broken Link

Screenshot of the broken link

image

**Removed it from all the language-specific files**

**Full Report** : https://gtmetrix.com/reports/expressjs.com/pXc1bK8U

### DIFF
--- a/_includes/footer/footer-de.html
+++ b/_includes/footer/footer-de.html
@@ -1,23 +1,3 @@
-<!-- NLS_CHARSET=UTF-8 -->
-<noscript>
-    <iframe src="//www.googletagmanager.com/ns.html?id=GTM-5XL76H" height="0" width="0" style="display: none; visibility: hidden;"></iframe>
-</noscript>
-<script>
-(function(w, d, s, l, i) {
-    w[l] = w[l] || [];
-    w[l].push({
-        'gtm.start': new Date().getTime(),
-        event: 'gtm.js'
-    });
-    var f = d.getElementsByTagName(s)[0],
-        j = d.createElement(s),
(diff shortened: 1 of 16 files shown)


crandmck:
I think this is a leftover from StrongLoop/IBM days when they were using Google Analytics to track visits to the site.  AFAIK, we are no longer doing this, so removing this stuff is fine.  This just removes the "noscript" version (ie. if browser JS is not enabled). 

My only question is whether we should remove the "normal" version (if client JS is enabled, which I assume is the norm), e.g. https://github.com/expressjs/expressjs.com/blob/gh-pages/_includes/footer/footer-en.html#L4-19

I can't see any reason to keep it, unless there's a GA account that someone is using and I don't know about it.

Thoughts? @dougwilson @wesleytodd

wesleytodd:
>unless there's a GA account that someone is using and I don't know about it.

Yeah, this would be good to know.  Maybe @dougwilson has it?  If so I would love to see it.  If not I agree we should remove it all.

dougwilson:
I don't have a GA account; my understanding is that it was from StrongLoop/IBM. We can probably just get rid of it unless we want to actually set up an account and point to that.

wesleytodd:
I think it would be good to have analytics.  I am not a fan of GA, but that is really the most common and powerful free solution.  I wonder if there is something the OpenJS Foundation would think of doing for member projects, of if one of the paid providers  with better data privacy policy would be willing to offer it for free in exchange for logo placement or something?

dougwilson (approved):
This is good regardless of having a replacement; if we actually add a replacement, we can add in whatever that is at that time 👍

PR #1074, opened 2019-10-11
Add protocol and host to the console log

sparksis:
Adding the protocol and host the console log message so the terminal can generate clickable links.


sparksis:
I've tested this with a few terminals.  It still doesn't fully work as some terminals will consume the `!` token when generating the links and will attempt to generate a link for `http://localhost:3000!` rather than `http://localhost:3000`.

Suggestions welcome.

dougwilson (approved):
problem solved by removing the ! (and updated the translations)

PR #1077, opened 2019-10-14
Fix a typo in serve-static

stsourlidakis:


dougwilson:
These readmes are all just pulled via a script from their source repos. If you make this PR over at https://github.com/expressjs/serve-static the source will be fixed, which will update this copy. If we merge the fix here, it will end up just reverting back without the source being fixed anyway :)

stsourlidakis:
Thanks, I'll open the PR on the correct repo.

PR #1078, opened 2019-10-20
Update writing-middleware.md

saurabhbazzad:
Added the example of error handling middleware as specified in Issue#1069


dougwilson:
We already have a full page on error handling middleware. Can we just add a link to it instead of duplicating the information? If not, why not?

crandmck:
Agree... this duplicates http://expressjs.com/en/guide/using-middleware.html#middleware.error-handling.

PR #1079, opened 2019-10-22
Change Exove logo

Hanzig:
Replace old Exove logo with new one.


wesleytodd:
Hey @Hanzig, this is one of the issues I have with showing company logos in the website, so do not take this as personal feedback.  BUT, how are we supposed to know you represent this company?

Also, why do all these random companies have free advertising on the express website without giving anything back to the project?  We (the maintainers) are going to have to spend time vetting this request and merging this pr?  Seems like we should just delete that page?

crandmck:
Excellent points @wesleytodd.

I think this is another case of "stuff left over from the StrongLoop/IBM era".  I think it was originally meant to show the wide adoption of Express, and of course, provide visibility for IBM and other companies that were active in some aspect with Express/Node then.  It's certainly not a complete list of ppl using Express (which I think would be virtually impossible to create).

At this point, I agree we should remove this page, for the reasons you cite (maintenance, inability to ascertain authenticity, etc).  Are there any arguments for keeping it?

wesleytodd:
I think the reason to keep it would be to charge a fee for companies to be listed here.  Fund raising for OSS is hard, and this seems like a very easy way for the project to have a bit of a budget for things (like swag or putting bounties on important work).  In a perfect world we would turn this into a way for companies to promote themselves and we take the $$$ and use it for improving the project.

dougwilson:
I think this is a good discussion; please open it in a separate issue to not get lost, as it seems off-topic from the direct updating of the logo at hand 👍

dougwilson (approved):
added transparency to the logo

PR #1080, opened 2019-10-25
Fixed a typo in router.md

RMuskovets:
There's a small change proposed by me: the comma after "mini-application" was in the quotes.
I've fixed that.


dougwilson:
This is correct in, at least, American English. The comma always goes inside the quotes. Here is a good reference: https://www.grammarly.com/blog/quotation-marks/

RMuskovets:
Oh, I didn't know that. Thanks!

Issue #1076, opened 2019-10-14
Fix body-parser link in https://expressjs.com/ko/api.html


2kindsofcs:
hyperlink to body-parser page in [this page](https://expressjs.com/ko/api.html) is broken. (express.com in Korean)
 it returns 404 error page. it should return [the right one](https://expressjs.com/en/resources/middleware/body-parser.html).

crandmck:
@2kindsofcs Please open a PR!  We rely entirely on native speakers in the community to maintain the translated pages.... Thanks!!

PR #1082, opened 2019-10-29
Fix broken links in express.com in Ko

2kindsofcs:
#1076 

I found that most of the links in this [page](https://expressjs.com/ko/api.html)(express.com in Korea) are broken. they are in form of something like this: ` /{{page.lang}}/resources/middleware/a-module.html`

As there is no `middleware` folder under the `expressjs.com/ko/resources` , the websites throws 404 page when a link is requested.  So I made `middleware` folder and it would fix the errors.

### DIFF
--- a/ko/resources/middleware/body-parser.md
+++ b/ko/resources/middleware/body-parser.md
@@ -1,0 +1,8 @@
+---
+layout: middleware
+title: Express body-parser middleware
+menu: resources
+lang: en
+redirect_from: '/resources/middleware/body-parser.html'
+module: body-parser
+---
--- a/ko/resources/middleware/compression.md
+++ b/ko/resources/middleware/compression.md
@@ -1,0 +1,8 @@
+---
+layout: middleware
+title: Express compression middleware
+menu: resources
+lang: en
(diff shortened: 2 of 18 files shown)


crandmck:
Hi @2kindsofcs thanks for the PR.  I'm sorry I didn't notice this before when I commented on #1076 but the proper way to fix this is to just remove the lang code from the link
```
[body-parser](/{{ page.lang }}/resources/middleware/body-parser.html)
```
so it becomes
```
[body-parser](/resources/middleware/body-parser.html)
```
There are a small number of occurrences of this link in the `/api/en/4x` files.

Because we set up English as the "fallback" language, this redirects to http://expressjs.com/en/resources/middleware/body-parser.html.

We haven't translated all the middleware docs, and none of the other languages have middleware stubs as you have in this PR.  Also, having files in the `/ko` directory with `lang: en` in the front-matter is not right.  The lang attribute should match the language.

BTW, I discovered that all the other languages have the same problem (unfortunately).

I'm sorry it's taken me so long to review your PR, but if you could do the above, it would be much appreciated.

PR #1088, opened 2019-12-06
Add SwissDevJobs Devops as a company using Express

Vrq:


Vrq:
@crandmck would you mind having a look?

Vrq:
@crandmck would you mind having a look and merge it?

crandmck:
The jobs listed there are not Express-specific, so not really appropriate. Plus, we don't generally provide job listings.

PR #1099, opened 2020-01-22
Added the link to blog of express

pankajkrr:


jonchurch:
Hello there! Thank you for contributing 🎉 

I'm not personally sure what the point of this section of the website is, I can infer that it's to list blogs that publish content about Express, based on what is already listed here.

The link you've added goes to a single blog post, not a collection of articles. There is also at least one other link already posted which goes to a collection with just a single blog post, however. 

Like I said, I'm not exactly sure what, if any, criteria there is for submitting resources here, but I wanted to make a note of this for anyone else who reviews this PR.

wesleytodd:
I feel like we discussed removing some of this.  I cannot remember where but I feel like we need to go ahead with that.  Keeping this kind of a list is unmaintainable.  Even if we don't do that quickly I think we want to close this due to @jonchurch's concerns.

PR #1102, opened 2020-01-30
Adding CBRE logo

chemecse:
Adds the CBRE logo to the "Companies using Express in production" page.

Screen Shot 2020-01-30 at 1 44 47 PM


wesleytodd:
Hi @chemecse, we have discussed a few times removing this page entirely.  So many companies are using express that the free advertising is really not something we are interested in maintaining anymore.  Sorry about this, but we will not be moving forward with more additions here.

chemecse:
@wesleytodd Makes sense, I appreciate the explanation.

Issue #1095, opened 2020-01-15
Translation error - German


moopat:
Across the website `Fork the website on GitHub` is wrongly translated to German:

`Verzweigung der Website auf GitHub`

It sounds like someone tried to translate it literally, but `fork` cannot be translated (or at least Gitlab doesn't do it; GitHub is not available in Geman).

A better translation would be `Fork der Webseite auf GitHub erstellen` or `Webseite auf GitHub forken`.

Let me know if I should make the changes myself or if a translator will take care of it, since according to [the Contribution guidelines](https://github.com/expressjs/expressjs.com/blob/gh-pages/CONTRIBUTING.md) German is professionally translated.

wesleytodd:
A PR for this change is welcome!  In the history there was some support providing translations, but not anymore.  Also PR's for improvements are **always** welcome!

moopat:
Okay, I submitted a PR: #1105

PR #1105, opened 2020-02-06
Fix a German translation issue, resolving #1095

moopat:


wesleytodd:
Any German speakers want to approve this?  Doesn't seem wrong to me, but I don't read German 😆

wesleytodd:
Oops, I see you had an explanation in that other issue, which I forgot about.  I will just merge this since it seems reasonable.

PR #1106, opened 2020-02-07
Fix typo in Ko

No-YE:
I found a typo in Ko, and fix it.
In English, it's like changing 'pplication' to 'application'.


wesleytodd:
Seems reasonable to me, thanks for the fix!

PR #1094, opened 2020-01-12
Update req-secure header and excerpt

markjszy:
Based on behavior of current site, it looks like req-secure ought to have the header reinstated. Also, es-lint fails with the JS block-style example for a few reasons, most importantly the "unused expression". For this example, I opted to just have a single line instead of the block-style code, which I am not sure could be made acceptable to es-lint without a less attractive option like introducing a variable (e.g. assign the result of the expression to isSecure...)

PR #1107, opened 2020-02-09
Update req-secure header and excerpt (quick workaround)

anton-marchenko:
There is a quick workaround for squashing for https://github.com/expressjs/expressjs.com/pull/1094
I just help for Mark to make PR faster: https://github.com/expressjs/expressjs.com/pull/1094#issuecomment-582473735

Based on behavior of current site, it looks like req-secure ought to have the header reinstated. Also, es-lint fails with the JS block-style example for a few reasons, most importantly the "unused expression". For this example, I opted to just have a single line instead of the block-style code, which I am not sure could be made acceptable to es-lint without a less attractive option like introducing a variable (e.g. assign the result of the expression to isSecure...)

revert to block code excerpt with eslint exception

move eslint exception comment

### DIFF
--- a/_includes/api/en/5x/req-secure.md
+++ b/_includes/api/en/5x/req-secure.md
@@ -1,5 +1,8 @@
- A Boolean property that is true if a TLS connection is established. Equivalent to: 
+<h3 id='req.secure'>req.secure</h3>
  
+ A Boolean property that is true if a TLS connection is established. Equivalent to the following:
+
+<!-- eslint-disable no-unused-expressions -->
 ```js
-req.protocol == 'https';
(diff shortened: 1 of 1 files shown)


dougwilson:
Maybe we should just match the 4.x docs, which passes the linting? The docs are the same for this between 4.x and 5.x, so not sure why they should be different anyhow. 4.x for reference: https://raw.githubusercontent.com/expressjs/expressjs.com/gh-pages/_includes/api/en/4x/req-secure.md

dougwilson:
https://github.com/expressjs/expressjs.com/pull/1094 has been squashed and merged. Thanks for your contribution!

PR #1103, opened 2020-02-04
Add Visitors Book project

anton-marchenko:
Add information about Visitors Book (open-source project that uses Express).

[Proof link](https://github.com/anton-marchenko/am.visitors-book/blob/02aacc1cfb58e87cb82af9ee4a4ffe80b0b0c4f8/package.json#L18) about Express usage.

This project was written by TDD methodology (and has 100% test coverage). Because of this, the project is important for community as study material.

PR #1108, opened 2020-02-15
Adding new projects has not supported temporarily

anton-marchenko:
@wesleytodd 

According to https://github.com/expressjs/expressjs.com/pull/1103#issuecomment-586562680 we should delete this description that does not waste the time developers who want to add new project here.


wesleytodd:
Awesome!! Thanks for doing this!  I will let this sit for a few days in case one of the other maintainers raises an objection.  But otherwise, LGTM.

jonchurch (approved):

crandmck:
Thanks for the PR @anton-marchenko ... I agree it makes sense remove the information that's not specifically about Express, since it just increases the doc. maintenance burden.  
Note that  several other pages have similar text:
- http://expressjs.com/en/resources/companies-using-express.html
- http://expressjs.com/en/resources/template-engines.html

We should probably make similar changes on those pages.

Issue #1126, opened 2020-04-01
Outstanding issues with the 5.x api docs


dougwilson:
5.x docs investigation notes:

- [x] missing docs for `req.host` (#1130)
- [x] `req.query` should say that it is `{}` when disabled, otherwise result of configured "query parser"
- [x] missing does for `app.router` (#1129)
- [x] `app.param` documents name as optional, but that whole system was removed.
- [x] `req.param` does not exist any longer (#1128)
- [x] missing docs around `next('router')`
- [x] missing docs on how to use promises in handlers/middlewares (#1109)

gireeshpunathil:
I can help with few of these; the key question is: how critical those are for express 5.0?

gireeshpunathil:
> req.param does not exist any longer

@dougwilson - I may be missing something; but https://github.com/expressjs/express/blob/2d519077ea6c7c52a16098c7cf6d8d667a863c7b/lib/request.js#L235 still shows the function definition; Is there a pending PR that removes this?

dougwilson:
This list is regarding the express 5.x code; linking to 4.x code would indeed showing it as existing, as it did exist in 4.x series :)

dougwilson:
Github provides a neat feature where you can browse the code for a particular version, but using the branch drop down and changing it to tag and picking the version tag of code to view. For example, the latest 5 request helpers can be viewed at the url https://github.com/expressjs/express/blob/5.0.0-alpha.8/lib/request.js . Another method is to git clone the repository and you can git checkout a specific version tag to explore the code of that version. I hope this is helpful :+1:

HarshithaKP:
>app.param documents name as optional, but that whole system was removed.

@dougwilson, can you pls explain what this is? I looked at the doc of app.param(), and found that it is complete. what is missing here?

HarshithaKP:
@dougwilson,  http://expressjs.com/en/guide/using-middleware.html#middleware.router has explanation for next('router'). Is this sufficient or more detail is required ?. If yes , where should I add it ?

ryhinchey:
I'm starting work on the 5.x api reference for how to use promises in handlers/middleware.  I see there are other places on the site that should reference promise based middleware and the express error handler (like the getting started docs).  I'll take a stab at those and create separate PRs for each

dougwilson:
> @dougwilson, can you pls explain what this is? I looked at the doc of app.param(), and found that it is complete. what is missing here?

[...]

PR #1127, opened 2020-04-02
doc: expand `req.query` change documentation

gireeshpunathil:
Refs: https://github.com/expressjs/expressjs.com/issues/1126

### DIFF
--- a/_includes/api/en/5x/req-query.md
+++ b/_includes/api/en/5x/req-query.md
@@ -1,7 +1,7 @@
 <h3 id='req.query'>req.query</h3>
 
 This property is an object containing a property for each query string parameter in the route.
-If there is no query string, it is the empty object, `{}`.
+When [query parser](#app.settings.table) is set to disabled, it is an empty object `{}`, otherwise it is the result of the configured query parser.
 
 <div class="doc-box doc-warn" markdown="1">
 As `req.query`'s shape is based on user-controlled input, all properties and values in this object are untrusted and should be validated before trusting. For example, `req.query.foo.toString()` may fail in multiple ways, for example `foo` may not be there or may not be a string, and `toString` may not be a function and instead a string or other user-input.
--- a/_includes/api/en/4x/req-query.md
+++ b/_includes/api/en/4x/req-query.md
@@ -1,7 +1,7 @@
 <h3 id='req.query'>req.query</h3>
 
 This property is an object containing a property for each query string parameter in the route.
-If there is no query string, it is the empty object, `{}`.
+When [query parser](#app.settings.table) is set to disabled, it is an empty object `{}`, otherwise it is the result of the configured query parser.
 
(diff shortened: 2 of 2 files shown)


dougwilson (changes requested):
The notes was to expand the req.query docs, not the migration guide.

You can keep it also in the guide if you are saying that this is a change in behavior between 4 and 5.

gireeshpunathil:
@dougwilson - got it, and added to the API doc too!

dougwilson:
Ok, but I don't think this behavior changed between 4 and 5. Can you elaborate on why you think it should be in the migration guide? Did the behavior actually change and I am just mistaken on it not?

_includes/api/en/5x/req-query.md
```diff
@@ -1,7 +1,7 @@
 <h3 id='req.query'>req.query</h3>
 
 This property is an object containing a property for each query string parameter in the route.
-If there is no query string, it is the empty object, `{}`.
+When disabled, the `req.query` will be an empty object (`{}`). Else, the result of configured query parser.
```
dougwilson: (Sorry that the mobile view doesn't allow me to make suggested changes).

Would the wording sound better as

> When `"query parser"` (ed: make a link) is set to disabled, it is the empty object `{}`, otherwise it is the result of the configured query parser.

My thought is beginning a sentence with "Else" seems weird (not sure if there is an English rule around that though), but also that "When disabled" is not clear it refers to the query parser setting when a user just jumps to req.query docs. Same with what "configured query parser" refers to, so having that called out and a back link to it seems useful.

gireeshpunathil: @dougwilson - made those changes, mostly as is. Please have a look at the link - as I am not sure about the link target as well as the link semantics.

gireeshpunathil:
> Can you elaborate on why you think it should be in the migration guide? Did the behavior actually change and I am just mistaken on it not?

My rationale is that because we are expanding on the API doc for better clarity, it also makes sense to add wherever it is mentioned. However, removed it from the migration guide.

_includes/api/en/5x/req-query.md
dougwilson: >  as I am not sure about the link target as well as the link semantics.

The best method to know is to test locally: https://github.com/expressjs/expressjs.com#local-setup

gireeshpunathil: unfortunately it failed (consistently) for me at the setup stage (missing ruby headers while installing json module). I will check with @HarshithaKP when she comes online and try to verify the link semantics.

meanwhile let me know if you are fine with the link target

gireeshpunathil: ok, link semantics is now verified, and fixed:
 - removed back-quote of href, as it was appearing in grey
 - removed the import of the link target document, as it wasn't necessary

dougwilson (approved):
LGTM

PR #1130, opened 2020-04-02
doc: add `req.host` property documentation

gireeshpunathil:
Refs: #1126

### DIFF
--- a/_includes/api/en/5x/req-host.md
+++ b/_includes/api/en/5x/req-host.md
@@ -1,0 +1,22 @@
+<h3 id='req.host'>req.host</h3>
+
+Contains the host derived from the `Host` HTTP header.
+
+When the [`trust proxy` setting](api.html#app.settings.table)
+does not evaluate to `false`, this property will instead get the value
+from the `X-Forwarded-Host` header field. This header can be set by
+the client or by the proxy.
+
+If there is more than one `X-Forwarded-Host` header in the request, the
+value of the first header is used. This includes a single header with
+comma-separated values, in which the first value is used.
+
+```js
+// Host: "example.com:3000"
+console.dir(req.host)
+// => 'example.com:3000'
+
+// Host: "[::1]:3000"
+console.dir(req.host)
+// => '[::1]:3000'
+```
--- a/_includes/api/en/5x/menu.md
+++ b/_includes/api/en/5x/menu.md
@@ -80,6 +80,8 @@
             </li>
             <li><a href="#req.fresh">req.fresh</a>
             </li>
+            <li><a href="#req.host">req.host</a>
+            </li>
             <li><a href="#req.hostname">req.hostname</a>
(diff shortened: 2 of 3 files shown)


_includes/api/en/5x/req-host.md
```diff
@@ -0,0 +1,18 @@
+<h3 id='req.host'>req.host</h3>
+
+Contains the hostname and port number derived from the `Host` HTTP header.
```
dougwilson: I would suggest this should say _either_ "Contains the host derived [..]" or "Contains the hostname and port number, concatenated with a colon, derived from [...]"

The only downside to the latter, is that it would also need to specify that IPv6 addresses will be wrapped in square brackets as well, since hostname for IPv6 removes the square brackets.

gireeshpunathil: @dougwilson  - pushed a change to that effect.
 - rephrased based on your second preference
 - added an IPv6 example for more clarity

jonchurch (approved):

_includes/api/en/5x/req-host.md
```diff
@@ -0,0 +1,24 @@
+<h3 id='req.host'>req.host</h3>
+
+Contains the hostname and port number, concatenated
+with a colon, derived from the `Host` HTTP header.
+IPv6 addresses will be wrapped in square brackets as well.
```
dougwilson: I realized that I gave two options before, and after looking at which one you picked, it turns out that the wording I gave was wrong 😢 The issue here is that most of the time, `req.host` is never going to actually contain the port number, and this will now be misleading as it seems to say it would indeed contain the port number. Maybe just going with the first one will work?

```suggestion
Contains the host derived from the `Host` HTTP header.
```

_includes/api/en/5x/req-host.md
```diff
@@ -0,0 +1,24 @@
...
+If there is more than one `X-Forwarded-Host` header in the request, the
+value of the first header is used. This includes a single header with
+comma-separated values, in which the first value is used.
+
+```js
+// Host: "example.com:3000"
+console.dir(req.host)
+// => 'example.com:3000'
+
+// Host: "::1:3000"
```
dougwilson: I'm not sure if this is supposed to be a HTTP Host header, if so:
```suggestion
// Host: "[::1]:3000"
```

dougwilson (changes requested):

_includes/api/en/5x/req-host.md
gireeshpunathil: @dougwilson - done, compressed as per your suggestion.

_includes/api/en/5x/req-host.md
gireeshpunathil: fixed.

dougwilson (approved):
thanks my friend!

PR #1131, opened 2020-04-03
doc: document `next('router')` behaviour

HarshithaKP:
Refs: https://github.com/expressjs/expressjs.com/issues/1126

### DIFF
--- a/_includes/api/en/5x/router-METHOD.md
+++ b/_includes/api/en/5x/router-METHOD.md
@@ -40,3 +40,27 @@ router.get(/^\/commits\/(\w+)(?:\.\.(\w+))?$/, function (req, res) {
   res.send('commit range ' + from + '..' + to)
 })
 ```
+
+You can use `next` primitive to implement a flow control between different
+middleware functions, based on a specific program state. Invoking `next` with
+the string `'router'` will cause all the remaining route callbacks on that router
+to be bypassed.
+
+The following example illustrates `next('router')` usage.
+
+```js
+function fn (req, res, next) {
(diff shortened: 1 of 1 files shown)


dougwilson (changes requested):
Thanks! I think we need to get some actual documentation around this feature, though. An example is always good for visual learners, but we should be providing a descriptive section regarding the feature, how it works, etc. and then we can following it with the example you made 👍

HarshithaKP:
@dougwilson, thanks. Described about `next('router')` feature followed by example. PTAL.

dougwilson:
Thank you!

_includes/api/en/5x/router-METHOD.md
```diff
@@ -40,3 +40,27 @@ router.get(/^\/commits\/(\w+)(?:\.\.(\w+))?$/, function (req, res) {
   res.send('commit range ' + from + '..' + to)
 })
 ```
+
+You can use `next` primitive to implement a flow control between different
+middleware functions, based on a specific program state. Invoking `next` with
+the name of a router will cause all the remaining route callbacks on that router
```
dougwilson:
```suggestion
the string `'router'` will cause all the remaining route callbacks on that router
```

HarshithaKP: @dougwilson, fixed it. PTAL.

gireeshpunathil (approved):

dougwilson (approved):

dougwilson:
squashed & merged

PR #1133, opened 2020-04-04
Update routing.md fixed missing quotes for route/path string

leobastin:
Fixed missing quotes for route/path string


dougwilson:
These are examples of `RegExp`s, which are delimited by the `/` character.

Issue #470, opened 2015-10-16
Update examples and docs for ES 6


crandmck:
All current docs and examples are written in ECMAScript 5.  We would like to stay current and provide docs and examples that use the latest ES 6 features.

hacksparrow:
@dougwilson what's your opinion about this?

dougwilson:
Seems OK, since people can write in whatever they like to actually use Express :) I can say that at least in corporate environments, there is virtually no ES6 knowledge, so if the examples were only ES6 it would probably confuse some percentage of our user base. Everything that can run ES6 runs ES5, so there isn't a big move to it in aware of (I'm mainly talking about all the new sugar syntax like const, let, class, etc.).

crandmck:
FYI this was suggested by @ritch.  I would think that the devil is in the details....  We certainly don't want to confuse anyone, but if there are cases where you can do things more efficiently with ES6, then we should provide an example and explanation.  This might actually take quite a bit of time and effort to do properly, so will need to work on according to priority.

ritch:
> I can say that at least in corporate environments, there is virtually no ES6 knowledge

I think we can help change that.

listepo:
You can now start rewrite on es6 (babel)

PR #1134, opened 2020-04-05
adding documentation around using promises in middleware

### LINKED ISSUE #1109: Using Promises in routes missing from 5.x docs
I just noticed when replying to https://github.com/expressjs/express/pull/4196 that the 5.x API documentation is lacking documentation around using Promises / async-await in middleware and handlers. I am just making this issue to track that and so we have an issue if someone wants to work on it.

ryhinchey:
Per the task here https://github.com/expressjs/expressjs.com/issues/1126

Closes https://github.com/expressjs/expressjs.com/issues/1109

### DIFF
--- a/en/guide/error-handling.md
+++ b/en/guide/error-handling.md
@@ -42,9 +42,27 @@ app.get('/', function (req, res, next) {
 })
 ```
 
-If you pass anything to the `next()` function (except the string `'route'`), Express regards the current request as being an error and will skip any remaining non-error handling routing and middleware functions.
+Starting with Express 5, route handlers and middleware that return a Promise
+will call `next(value)` automatically when they reject or throw an error.
+For example:
 
-If the callback in a sequence provides no data, only errors, you can simplify this code as follows:
+```js
+app.get('/user/:id', async function (req, res, next) {
+  var user = await getUserById(req.params.id)
+  res.send(user)
+})
+```
+
+If `getUserById` throws an error or rejects, `next` will be called with either
+the thrown error or the rejected value. If no rejected value is provided, `next`
+will be called with a default Error object provided by the Express router.
+
+If you pass anything to the `next()` function (except the string `'route'`),
+Express regards the current request as being an error and will skip any
+remaining non-error handling routing and middleware functions.
+
+If the callback in a sequence provides no data, only errors, you can simplify
+this code as follows:
 
 ```js
 app.get('/', [
--- a/en/guide/writing-middleware.md
+++ b/en/guide/writing-middleware.md
@@ -41,12 +41,14 @@ The following figure shows the elements of a middleware function call:
 </td></tr>
 </table>
 
+Starting with Express 5, middleware functions that return a Promise will call `next(value)` when they reject or throw an error. `next` will be called with either the rejected value or the thrown Error.
+
 <h2>Example</h2>
 
 Here is an example of a simple "Hello World" Express application.
-The remainder of this article will define and add two middleware functions to the application:
-one called `myLogger` that prints a simple log message and another called `requestTime` that
-displays the timestamp of the HTTP request.
+The remainder of this article will define and add three middleware functions to the application:
+one called `myLogger` that prints a simple log message, one called `requestTime` that
+displays the timestamp of the HTTP request, and one called `validateCookies` that validates incoming cookies.
 
 ```js
 var express = require('express')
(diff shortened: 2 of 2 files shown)


en/guide/error-handling.md
```diff
@@ -42,6 +42,49 @@ app.get('/', function (req, res, next) {
...
+app.get('/user/:id', async function (req, res, next) {
+  try {
+    var user = await getUserById(req.params.id)
+    res.send(user)
+  } catch (error) {
+    // handle the error
+    next(error)
+  }
+})
+```
```
jonchurch: I think what's most important in this doc is that rejected promises automatically call next with an error. 

So for the async await case, catching the error like this isn't really any different from how things would be done in v4. The real difference is in a case like this:

```js
app.get('/user/:id', async function (req, res, next) {
  // simulating an uncaught rejection 
  // of the promise implicitly returned by async func
  await Promise.reject("BOOM")
  res.send("ok")
})
```

In v4, `res.send` would never be called, and the error wouldn't be handled. In v5, `res.send` is still not called, but Express will catch this error and forward it to an error handler.

Tangent time: The above pattern is why I'm excited about this feature. If you can't recover from a thrown error in a route handler, say your `getPosts` controller threw, then there might not be a reason to handle the error at all in a catch. You can let it be intercepted by an error handler where you can decide how to respond. That reduces boilerplate of try/catch in handlers and makes them more elegant, IMO. Handlers become fail fast, if they can't get the data they need for a successful response, then let the error bubble up to an error handler and keep all that logic out of the response handler itself.

en/guide/error-handling.md
```diff
@@ -42,6 +42,49 @@ app.get('/', function (req, res, next) {
...
+```js
+app.get('/user/:id', function (req, res, next) {
+  return getUserById(req.params.id).then(function (user) {
+    res.send(user)
+  })
+})
+```
+
+If `getUserById` throws an error or rejects, `next` will be called with either the thrown error or the rejected value.  
+If no rejected value is provided, `next` will be called with the string `Rejected promise`.
```
dougwilson: > will be called with the string `Rejected promise`

AFAIK it is called with an actual `Error` object and not with a string.

ryhinchey: you're 100% correct. I'll update this

en/guide/writing-middleware.md
```diff
@@ -140,6 +142,82 @@ app.listen(3000)
...
 When you make a request to the root of the app, the app now displays the timestamp of your request in the browser.
 
+<h3>Middleware function validateCookies</h3>
+
+Finally, we'll create a middleware function that validates incoming cookies and sends a 400 response if cookies are invalid.
+
+Here's a simple function that validates the presence of certain cookies.
+
+```js
+function cookieValidator (cookies) {
```
wesleytodd: Should this show an actual async operation?  Just so we don't confuse beginners.

en/guide/error-handling.md
```diff
@@ -42,6 +42,49 @@ app.get('/', function (req, res, next) {
...
+
+If you want to handle the error before passing it along to `next`, you can use a try/catch block (for functions using async/await) or add a `.catch` to your Promise (or an onRejected callback to `.then`).  For example:
+
+```js
+app.get('/user/:id', function (req, res, next) {
```
wesleytodd: I think we would do better with `async`/`await` in most examples.  In this case, this might be more simple for beginners to understand:

```
app.get('/user/:id', async function (req, res, next)
  const user = await getUserById(req.params.id)
  res.send(user)
})
```

The nice thing about async await for these examples is you can drop a bunch of things which make the code harder to read and might confuse folks just learning where `async`/`await` is more common.  Thoughts?

en/guide/error-handling.md
ryhinchey: thanks @jonchurch!  I'll update this section

en/guide/error-handling.md
ryhinchey: agree @wesleytodd and I'll update.  What's the lowest supported node version for Express 5? 

While async/await is what most of the community is using at this point, given the versions of node that Express 5 supports, I wanted to make sure some examples didn't include async/await syntax so people didn't think they _had_ to use async/await.

ryhinchey: Agree about async/await being more common and I'll make the update.

dougwilson: As for the minimum version, it is no lower than 4, but likely going to be either 6 or 8. It hasn't been set in stone yet, but the guide pages today don't even support all the versions that 4 does due to the usage of arrow functions.

ryhinchey: > As for the minimum version, it is no lower than 4, but likely going to be either 6 or 8. It hasn't been set in stone yet, but the guide pages today don't even support all the versions that 4 does due to the usage of arrow functions.

Once the minimum node version for express 5 is set, I think it'd be great to go over the entire site and update it to include the most modern syntax possible.

en/guide/writing-middleware.md
ryhinchey: yes

en/guide/error-handling.md
ryhinchey: Here's an existing issue https://github.com/expressjs/expressjs.com/issues/470 related to updating to es6 syntax.  Once we  figure out the minimum supported version, someone can make the syntax consistent across the site

en/guide/writing-middleware.md
```diff
@@ -140,6 +142,53 @@ app.listen(3000)
...
+app.use(function (err, req, res, next) {
+  res.status(400).send(err.message)
+})
+
+app.listen(3000)
+
+```
+
+<div class="doc-box doc-notice" markdown="1">
+Note how `next()` is called after `await cookieValidator(req.cookies)`.  This ensures that if `cookieValidator` resolves, the next middleware in the stack will get called.  If you pass anything to the `next()` function (except the string `'route'`), Express regards the current request as being an error and will skip any remaining non-error handling routing and middleware functions.
```
dougwilson:
```suggestion
Note how `next()` is called after `await cookieValidator(req.cookies)`.  This ensures that if `cookieValidator` resolves, the next middleware in the stack will get called.  If you pass anything to the `next()` function (except the string `'route'` or `'router'`), Express regards the current request as being an error and will skip any remaining non-error handling routing and middleware functions.
```

en/guide/writing-middleware.md
```diff
@@ -140,6 +142,53 @@ app.listen(3000)
...
+async function cookieValidator (cookies) {
+  try {
+    await externallyValidateCookie(cookies.testCookie)
+  } catch {
+    throw new Error('Invalid cookies')
+  }
+}
+```
+
+Here we use [`cookie-parser`](https://github.com/expressjs/cookie-parser) to parse incoming cookies off the `req` object and pass them to our `cookieValidator` function.  The `validateCookies` middleware returns a Promise that upon rejection will automatically trigger our error handler.
```
dougwilson:
```suggestion
Here we use the [`cookie-parser`](/resources/middleware/cookie-parser.html) middleware to parse incoming cookies off the `req` object and pass them to our `cookieValidator` function.  The `validateCookies` middleware returns a Promise that upon rejection will automatically trigger our error handler.
```

ryhinchey:
thanks for the review @dougwilson!  I committed your suggestions 😄

jonchurch (approved):
LGTM

dougwilson (approved):

dougwilson:
squashed & merged

Issue #1125, opened 2020-04-01
Automated express repo docs update


dougwilson:
We need some kind of automated method (or just a tool like the readme update one) to update the page https://github.com/expressjs/expressjs.com/blob/e703270d64b20abd682e01d428bda8f068c13aca/en/resources/contributing.md

jonchurch:
Is the goal to keep this doc in step with the [`Contributing.md`](https://github.com/expressjs/express/blob/master/Contributing.md) file of the express repo?

dougwilson:
Ideally all of the docs from the express repo that are inlined on that page: `Contributing.md`, `Collaborator-Guide.md`, and `Security.md`

PR #1135, opened 2020-04-06
Automate contributing.md 

jonchurch:
Took a stab at closing #1125 

Adds a script to generate [en/resources/contributing.md](https://github.com/expressjs/expressjs.com/blob/gh-pages/en/resources/contributing.md) document by pulling in `Contributing.md`, `Collaborator-Guide.md`, and `Security.md` from the Express repo.

The script adds a Note about the origin of each part of the document, and also increments the header size of each header in the doc (making `#` become `##` etc.) so we don't end up with several different h1 elements on the page. 

**This PR doesn't contain an updated version of the contributing.md file after running the script.** I left it out, but can commit it if that's desired.

This probably isn't super portable, I'm not sure that Windows machines come with `sed` installed, so maybe it's better off if we do this in JS. I looked at the [get-readmes.sh](https://github.com/expressjs/expressjs.com/blob/gh-pages/get-readmes.sh) script we already use, which relies on the non-standard `jq`, so I just went for it using what I know.

Ideally this would be run via a github action when changes are committed to any of those three files in Express.

### DIFF
--- a/get-contributing.sh
+++ b/get-contributing.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+DEST="en/resources/contributing.md"
+
+# This script replaces the contents of a section with the contents from
+# the annotated source address.
+
+href=''
+level=''
+while IFS= read -r line; do
+  if [[ -n "$href" ]] && [[ "$line" != '#'* || "$line" == "$level"'#'* ]]; then
+    continue
+  fi
+
+  href=''
+  if [[ "$line" == '#'* ]]; then
+    title=${line##*\#}
+    level="${line:0:$((${#line} - ${#title}))}"
+  elif [[ "$line" == '<!-- SRC:'* ]]; then
+    href=${line:10}
+    href=${href% *}
+  fi
+
+  echo "$line"
+
+  if [[ -n "$href" ]]; then
+    echo
+    curl -s "$href" | sed -En '/^##|^[^#]/,$p' | sed 's/^#/&'${level:1}'/g'
+    echo
+  fi
+done <<<"$(< $DEST)" > $DEST
--- a/en/resources/contributing.md
+++ b/en/resources/contributing.md
@@ -22,36 +22,34 @@ The Express technical committee consists of active project members, and guides d
 
 ## Community contributing guide
 
-> NOTE: This is copied from the [Contributing Guide in the repo](https://github.com/expressjs/express/blob/master/Contributing.md).
+<!-- SRC: https://raw.githubusercontent.com/expressjs/express/master/Contributing.md -->
 
-The goal of this guide is to create a contribution process that:
+The goal of this document is to create a contribution process that:
 
 * Encourages new contributions.
 * Encourages contributors to remain involved.
 * Avoids unnecessary processes and bureaucracy whenever possible.
-* Creates a transparent decision making process which makes it clear how
+* Creates a transparent decision making process that makes it clear how
 contributors can be involved in decision making.
 
-This document is based on much prior art in the Node.js community, io.js,
-and the Node.js project.
-
-Vocabulary:
+### Vocabulary
 
 * A **Contributor** is any individual creating or commenting on an issue or pull request.
 * A **Committer** is a subset of contributors who have been given write access to the repository.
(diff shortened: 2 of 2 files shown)


wesleytodd:
I wonder if this would be more simple to do in a JS Action?  Have you messed with Actions yet?  The JS version is pretty straight forward, and I have some examples of getting and modifying files if you want.

jonchurch:
I agree it should be run as an action. And it's not a particularly complicated problem, whether the code is JS or sed based it's just text munging w/ regex and making an HTTP request. 

What's more interesting re: Actions is how we would trigger the workflow. I had originally envisioned the workflow running only when changes happen in the Express repo to the files we're combining. The only way I know to respond to changes external to a repo is with a [`repository_dispatch`](https://help.github.com/en/actions/reference/events-that-trigger-workflows#external-events-repository_dispatch) event. However, only workflows committed to the `master` branch can respond to that particular type of event. This site's main branch is the `gh-pages` branch currently. So as is, we can't use that trigger.

We can switch to using master instead of gh-pages to get around this particular issue (gh-pages will let you [publish from master](https://help.github.com/en/github/working-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#choosing-a-publishing-source), it's just a config change) or run the action on a different trigger.

I think we're getting into separate issue territory with all that, though. I'd love to see the [`get-readmes.sh`](https://github.com/expressjs/expressjs.com/blob/gh-pages/get-readmes.sh) script run via an action as well, in addition to generally making use of more Actions in this repo.

get-contributing.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+GHURL="https://raw.githubusercontent.com/expressjs/express/master"
+DEST="en/resources/contributing.md"
+
+# Remove everything after the <!-- marker -->
+sed -i "" '1,/<!-- marker -->/!d' $DEST
+
+curl -s "$GHURL/Contributing.md" | sed '1 i\
```
jonchurch: Note to self, this first `sed` command could really just be an `echo` appending the line into the file. All that sed command is doing is inserting the "Note" about origin of the text before the content of the curl'd file.

dougwilson:
Yep, I can definitely see an action being the north star, but I think this gets us a lot closer; it codifies what an action would need to do and it even pulls in a bunch of corrections we needed to make in the process :)

get-contributing.sh
dougwilson: Even on that note, I think we likely don't need to actually link out to it from the website; I'm not sure it _really_ adds much. I think the links were mainly there as part of it being _manual_ in order to act as a notice for us / folks trying to update it. We could likely just leave HTML comments instead and let the doc just flow on the website naturally.

dougwilson:
I also noticed that the underlying documents are not consistent in using the same heading levels... lol. We should just fix them in the express repo directly to have one h1 at the top, and h2+ for the rest of the document vs change anything here to compensate.

dougwilson:
Ok, the docs in the main express repo now have headings that make sense. It seems like there are probably just a couple tweaks that need to be made still:

1. The actual ToC at the top of the page needs to be fixed to work
2. There are points where whitespace needs to be added for Markdown to render right

Screen Shot 2020-04-14 at 9 19 40 PM

dougwilson:
Ok, just playing around to get this into something, I basically ended up with a completely different script (sorry! I did iterate from your original script). It is still called `get-contributing`, but it's pretty generic, we can probably use it everywhere... it just pulls in a Markdown file and looks for `<!-- SRC: url -->` and replaces everything within that heading section with the contents of that url with an extra indent and removing the h1 at the top of the imported file.

jonchurch:
I like the approach of pulling in file content based on that SRC comment 👍

dougwilson:
Nice. Another nice part is that if we end up moving some of the "global" docs to the website instead of the express code repo, it would just be a commit to remove the comment and that's it. It does remind me I should make it handle the SRC not being able to pull in a way that keeps existing contents and writing out a warning from the script.

dougwilson (approved):
Thanks for taking a look over the changes again @jonchurch ! Sorry it hung around without being merged, so I just ran it now to sync it up and merge it.

PR #1138, opened 2020-04-10
fix table style to block so as not to overflow into surrounding text

thomasgrz:
Had a hard time reading the Overview section at http://expressjs.com/en/guide/writing-middleware.html#mw-fig because the diagram was overflowing and hiding the preceding paragraphs.

**This is what it currently looks like:**
Screen Shot 2020-04-09 at 9 57 10 PM
**And this is what the change affects:**
Screen Shot 2020-04-09 at 9 57 26 PM

Notice: the diagram no longer overflows or obfuscates the two preceding lines of text.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1149,7 +1149,7 @@ h2 a {
 
 .callout {position: relative;}
 
-#mw-fig { padding: 0; border: 0; width: 960px; margin-bottom: 20px;}
+#mw-fig { border-collapse: separate; padding: 0; border: 0; width: 960px; margin-bottom: 20px;}
 #mw-fig-imgcell {margin: 0; padding: 0px; border: 0; width: 410px;}
 #mw-fig-img {margin: 0px; padding: 0px; width: 410px; height: 308px;}
 .mw-fig-callouts {margin: 0; padding: 0 0 0 5px; border: 0; width: 550px;}


jonchurch:
Good catch! I'm not sure how long it has been like this, or if another change brought this about (it's probably just been like this for years, the blame didn't reveal anything recent). 

I'd suggest making the change in the css file instead of as an inline style change. The line to change would be [here](https://github.com/expressjs/expressjs.com/blob/gh-pages/css/style.css#L1152).

I'm not convinced that `display: block` is the correct solution here, although it certainly does work. Looking deeper into it, there is a general style who's selector is taking precedence over the table's. It's applying -60px of margin-top, which looks to be the reason this is overflowing.

That style is: 
```css
#page-doc *:target {
    margin-top: -60px;
    padding-top: 60px;
    z-index: -1;
}
```

I mention all this just so we have context. I do think we should fix this ASAP. @ryhinchey do you have any thoughts?

thomasgrz:
I agree w/ your suggestion of targeting in the CSS file. I probably should've dug a bit deeper. I honestly just fixed on the fly in devtools while reading and then applied the solution that worked for me there.

I'm happy to adjust based on you (@jonchurch ) & @ryhinchey 's feedback!

ryhinchey:
> I mention all this just so we have context. I do think we should fix this ASAP. @ryhinchey do you have any thoughts?

@jonchurch I agree we should fix this asap and that the styling should be done in a css file, not inline.

The html and css of the site probably hasn't been looked at in years and perhaps we should take it on at some point to update it so it's more maintainable

jonchurch:
For some reason `border-collapse: separate` added to the `#mw-fig` style also works. 🤷 

To close the loop on this, @thomasgrz @ryhinchey I'm happy to accept changes to the `#mw-fig` style linked to in my comment above that implement `display: block`  or `border-collapse: separate` (if either of you have an idea which would be "better" css wise, I sure don't) in order to get this fixed.

thomasgrz:
@jonchurch @ryhinchey my apologies. Didn't mean to make ya wait a week for a one-liner. 

Went with the `#mw-fig` selector for its minimal impact and higher specificity than `#page-doc`. Thanks for the patience!

jonchurch (approved):
LGTM :shipit:

ryhinchey (approved):
LGTM!

jonchurch:
@thomasgrz Just a note, neither of us have merge permissions here so this will be merged once someone on @expressjs/expressjs-com has time 👍

dougwilson (approved):
thanks everyone

dougwilson:
Ugh, what is going on? Travis CI is running, but status is not updating on PR :(

dougwilson:
Yay, CI worked.

PR #1139, opened 2020-04-11
Update template engines list

bgub:
I removed some template engines that are either unmaintained or haven't seen development in years.

Haml.js was last committed to in 2014, h4e was last committed to in 2014, hulk-hogan is unmaintained, Combyne was last committed to in 2017, Swig is unmaintained, Webfiller was last committed to in 2012, Exbars was last committed to in 2016.

Additionally, I added an embedded JS template engine of mine called [Eta](https://eta.js.org). Since Eta is built off the same underlying code as Squirrelly, I figured the positioning made sense?


dougwilson:
> I removed some template engines that are either unmaintained or haven't seen development in years.

We don't currently have a specific rule set around this, though we are working towards one. Unless they flat out do not work with Express, please leave them until we have an actual set of rules for making such a determinitation.

> Additionally, I added an embedded JS template engine of mine called Eta. Since Eta is built off the same underlying code as Squirrelly, I figured the positioning made sense?

I would suggest splitting that change into a separate PR from this, unless you are willing to have it hang around while we determine the state of removing the ones you have removed in the PR.

bgub:
@dougwilson that makes complete sense. I created a PR to add Eta at #1140 :)

dougwilson (approved):

PR #1141, opened 2020-04-13
doc: when 'trust proxy' = true, req.ip stays a single ip

muxator:
Before this change, the documentation stated that with 'trust proxy' set, both
req.ip and req.ips would be populated with the list of addresses from
X-Forwarded-For header.

This is true only for req.ips, while req.ip will only get the left-most entry.

### DIFF
--- a/en/guide/behind-proxies.md
+++ b/en/guide/behind-proxies.md
@@ -84,7 +84,7 @@ Enabling `trust proxy` will have the following impact:
   </li>
   <li markdown="1">`X-Forwarded-Proto` can be set by the reverse proxy to tell the app whether it is `https` or  `http` or even an invalid name. This value is reflected by [req.protocol](/{{ page.lang }}/api.html#req.protocol).
   </li>
-  <li markdown="1">The [req.ip](/{{ page.lang }}/api.html#req.ip) and [req.ips](/{{ page.lang }}/api.html#req.ips) values are populated with the list of addresses from `X-Forwarded-For`.
+  <li markdown="1">The [req.ip](/{{ page.lang }}/api.html#req.ip) and [req.ips](/{{ page.lang }}/api.html#req.ips) values are populated based on the socket address and `X-Forwarded-For` header, starting at the first untrusted address.
   </li>
 </ul>
 


en/guide/behind-proxies.md
```diff
@@ -74,7 +74,7 @@ Enabling `trust proxy` will have the following impact:
   </li>
   <li markdown="1">`X-Forwarded-Proto` can be set by the reverse proxy to tell the app whether it is `https` or  `http` or even an invalid name. This value is reflected by [req.protocol](/{{ page.lang }}/api.html#req.protocol).
   </li>
-  <li markdown="1">The [req.ip](/{{ page.lang }}/api.html#req.ip) and [req.ips](/{{ page.lang }}/api.html#req.ips) values are populated with the list of addresses from `X-Forwarded-For`.
+  <li markdown="1">The [req.ip](/{{ page.lang }}/api.html#req.ip) value is got from the left-most entry in the `X-Forwarded-For` header, while [req.ips](/{{ page.lang }}/api.html#req.ips) is populated with the list of addresses from `X-Forwarded-For`.
```
dougwilson: Great catch! In fact, this is still incorrect, as in this part "enabled" means anything except `false`, so it is about `true`, `2`, `10.0.0.1/16`, etc. values. Many times it will not be the left-most entry (only in the case of `true`, which is only one of the enabled states).

muxator: Thanks, you clarified another doubt I had.

This passage talks of "enabling", but earlier in the doc it seems that this behaviour is attained when it is exactly `true` (see [the first table entry here](https://expressjs.com/en/guide/behind-proxies.html)):


Type | Value
-- | --
Boolean | If `true`, the client’s IP address is understood as the left-most entry in the X-Forwarded-* header. If `false`, the app is understood as directly facing the Internet and the client’s IP address is derived from req.connection.remoteAddress. This is the default setting.

What's the best phrasing to choose?

I am no Express pro, sorry if I may not being precise. Thanks again.

dougwilson: You're right, the page does not define enabling, as it just uses the term from how app settings work-- any value for a setting that is being enabled. The call of `app.set` or `app.enable` enables a setting while `app.disable` disables a setting.

dougwilson: We should improve the wording somehow--this section is supposed to explain what APIs of Express are altered by changing the "trust proxy" setting, if that makes sense--no matter what value it is set to.

muxator: > We should improve the wording somehow

No native speaker, sorry :)

dougwilson:
Hi @muxator are you going to take another stab at the wording? Or you can open an issue about this and we can work on a PR -- in a PR review I can only give suggested changes around the areas changed, but it seems like other areas need wording changes, so not sure a good way to go about that on this PR if you didn't want to do them 👍

dougwilson (approved):
i know this has sat here as you're not a native speaker. i took a stab at updating the wording. if this is still wrong, feel free to follow up with another pr. i just wanted to get the prs cleared out :)

PR #1142, opened 2020-04-15
Typo in generator.md

caffeineDream:
Typo, npx -> npm


dougwilson:
npx is the correct command in this case.

PR #1147, opened 2020-05-01
Fix inconsistency in the documentation of req.accepts

LoicPoullain:
According to the description above the example and to the [accepts](https://github.com/jshttp/accepts) documentation, `req.accepts` returns `false` when no types match. This is also the behavior that we observe when running the example.


dougwilson (approved):
thank you. this applies to the 5.x docs as well, so i copied it there for you

PR #1152, opened 2020-05-15
Update 4x and 5x docs to include `router(req, res, next)` documentation

### LINKED ISSUE #1151: Add API docs for router(req, res, [next])
The instance function `router(req, res, [next])` I can't find documented in the API docs. We should get it listed there and documented, somewhere in the http://expressjs.com/en/4x/api.html#router section.

The same docs would be in the 4x and 5x API docs.

SimplyComplexable:
API docs currently provide an example of using the `Router.router()` instance imperatively
as a middleware, but it isn't immediately clear that this can be used as a regular function
within a handler by explicitly passing in the arguments.

Resolves
#1151

### DIFF
--- a/_includes/api/en/4x/menu.md
+++ b/_includes/api/en/4x/menu.md
@@ -190,6 +190,7 @@
         <ul id="router-menu">
             <li><em>Methods</em>
             </li>
+            <li><a href="#router.instance">router(req, res, next)</a>
             <li><a href="#router.all">router.all()</a>
             </li>
             <li><a href="#router.METHOD">router.METHOD()</a>
--- a/_includes/api/en/4x/router-instance.md
+++ b/_includes/api/en/4x/router-instance.md
@@ -1,0 +1,43 @@
+<h3 id='router.instance'>router(req, res, next)</h3>
+
+Once a router instance has been created and all the relevant route handlers have been added
+you can use in a few different ways.
+
+```js
+const userRouter = express.Router()
+
+userRouter.get('/events', (req, res, next) => {
+  const { userId } = req.query
+  // return user specific events
+})
+
+userRouter.get('/tasks', (req, res, next) => {
+  const { userId } = req.query
+  // return user specific tasks
+})
+```
+
+  - Directly pass it to another router's route handler.
+
+```js
+app.use('/calendar', userRouter)
+```
+
+  - Or you can call pass the `req`, `res`, and `next` arguments to the `Router` instance as a function
+  within another router's route handler.
+  
+```js
+app.use('*', (req, res, next) => {
+  if (req.query.userId) {
+    // if a user id is passed in use the user specific router
+    userRouter(req, res, next)
+  } else {
+    next()
+  }
+})
+```
+
+This allows you to have a more fine-grained control over what the router instance is responsible for, 
+while still maintaining a clear separation of concerns between different functionalities. In the example 
+above, the logic is split by whether a user ID is specified in the request query params. This allows
+for easily separating logic for authenticated users versus visitors.
--- a/_includes/api/en/4x/router.md
+++ b/_includes/api/en/4x/router.md
@@ -38,21 +38,21 @@ In this example the `req`, `res` and `next` arguments are imperatively passed to
 explicitly pass them within a callback.
 
 ```js
-const userRouter = express.Router();
+const userRouter = express.Router()
 
 userRouter.get('/calendar', (req, res, next) => {
-  const { user_id } = req.query;
+  const { userId } = req.query
   // return user specific calendar information
-});
+})
 
 app.use('*', (req, res, next) => {
-  if (req.query.user_id) {
+  if (req.query.userId) {
     // if a user id is passed in use the user specific router
-    userRouter(req, res, next);
+    userRouter(req, res, next)
   } else {
-    next();
+    next()
   }
-});
+})
 ```
 
 
--- a/_includes/api/en/4x/router.md
+++ b/_includes/api/en/4x/router.md
@@ -34,6 +34,27 @@ You can then use a router for a particular root URL in this way separating your 
 app.use('/calendar', router)
 ```
 
+In this example the `req`, `res` and `next` arguments are imperatively passed to the `router` instance. You can also 
+explicitly pass them within a callback.
+
+```js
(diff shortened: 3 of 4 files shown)


dougwilson (changes requested):
Thank you so much! It looks like the CI is noting that `npm test` does not pass, from a linting issue. The code blocks should be StandardJS formatted is all.

dougwilson:
Thanks so much! there is the formatting noted above, but also we probably also want to add some API docs itself, and examples can always be used in addition to API docs. An API doc would be a section about the specific API in question (`router(req, res, next)`) and a description about what it is and what it does. Then you can (optionally) follow up with an example of it in use.

SimplyComplexable:
@dougwilson to clarify, are you just saying to add a separate file (i.e. router-function) in the same folders? Or is there a different area that your suggesting this be added to?

dougwilson:
I will need to take a look if you are looking for that level of detailed instructions. But maybe this will help: each API in express has a section that has a header listing the API, a description about the API, and optionally an example of it's usage. That is what we need for router(req, res, next) API.

dougwilson (changes requested):
An API doc would be a section about the specific API in question (`router(req, res, next)`) and a description about what it is and what it does. Then you can (optionally) follow up with an example of it in use.

These changes are adding an example to the `Router` constructor that is unrelated to that API. A new API section needs to be created with a header of the API (`router(req, res, next)`), a description about what it is and what it does, then optionally add an example to that section about said API.

SimplyComplexable:
Thanks @dougwilson, I'll be working on this today.

SimplyComplexable:
@dougwilson I added a dedicated page for this. I think this is what you were going for. Let me know if there's anything else.

SimplyComplexable:
@dougwilson any update on this?

crandmck:
Sorry I haven't been around to comment on this earlier @SimplyComplexable ...

Unfortunately, since this applies to applies to 5.x as well as 4.x, here we are seeing the cost of publishing the 5x API docs.  We have to do some manual duplication. Not end of the world, but :-( 

Anyway, just a few things that need to be done:
- Update `_includes/api/en/5x/menu.md` in the same way as `4x/menu.md` so the new heading is in the navigation.
- Add `_includes/api/en/5x/router-instance.md` that duplicates `_includes/api/en/4x/router-instance.md`

It's getting there, .... I think it's a good addition; based on what @dougwilson said in https://github.com/expressjs/expressjs.com/issues/1151, it sounds like it's omission/doc-bug.

Although I made some inline comments, I'm wondering now about the first example `app.use('/calendar', userRouter)`.  I'm still wrapping my head around the material in this section, but is it actually relevant here? If so, then it would be informative to explain how.

_includes/api/en/4x/menu.md
```diff
@@ -190,6 +190,7 @@
         <ul id="router-menu">
             <li><em>Methods</em>
             </li>
+            <li><a href="#router.instance">router(req, res, next)</a>
```
crandmck, line 193: None of the other headings in the API ref include the arguments, but rather they just have the `()` even when the method takes args.  That's just a convention, but I think we should follow it. So it would just be `router()`.

_includes/api/en/4x/router-instance.md
```diff
@@ -0,0 +1,43 @@
+<h3 id='router.instance'>router(req, res, next)</h3>
```
crandmck, line 1: If title changes, need to change here, too.

_includes/api/en/4x/router-instance.md
```diff
@@ -0,0 +1,43 @@
...
+
+Once a router instance has been created and all the relevant route handlers have been added
+you can use in a few different ways.
```
crandmck, line 4: Pls reword (combines passive voice & 2nd person).  Suggest:
"Once you create a router instance and add all relevant route handlers, you can use it in a few different ways.

_includes/api/en/4x/router-instance.md
```diff
@@ -0,0 +1,43 @@
...
+  // return user specific events
+})
+
+userRouter.get('/tasks', (req, res, next) => {
+  const { userId } = req.query
+  // return user specific tasks
+})
+```
+
+  - Directly pass it to another router's route handler.
```
crandmck, line 20: Do not use a bullet here (bullets are for list items).

I think it's technically correct, but calling `app` in example below a "router" is potentially confusing, since it's referred to elsewhere as an `app` object, not a `Router` object.

Suggest:
`Directly pass it to a method of a route handler object, such as the [app](#app) object, as shown here:`

_includes/api/en/4x/router-instance.md
```diff
@@ -0,0 +1,43 @@
...
+
+```js
+app.use('/calendar', userRouter)
+```
+
+  - Or you can call pass the `req`, `res`, and `next` arguments to the `Router` instance as a function
```
crandmck, line 26: Do not use a bullet here (bullets are for list items).

Also "...can call pass..." --> "... can pass ..."

_includes/api/en/4x/router-instance.md
```diff
@@ -0,0 +1,43 @@
...
+  within another router's route handler.
```
crandmck, line 27: This wording is very confusing to me...  It says
"...passing req, res, and next arguments to the `Router` instance as a function....",  but AFAICT the function is being passed to an `app` instance, not a `Router` instance.  

Maybe try rewording into multiple sentences ..?

_includes/api/en/4x/router-instance.md
```diff
@@ -0,0 +1,43 @@
...
+  if (req.query.userId) {
+    // if a user id is passed in use the user specific router
+    userRouter(req, res, next)
+  } else {
+    next()
+  }
+})
+```
+
+This allows you to have a more fine-grained control over what the router instance is responsible for,
```
crandmck, line 40: Consider moving the explanation above the code example.  People sometimes overlook it when it comes after, and can help sometimes to understand the code if you read the text first.  This is just a suggestion.

crandmck (changes requested):

dougwilson:
I'm going to close this due to no follow ups to the suggestions above in a while. If you would like to pick this up again, just re-open a new PR 👍

PR #1163, opened 2020-06-05
Add a note about start command in 'Process Managers' page

### LINKED ISSUE #852: Add a note about start command in "Process Managers" page
Per comment in old PR, need to clarify commands used to start apps via PMs:

> My preference would be to just add a short note in each section stating that the example command (e.g. `pm2 start app.js`) is for a simple app where`app.js` is the main script and if the app is creatd w generator, you would use a different command.

rafamarts:
Closed #852

I've added the note as suggested, for English and Portuguese. Thank you!

### DIFF
--- a/de/advanced/pm.md
+++ b/de/advanced/pm.md
@@ -147,7 +147,7 @@ Wenn Sie eine Anwendung mit dem Befehl `pm2` starten, müssen Sie den Pfad zur A
 
 <pre>
 <code class="language-sh" translate="no">
-$ pm2 start app.js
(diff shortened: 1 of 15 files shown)


dougwilson:
We should probably just change the examples to `pm2 start npm -- start` instead to avoid the note altogether, as then it would just use the `npm start` script to start the app and the file names and arrangements don't matter any longer.

dougwilson (approved):
Since I didn't hear back anything about my comment or any object to the commit I pushed, I squashed the changes to merge 👍

PR #1164, opened 2020-06-05
Update Copyright specifications.

zriyansh:


dougwilson:
Thank you, but as this pertains to US copyright notice, the year in the notice is the year of first publication.

PR #1179, opened 2020-07-20
res.send documentation change

schamberg97:
This documentation is incorrect, because res.send also accepts booleans. Example could be even found in express tests: 
```js
it('should return true when the resource is not modified', function(done){
      var app = express();
      var etag = '"12345"';

      app.use(function(req, res){
        res.set('ETag', etag);
        res.send(req.fresh); // <-- req.fresh returns Boolean
      });

      request(app)
      .get('/')
      .set('If-None-Match', etag)
      .expect(304, done);
    })
```

### DIFF
--- a/_includes/api/en/4x/res-send.md
+++ b/_includes/api/en/4x/res-send.md
@@ -2,7 +2,7 @@
 
 Sends the HTTP response.
 
-The `body` parameter can be a `Buffer` object, a `String`, an object, or an `Array`.
+The `body` parameter can be a `Buffer` object, a `String`, an object, `Boolean`, or an `Array`.
 For example:
 
 ```js
--- a/_includes/api/en/5x/res-send.md
+++ b/_includes/api/en/5x/res-send.md
@@ -2,7 +2,7 @@
 
 Sends the HTTP response.
 
-The `body` parameter can be a `Buffer` object, a `String`, an object, or an `Array`.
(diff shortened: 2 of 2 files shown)


gireeshpunathil:
LGTM

FWIW, a more compact test:

```js
require('express')().get('/', (q, r) => {
  r.send(new Boolean(true))
}).listen(12000, () => {
  require('http').get('http://localhost:12000', (m) => {
    m.on('data', (d) => console.log(d.toString()))
  })
})
```
outputs `true`.

schamberg97:
@gireeshpunathil True, just wanted to point out that even the tests depend on this feature, but it’s not even documented

dougwilson (changes requested):
Can you also update the 5x documentation as well?

schamberg97:
Umm, sure, I’ll do it in a couple of hours, when I get to my computer

dougwilson:
no rush @schamberg97 . just ping us when updated 👍

schamberg97:
> no rush @schamberg97 . just ping us when updated 👍

@dougwilson Done 👍

dougwilson (approved):

crandmck:
This looks ready to land @dougwilson ... but I won't take that liberty since I'm late to the party here.

crandmck:
OK, since Doug approved this straightforward change, and no one has objected, I'm going to go ahead and land it.

PR #1182, opened 2020-07-25
Changed hello world arrow function bodies to statements

### LINKED ISSUE #1181: Confusing hello world documentation. res.send is not a statement, but a returned expression.
On the page https://expressjs.com/en/starter/hello-world.html the following code causes confusion:

```javascript
app.get('/', (req, res) => res.send('Hello World!'))
```
Note that the `res.send('Hello World!))` is an expression here returned by the fat arrow function, because it's not enclosed in curly braces. The code is equivalent to:
```javascript
app.get('/', (req, res) => {
  return res.send('Hello World!')
})
```
(see [MDN web docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions))

Two getting started pages later (https://expressjs.com/en/starter/basic-routing.html ) the following syntax is used in  the examples, without the `return` statement:

```javascript
app.get('/', function (req, res) {
  res.send('Hello World!')
})
```

Either I'm missing something and it's OK and even encouraged to use either form. I don't mean either arrow function expression or regular function expression, but the function returning `res.send('Hello World!)` or not. 

In any case I would prefer to see the Hello World example changed to:
```javascript
app.get('/', (req, res) => {
  res.send('Hello World!')
})
```
This would avoid needless confusion.

mjvandermeulen:
In the hello world example the arrow functions were returning expressions. In order to avoid confusion and be consistent with the rest of the documentation these functions have been updated. The bodies are now statements.

fixes #1181

### DIFF
--- a/de/starter/hello-world.md
+++ b/de/starter/hello-world.md
@@ -17,16 +17,17 @@ Erstellen Sie im Verzeichnis `myapp` eine Datei namens `app.js` und fügen Sie d
 
 <pre>
 <code class="language-javascript" translate="no">
-var express = require('express');
-var app = express();
+const express = require('express')
+const app = express()
+const port = 3000
 
-app.get('/', function (req, res) {
-  res.send('Hello World!');
-});
+app.get('/', (req, res) => {
+  res.send('Hello World!')
+})
 
-app.listen(3000, function () {
-  console.log('Example app listening on port 3000!');
-});
+app.listen(port, () => {
+  console.log(`Example app listening at http://localhost:${port}`)
+})
 </code>
 </pre>
 
--- a/en/starter/hello-world.md
+++ b/en/starter/hello-world.md
@@ -18,9 +18,13 @@ const express = require('express')
 const app = express()
 const port = 3000
 
-app.get('/', (req, res) => res.send('Hello World!'))
+app.get('/', (req, res) => {
+  res.send('Hello World!')
+})
 
-app.listen(port, () => console.log(`Example app listening at http://localhost:${port}`))
+app.listen(port, () => {
+  console.log(`Example app listening at http://localhost:${port}`)
+})
 </code></pre></div>
 
 This app starts a server and listens on port 3000 for connections. The app responds with "Hello World!" for requests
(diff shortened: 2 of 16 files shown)


gireeshpunathil (approved):

gireeshpunathil:
> In the hello world example the arrow functions were returning expressions

Just to be clear, both code are functionally correct. This change makes it consistent with other documented examples.

dougwilson (changes requested):
If you can also change the same issue on the translated pages (the snippets are all in English) that would help keep it consistent.

mjvandermeulen:
@dougwilson I have pushed the code with the updates to all translations.

### Approach to changing all the other languages.

Since there were many languages that I don't speak fluently I used the following approach with one exception:
- I replaced the full content of the `<code .....javascript...></code>` block with the new english version.

#### Exception:
- The ukrainian (I assume uk stands, incorrectly for Ukraine here. Note that the ISO abbreviation is ua) version says "Hello World" in ukrainian. Out of respect for the work of the translator I have limited the changes to just update the functions that were corrected in the english version. (I would vote for a fully english version though, but that's just my opinion)

### Side effects of my approach in fixing the current issue

To save time and avoid broken code, I used the approach listed under the first bullet above, causing the following additional changes in many languages:

- Many of the translated hello worlds were still written with "normal" function expressions,  instead of arrow functions, and were therefore not in need of correction, according to this issue, but were changed anyways.
- In many translations `var`s were changed to `const`s
- Assigning the `port` constant and then using it in a string literal was not used in many languages.

dougwilson:
Thank you for the work you put into this @mjvandermeulen . I think you took the right approach here, and got the various translations at least into a more consistent state than they were before.

crandmck:
Yes, thanks @mjvandermeulen this LGTM.

@dougwilson I think he has addressed your requested changes.  I would land it, but will defer to you since you were involved earlier.

One small nit, re

> ... I assume uk stands, incorrectly for Ukraine here. Note that the ISO abbreviation is ua ...

https://www.loc.gov/standards/iso639-2/php/langcodes_name.php?iso_639_1=uk shows UK is the ISO 639-1 Language Code for Ukrainian, unless I'm missing something.

mjvandermeulen:
@crandmck I stand corrected about the uk abbreviation. I was looking into codes for countries, not for languages.

dougwilson (approved):
I squashed the commits, thank you!

PR #1185, opened 2020-07-26
Change bold to angle brackets for systemd example

### LINKED ISSUE #1171: Bold missing
In the code snippet below: 

> Here’s an example unit file to manage a Node app directly (replace the bold text with values for your system and app):

There isn't actually any bolded text.

crandmck:
Closes #1171


dougwilson (approved):
I think this is a good solution too. I copied it into the other translations that are actually still in English as well.

PR #1203, opened 2020-08-26
Added security advisory on setting trust proxy to true on production.

uzyn:
As the default is `false`, it is often tempted to simply set it to `true` on production when an express server is operating behind a proxy. This causes a common security issue where a client is now able to pass in arbitrary IPs in `X-Forwarded-For` and trick Express into accepting it as the client IP.

### DIFF
--- a/en/guide/behind-proxies.md
+++ b/en/guide/behind-proxies.md
@@ -20,6 +20,8 @@ Although the app will not fail to run if the application variable `trust proxy` 
       <td>Boolean</td>
 <td markdown="1">
 If `true`, the client's IP address is understood as the left-most entry in the `X-Forwarded-*` header.
+  
+Note that setting it to `true` on production environment might cause security issue as it would allow the client to spoof IP address. It is recommended to set it specifically to the number of hops instead _(see Number type)_.
 
 If `false`, the app is understood as directly facing the Internet and the client's IP address is derived from `req.connection.remoteAddress`. This is the default setting.
 </td>


dougwilson (changes requested):
So of course the current description is factually accurate in what it does. I think if we are going to add a note like this, it needs to explain the implications and how folks will know one way or another. For example, it says setting it to true = it allows client to spoof IP address, which is certainly not true, as it depends on the environment set up. This statement is quite misleading. On the same note, this goes on to recommend using hop count... which has the same same issue as true, where is may or may not be an issue, depending on the way in which the environment is set up.

uzyn:
Point taken. And I agree that "spoof" might be the wrong term to use.

I feel that there should be a security advisory on this as it is intuitively inclined to be set to `true` when there's a proxy in the middle, reversing the default value of `false`, without being aware of what the security implication it might bring, especially for a production set up.

Just curious, for a production set up, will it ever be preferable to have `true` vs no. of hops.

dougwilson:
Sure, I can provide a few examples, especially where I use `true` in my own production environment:

There are many places where the edge routing system will strip off certain headers on the incoming requests, like `x-forwarded-for` and then insert their own. We have a few applications that vary in their "hop depth" due to various conditions, especially around fail overs where one DC is suddenly getting the traffic coming from the gateway from another DC with a failover condition. When that happens, the native DC for the app has 2 addresses in the `x-forwarded-for` header, but the requests from the failed over DC will have 3-4 addresses, depending on what exactly the route it took.

I'm sure there are other cases as well, but at least that is my own use-case for it 👍 

As for a security advisory, I'm not a direct member of the security community, so have no idea what the sentiment or opinion of it is there. I can certainly say that of course `true` is not the default value and something the user has to explicitly enable in their app's code. This is similar to many of Node.js's own features, where a user can set something that of course can cause security issues. Of course, even if we removed `true`, users can still of course pass `() =>true`, `Infinity`, or even a really large number, among other values that are effectively the same. What exactly constitutes a security advisory, exactly, is a bit beyond me, though.

As far as this pull request: are you willing to help fix the wording within here, or would you like us to help make an alternative PR instead? Just don't want to leave this PR lingering, so wanted to get a sense for what you wanted to do with it.

uzyn:
Sure I'll fix it. I just want to try to word it to convey it best as a security advisory as I have personally seen a few Express deployment running behind a proxy or load balancer simply setting it to `true` without realizing that it raises some security concern. 

What do you think of the following?

> For production, you would generally want to set `trust proxy` to an integer indicating the number of hops (proxy servers) between the end client to express server rather than just setting it to `true`.

dougwilson:
Well, I mean, the _best_ solution is not to use the boolean or Integers, if you want to be the "safest", and instead you want to actually list out the IPs and/or subnets of the actual proxies.

Mellbabe (approved):

Mellbabe (approved):

Mellbabe (approved):

PR #1205, opened 2020-09-10
Found a redundant word in "expressjs.com/en/advanced/best-practice-se…

adyjs:
…curity.md" document

there is 2 "to" here.
"So, best practice is to to turn off the header with the `app.disable()` method:"

after correcting
"So, best practice is to turn off the header with the `app.disable()` method:"


adyjs:
Hi , everyone , this is my first time try to do a pull-request ,
so , if I do something wrong , please tell me , thanks.

rodion-arr (approved):

dougwilson (approved):
good catch!

PR #1211, opened 2020-10-02
Mention both 'route' and 'router' options for next()

jorenvandeweyer:
Mention both options to skip or jump out of a route / router with the `next` function.

### DIFF
--- a/_includes/api/en/4x/routing-args.html
+++ b/_includes/api/en/4x/routing-args.html
@@ -36,7 +36,8 @@ Callback functions; can be:
 <p>
 You can provide multiple callback functions that behave just like middleware, except
 that these callbacks can invoke <code>next('route')</code> to bypass
-the remaining route callback(s) or invoke <code>next('router')</code> to jump out of current router. You can use this mechanism to impose pre-conditions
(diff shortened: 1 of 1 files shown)


rodion-arr (approved):
LGTM

_includes/api/en/4x/routing-args.html
```diff
@@ -36,7 +36,8 @@ <h4> Arguments</h4>
 <p>
 You can provide multiple callback functions that behave just like middleware, except
 that these callbacks can invoke <code>next('route')</code> to bypass
-the remaining route callback(s). You can use this mechanism to impose pre-conditions
+the remaining route callback(s) or invoke <code>next('router')</code> to jump out of the
+current router. You can use this mechanism to impose pre-conditions
```
dougwilson, line 40: This next sentence needs to be updated now that `'router'` is included in the previous sentence, as it goes on only to explain `'route'` but not `'router'`.

dougwilson (changes requested):

dougwilson:
I'm going to close this due to no follow ups to the suggestions above in a while. If you would like to pick this up again, just re-open a new PR 👍

PR #1231, opened 2020-12-12
Clarify `req.originalUrl` behaviour with query string in request

### LINKED ISSUE #1229: Error/issues in `req.originalUrl` docs
In the `req.originalUrl` docs ([here](https://expressjs.com/en/api.html#req.originalUrl)), it says

> In a middleware function, req.originalUrl is a combination of req.baseUrl and req.path, as shown in the following example.

That's not true, however, if there's a querystring in the URL. As shown in the example right before the above sentence, `originalUrl` retains the query string, whereas `path` does not (see the example [here](https://expressjs.com/en/api.html#req.path)).

It's also not clear why this is being called out in the first place. Yes, one can demonstrate the relationship, but why is that important? Perhaps a sentence could be added explaining that.

Finally, is this relationship restricted to middleware (as the sentence implies)? It seems to hold true in my route handlers as well. (There are ways in which `req` in a middleware and `req` in a route handler [differ](https://github.com/expressjs/expressjs.com/issues/1227), so it's worth clarifying, IMHO.)

Thanks!

rodion-arr:
Fixes #1229 by updating example and adding more details of how [`req.originalUrl`](https://expressjs.com/en/api.html#req.originalUrl) behaves when query string is present in request.

### DIFF
--- a/_includes/api/en/4x/req-originalUrl.md
+++ b/_includes/api/en/4x/req-originalUrl.md
@@ -14,11 +14,12 @@ console.dir(req.originalUrl)
 // => '/search?q=something'
 ```
 
-In a middleware function, `req.originalUrl` is a combination of `req.baseUrl` and `req.path`, as shown in the following example.
+`req.originalUrl` is available both in middleware and router objects, and is a
+combination of `req.baseUrl` and `req.url`. Consider following example:
 
 ```js
-app.use('/admin', function (req, res, next) { // GET 'http://www.example.com/admin/new'
-  console.dir(req.originalUrl) // '/admin/new'
+app.use('/admin', function (req, res, next) { // GET 'http://www.example.com/admin/new?sort=desc'
+  console.dir(req.originalUrl) // '/admin/new?sort=desc'
   console.dir(req.baseUrl) // '/admin'
   console.dir(req.path) // '/new'
   next()
--- a/_includes/api/en/5x/req-originalUrl.md
+++ b/_includes/api/en/5x/req-originalUrl.md
@@ -14,11 +14,13 @@ console.dir(req.originalUrl)
 // => "/search?q=something"
 ```
 
-In a middleware function, `req.originalUrl` is a combination of `req.baseUrl` and `req.path`, as shown in the following example.
+`req.originalUrl` is available both in middleware and router objects, and is a
+combination of `req.baseUrl` and `req.url`. Consider following example:
 
 ```js
-app.use('/admin', function (req, res, next) { // GET 'http://www.example.com/admin/new'
-  console.dir(req.originalUrl) // '/admin/new'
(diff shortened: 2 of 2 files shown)


_includes/api/en/4x/req-originalUrl.md
```diff
@@ -14,11 +14,13 @@ console.dir(req.originalUrl)
 // => '/search?q=something'
 ```
 
-In a middleware function, `req.originalUrl` is a combination of `req.baseUrl` and `req.path`, as shown in the following example.
+`req.originalUrl` available both in middleware and router objects. It is a
+combination of `req.baseUrl` and `req.path` values. It will also include query
+string in case it will be present in request. Consider following example:
```
lobsterkatie: Just some minor proofreading/language smoothing:

```suggestion
`req.originalUrl` is available both in middleware and router objects. It is a
combination of `req.baseUrl`, `req.path`, and any query string included
in the request. Consider following example:
```

_includes/api/en/5x/req-originalUrl.md
```diff
@@ -14,11 +14,14 @@ console.dir(req.originalUrl)
 // => "/search?q=something"
 ```
 
-In a middleware function, `req.originalUrl` is a combination of `req.baseUrl` and `req.path`, as shown in the following example.
+`req.originalUrl` available both in middleware and router objects. It is a
+combination of `req.baseUrl` and `req.path` values. It will also include query
+string in case it will be present in request. Consider following example:
```
lobsterkatie: Same here:

```suggestion
`req.originalUrl` is available both in middleware and router objects. It is a
combination of `req.baseUrl`, `req.path`, and any query string included
in the request. Consider following example:
```

lobsterkatie (commented):
Thanks for doing this! (I am the OP of the issue this solves, so I appreciate it!)

_includes/api/en/4x/req-originalUrl.md
dougwilson: Maybe we should really just say that it is the combination of `req.baseUrl` and `req.url`? I think, looking back at the issue, this is probably how it should have been specified, as `req.path` is just `req.url` without the query string.

lobsterkatie:
> Maybe we should really just say that it is the combination of `req.baseUrl` and `req.url`? I think, looking back at the issue, this is probably how it should have been specified, as `req.path` is just `req.url` without the query string.

Ah - yeah, that's easier. Thanks, @dougwilson.

```
`req.originalUrl` is available both in middleware and router objects, and is a
combination of `req.baseUrl`, `req.url`. Consider following example:
```

@rodion-arr, what do you think?

dougwilson (approved):
Thanks, you guys rock

PR #1233, opened 2020-12-24
Replacing next(err) with next("route")

guru-apr21:
Calls to `next()` and `next("route")` indicate that the current handler is complete and in what state. I think this is the correct definition. Previously it was "Calls to `next()` and `next(err)` indicate that the current handler is complete and in what state",


dougwilson:
The text as it is is correct. The `next(err)` indicates it is in the error state.

PR #1244, opened 2021-02-10
How to build a Serverless CMS App with ExpressJS

moluwole:
This links back to a guide on ButterCMS on how to build a Serverless CMS App with ExpressJS using ButterCMS. 

_"ButterCMS is an API-based or “headless” CMS. We allow you to do the same things as WordPress in terms of enabling your clients, marketing team, and non-developers to use a friendly interface to create marketing pages, write blog posts, and manage content, but we enable it in a very different way"_

### DIFF
--- a/en/resources/learning.md
+++ b/en/resources/learning.md
@@ -56,6 +56,7 @@ self-published, February 2018.
 - [Code with Hugo blog: Express tag](https://codewithhugo.com/tags/express)
 - [Dev.to blog: Express category](https://dev.to/ghvstcode/understanding-express-middleware-a-beginners-guide-g73)
 - [LinuxStans Blog: Express tag](https://linuxstans.com/tag/express-js/)
+- [ButterCMS blog: Express category](https://buttercms.com/blog/tag/express/)
 
 ### Add your blog here!
 


3imed-jaberi (approved):

dougwilson (changes requested):
I think this is added to the wrong section. The section here is for blogs, but this looks like a documentation site. Can you clarify on if this is a blog or not?

en/resources/learning.md
```diff
@@ -56,6 +56,7 @@ self-published, February 2018.
 - [Code with Hugo blog: Express tag](https://codewithhugo.com/tags/express)
 - [Dev.to blog: Express category](https://dev.to/ghvstcode/understanding-express-middleware-a-beginners-guide-g73)
 - [LinuxStans Blog: Express tag](https://linuxstans.com/tag/express-js/)
+- [ButterCMS blog: Express category](https://buttercms.com/blog/tag/node/)
```
dougwilson: Thanks for the updated changes! I think you accidentally used the wrong tag here. The link text says "Express category" (which is what we do want to link to from Express site), but then the link itself seems to go to a general "node" tag instead of express.

dougwilson (changes requested):

en/resources/learning.md
moluwole: Sorry about this @dougwilson 
I've updated it

dougwilson (approved):

PR #1248, opened 2021-03-02
Update description of xssFilter for helmet in best-practice-security.md

cseas:
The helmet package disables the XSS filter built into web browsers by setting the `X-XSS-Protection` header's value to `0`.
https://github.com/helmetjs/helmet#reference

The filter has been retired in Chrome & Edge, and Firefox has refused to implement it due to known issues:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection

The [recommended practice](https://github.com/helmetjs/helmet/issues/230#issuecomment-614106165) is to disable the built-in filter and use alternate methods for protecting against XSS attacks like using a CSP header, sanitize inputs, etc.


dougwilson (approved):
thanks!

PR #1251, opened 2021-03-12
Add a notice for the different match-patterns between middleware and routers

kyle-chine-leismore:
All app.method() functions have linked their path parameter explanation to <https://expressjs.com/en/4x/api.html#path-examples> (And 5x API), However, it is said:

> This will match paths **starting with /abcd**: ...

Which is not true for routers (For routers, exactly matching is implemented).

So I think it is necessary to add a notice there.

### DIFF
--- a/_includes/api/en/4x/app-use.md
+++ b/_includes/api/en/4x/app-use.md
@@ -65,6 +65,9 @@ app.use(function (err, req, res, next) {
 The following table provides some simple examples of valid `path` values for
 mounting middleware.
 
+**Routers implement exactly matching for paths, instead of comparing only leading string**
+**For instance, the path `/abc` does NOT match `/abcd` in routers. (For middleware, it DOES match!)**
+
 <div class="table-scroller">
 <table class="doctable" border="1">
 
--- a/_includes/api/en/5x/app-use.md
+++ b/_includes/api/en/5x/app-use.md
@@ -65,6 +65,9 @@ app.use(function (err, req, res, next) {
 The following table provides some simple examples of valid `path` values for
 mounting middleware.
 
+**Routers implement exactly matching for paths, instead of comparing only leading string**
+**For instance, the path `/abc` does NOT match `/abcd` in routers. (For middleware, it DOES match!)**
+
 <div class="table-scroller">
 <table class="doctable" border="1">
 


dougwilson (changes requested):
Your statement here seems incorrect, but it may just be the wording. Can you perhaps show what you mean in code that demonstrates this and we can help word it better with what you are referring to?

kyle-chine-leismore:
> Your statement here seems incorrect, but it may just be the wording. Can you perhaps show what you mean in code that demonstrates this and we can help word it better with what you are referring to?

```javascript
const app = express()
const port = 3000

// API abc
app.get('/api/abc', (req, res) => {
  res.send('Hello API abc!')
})

// API abc2
app.get('/api/abc2', (req, res) => {
  res.send('Hello API abc2!')
})

app.listen(port, () => {
  console.log(`Example app listening at http://localhost:${port}`)
})
```

If the request URL is `http://localhost:3000/api/abc2`, the actual result will be: API abc2 is triggered, while API abc will not.

However the current document <https://expressjs.com/en/4x/api.html#app.get.method> said:

> app.get(path, callback [, callback ...])
> Routes HTTP GET requests to the specified path with the specified callback functions.
> 
> Arguments
> path:
> 
> The path for which the middleware function is invoked; can be any of:
>    A string representing a path.
>    A path pattern.
>    A regular expression pattern to match paths.
>    An array of combinations of any of the above.
> For examples, see Path examples <https://expressjs.com/en/4x/api.html#path-examples>. 

And at <https://expressjs.com/en/4x/api.html#path-examples>

It is said that the path matching is implemented by comparing the leading string instead of full string, which should lead to the result that both API abc and abc2 would be triggered for the request URL <http://localhost:3000/api/abc2>

> Path: This will match paths **starting** with /abcd:
```
app.use('/abcd', function (req, res, next) {
  next()
})
```

So I think it might necessary to add a note to remind readers that the URL comparison behavior is different between router and middleware.

I am not a native English speaker, so please forgive my language mistakes.

dougwilson:
Hi @kyle-chine-leismore thank you so much for that post, I really appreciate it! I guess I still don't understand what you wrote in the actual PR, though:

> For instance, the path `/abc` does NOT match `/abcd` in routers. (For middleware, it DOES match!)

What do you mean that it DOES match in middleware? If I create a middleware like `app.use('/abc', (req, res) => res.send('hi!'))` and call `GET /abc` I see the hi, but calling `/abcd` gets a 404. That seems contrary to what you wrote. Can you demonstrate the specific example you wrote in the PR?

kyle-chine-leismore:
> > For instance, the path `/abc` does NOT match `/abcd` in routers. (For middleware, it DOES match!)
> 
> What do you mean that it DOES match in middleware? If I create a middleware like `app.use('/abc', (req, res) => res.send('hi!'))` and call `GET /abc` I see the hi, but calling `/abcd` gets a 404. That seems contrary to what you wrote. Can you demonstrate the specific example you wrote in the PR?

OK. let's say the request URL is: `http://localhost:3000/abcd`.

For middleware:

```
app.use('/' ...)
app.use('/abc' ...)
```

Both of the above middlewares will be called, because the path is compared by leading string only.

Meanwhile for routers:

```
app.get('/' ...)
app.get('/abc' ...)
```
None of the above routers will be called except the path exactly matched with `/abcd`, am I right?

Oh sorry I made a mistake, I should write:

```
For instance, the path `/abcd` does NOT match `/abc` in routers. (For middleware, it DOES match!)
```

dougwilson:
Hi @kyle-chine-leismore I'm still not actually able to reproduce what you are claiming. Here is a full application:
```
$ cat app.js
const app = express()

app.use('/abc', (req, res) => res.send('abc path'))
app.listen(3000)

$ $ npm i express
+ express@4.17.1
added 50 packages from 37 contributors and audited 51 packages in 4.511s
found 0 vulnerabilities

$ node app &
[1] 483

$ curl -s http://localhost:3000/abc
abc path

$ curl -s http://localhost:3000/abcd
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>Cannot GET /abcd</pre>
</body>
</html>
```

kyle-chine-leismore:
```javascript
const express = require('express');
const app = express();

app.use('/abc', (req, res, next) => {
  console.log('USE /abc');
  next();
});

app.listen(3000);
```

`curl http://localhost:3000/abc`:

```
console.log: USE /abc
```

`curl http://localhost:3000/abcd` or `curl http://localhost:3000/ab` or anything other than `/abc` does not trigger `app.use('/abc')`

Does this mean that for `app.use`, except '/' (which will match anything), all the other paths will also be compared by full path string, not only by leading string? If so, maybe the document <https://expressjs.com/en/4x/api.html#path-examples>

Which says:

> This will match paths **starting** with /abcd:
```javascript
app.use('/abcd', function (req, res, next) {
  next()
})
```

is wrong?

PR #1256, opened 2021-04-08
use cdn.jsdelivr.net replace ajax.googleapis.com To import jQuery

cnxzq:


CNAME
```diff
@@ -1 +1 @@
-expressjs.com
+zqzyz.com/expressjs.com
```
rodion-arr, line 1: Sorry, we can't accept this change

rodion-arr (changes requested):

PR #1264, opened 2021-05-30
fix comma position

sin13:


aravindvnair99 (approved):

rodion-arr (commented):
I see same inconsistency in other places in docs. Maybe we can align all optional params style under this PR?

dougwilson:
The way it is written is the correct way to notate optional positional args. You can see this in Node.js's own documentation: https://nodejs.org/dist/latest-v16.x/docs/api/http.html

PR #1267, opened 2021-06-11
Add require('path') to starter/static-files path example

arye-eidelman:
I was going through the getting started guides and got an error that 'path' is not defined.


dougwilson (approved):
Makes sense to me :)

PR #1270, opened 2021-06-15
Deprecation of body-parser.md

elubdynasty:
This body-parser is already deprecated. Use ```  app.use(express.json()); ``` instead


dougwilson:
The body-parser module is in no way deprecated. If it was, it would be clearly stated in the readme. The express.json is just a convient short cut to using the module directly, which may be necessary if you need a different version.

PR #1276, opened 2021-07-15
--save is no longer necessary

NicolasRomeroF:
Since npm 5.0.0, installed modules are added as a dependency by default, so the --save option is no longer needed.

### DIFF
--- a/en/starter/installing.md
+++ b/en/starter/installing.md
@@ -34,7 +34,7 @@ Enter `app.js`, or whatever you want the name of the main file to be. If you wan
 Now install Express in the `myapp` directory and save it in the dependencies list. For example:
 
 ```sh
-$ npm install express --save
+$ npm install express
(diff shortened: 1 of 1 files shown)


aravindvnair99 (approved):

dougwilson:
The note below the command already says this. I'm not sure why removing the option will help anyone; at most it would cause confusion if someone is using an older version. Is there something in particular to gain by removing the option? Does having it cause issues?

NicolasRomeroF:
It didn't cause me any issues, it just got me confused at the beggining. I'm actually not very experienced with npm and node, so I been installing packages in my projects without using the --save option (because I did not know about this), and reading this got me wondering why I should use the --save option. But you're right, it does say why in the note below. The Spanish version of the docs does not say it, but that should be another pr, so I'll close this one.

PR #1277, opened 2021-07-25
documentation bug solved

Rizwan-Hasan:
**"res.locals.hello_world" =>  is not working and gives undefined**
Problem

**"res.app.locals.hello_world" => working fine.**
Working

### DIFF
--- a/_includes/api/en/4x/res-locals.md
+++ b/_includes/api/en/4x/res-locals.md
@@ -1,4 +1,4 @@
-<h3 id='res.locals'>res.locals</h3>
+<h3 id='res.app.locals'>res.app.locals</h3>
 
 An object that contains response local variables scoped to the request, and therefore available only to
 the view(s) rendered during that request / response cycle (if any). Otherwise,
@@ -9,8 +9,8 @@ authenticated user, user settings, and so on.
 
 ```js
 app.use(function (req, res, next) {
-  res.locals.user = req.user
-  res.locals.authenticated = !req.user.anonymous
+  res.app.locals.user = req.user
+  res.app.locals.authenticated = !req.user.anonymous
   next()
 })
 ```


dougwilson:
Hi, there are two locals locations. Your change is removing the docs for `res.locals`, which does exist. Just your example screenshots are expecting one to show up in the other.

Rizwan-Hasan:
> Hi, there are two locals locations. Your change is removing the docs for `res.locals`, which does exist. Just your example screenshots are expecting one to show up in the other.

Wow, thanks for the quick response. Will you kindly lead me to how to do it then?

dougwilson:
If you put a value in `app.locals` then you need to access though `app.locals`, but if instead you place the value in `res.locals`, then you would be able to retrieve thw value under `res.locals`.

Rizwan-Hasan:
Understood, thanks for clearing up my confusion. 😊

Rizwan-Hasan:
Is it ok to add this on the doc then?

dougwilson:
It is already there; the docs are about how things work, not an endless enumeration of how things do not work. Are you saying there is a section in the docs that state values you set in `app.locals` are accessible from `res.locals`?

Rizwan-Hasan:
The doc explains about creating an app.locals and accessing via req.app.locals but I tried to get it through res.locals thinking that it will inherit the object I've assigned into the app.locals. My mistake. But from this mistake, I get to know that it can be accessed through res.app.locals also which is not in the documentation. Now I ended up with the question of using it like this is standard or not.

dougwilson:
It should be in there.

> The Express application object can be referred from the request object and the response object as req.app, and res.app, respectively.

Rizwan-Hasan:
Finally, I've found it but in another section. But as long as it's standard and mentioned in the doc, I'm good to go with it. Thanks for taking your time and clearing things for me. 🥳🥳🥳

PR #1287, opened 2021-09-20
Update res type docs

jonathansamines:
Update res.type documentation to fix reference to mime lookup method


dougwilson (approved):
Good catch, thank you!

Issue #1313, opened 2021-12-10
[Documentation] [5.x] New Async Functionality Missing From Docs


xjamundx:
The API docs here:
https://expressjs.com/en/5x/api.html

And the migration guide:
https://expressjs.com/en/guide/migrating-5.html

Both currently fail to mention the killer feature of express 5, which is async/promise support middleware and routes.

Some of this might be https://github.com/expressjs/express/pull/4321 didn't ship yet, but either way, happy to start a PR with documentation changes reflecting the update. Please let me know if that would be appreciated.

euoia:
For anyone following this issue, the documentation is in the expressjs.com repository: https://github.com/expressjs/expressjs.com

The API docs are here:
https://github.com/expressjs/expressjs.com/tree/gh-pages/_includes/api/en/5x

The migration guide is here:
https://github.com/expressjs/expressjs.com/blob/gh-pages/en/guide/migrating-5.md

euoia:
I have made a start on a pull request to update the documentation: https://github.com/expressjs/expressjs.com/pull/1315

PR #1315, opened 2022-01-03
Add async functionality to docs

euoia:
This pull request adds details about the new handling of `async` middleware to the documentation.

Additionally, I have:
* Updated v5 examples to use arrow functions.
* Updates some v5 examples to demonstrate use of `async` request handlers.
* Expanded the documentation around how route callbacks are used.

One thing I wasn't sure about - I have added the `async` keyword to some handler callbacks that don't actually `await` on anything as a way of demonstrating this new functionality. I wonder whether it would be better to only use `async` callbacks in code that demonstrates use of `await`? In practise, however, I (and many others, likely) will simply mark all their request handlers as async (there's no downside, is there?).

This addresses https://github.com/expressjs/expressjs.com/issues/1313.

### DIFF
--- a/_includes/api/en/5x/routing-args.html
+++ b/_includes/api/en/5x/routing-args.html
@@ -39,6 +39,8 @@ that these callbacks can invoke <code>next('route')</code> to bypass
 the remaining route callback(s). You can use this mechanism to impose pre-conditions
 on a route, then pass control to subsequent routes if there is no reason to proceed with the current route.
 </p><p>
+When a callback function throws an error or returns a rejected promise, `next(err)` will be invoked automatically.
+</p><p>
 Since <a href="#router">router</a> and <a href="#application">app</a> implement the middleware interface,
 you can use them as you would any other middleware function.
 </p><p>
--- a/en/guide/migrating-5.md
+++ b/en/guide/migrating-5.md
@@ -40,6 +40,7 @@ You can then run your automated tests to see what fails, and fix problems accord
 
 <ul class="doclist">
   <li><a href="#path-syntax">Path route matching syntax</a></li>
+  <li><a href="#rejected-promises">Rejected promises handled from middleware and handlers</a></li>
   <li><a href="#app.router">app.router</a></li>
   <li><a href="#req.host">req.host</a></li>
   <li><a href="#req.query">req.query</a></li>
@@ -121,6 +122,12 @@ Path route matching syntax is when a string is supplied as the first parameter t
   * `/\\d+` is no longer valid and must be written as `/(\\d+)`.
 - Special `*` path segment behavior removed.
   * `/foo/*/bar` will match a literal `*` as the middle segment.
+
+<h4 id="rejected-promises">Rejected promises handled from middleware and handlers</h4>
+
+Request middleware and handlers that return rejected promises are now handled by forwarding the rejected value as an `Error` to the error handling middleware. This means that using `async` functions as middleware and handlers are easier than ever. When an error is thrown in an `async` function or a rejected promise is `await`ed inside an async function, those errors will be passed to the error handler as if calling `next(err)`.
+
+Details of how Express handles errors is covered in the [error handling documentation](/en/guide/error-handling.html).
 
 <h4 id="app.router">app.router</h4>
 


_includes/api/en/5x/app-all.md
```diff
@@ -11,7 +11,7 @@ The following callback is executed for requests to `/secret` whether using
 GET, POST, PUT, DELETE, or any other HTTP request method:
 
 ```js
-app.all('/secret', function (req, res, next) {
+app.all('/secret', async (req, res, next) => {
```
nfantone: Are we sure we  need to update **all** handlers in the docs to be `async` now? Good ol' sync handlers are still completely fine and supported. IMHO, we should change them only where it makes sense.

_includes/api/en/5x/app-use.md
```diff
@@ -97,31 +97,31 @@ app.use('/abcd', function (req, res, next) {
 This will match paths starting with `/abcd` and `/abd`:
 
 ```js
-app.use('/abc?d', function (req, res, next) {
+app.use('/abc?d', (req, res, next) => {
```
nfantone: This and all the others paths featuring modifiers no longer work with the latest `router`. We should either remove them from the docs or re-write them to use matching groups and/or params.

```suggestion
app.use('/ab(c?)d', (req, res, next) => {
```

_includes/api/en/5x/app-use.md
```diff
@@ -97,31 +97,31 @@ app.use('/abcd', function (req, res, next) {
...
+app.use('/ab+cd', (req, res, next) => {
   next()
 })
 ```
 
 This will match paths starting with `/abcd`, `/abxcd`, `/abFOOcd`, `/abbArcd`, and so on:
 
 ```js
-app.use('/ab*cd', function (req, res, next) {
+.use('/ab*cd', (req, res, next) => {
```
nfantone: Accidental `app` removed?

_includes/api/en/5x/app-use.md
```diff
@@ -97,31 +97,31 @@ app.use('/abcd', function (req, res, next) {
...
   next()
 })
 ```
 
 This will match paths starting with `/ad` and `/abcd`:
 
 ```js
-app.use('/a(bc)?d', function (req, res, next) {
+app.use('/a(bc)?d', (req, res, next) => {
```
nfantone: Side comment (no action needed): somewhat confusingly, this is still very valid and `/a(bc)?d` works as expected with the new `router`.

_includes/api/en/5x/router-param.md
```diff
@@ -80,7 +77,7 @@ const router = express.Router()
 
 // customizing the behavior of router.param()
 router.param(function (param, option) {
```
nfantone: Did you not turn this particular `function` into a fat arrow for a reason? Or just an omission?

_includes/api/en/5x/routing-args.html
```diff
@@ -26,18 +26,19 @@ <h4> Arguments</h4>
...
 <li>An array of middleware functions.</li>
-<li>A combination of all of the above.</li>
+<li>Any combination of the above.</li>
 </ul>
 <p>
-You can provide multiple callback functions that behave just like middleware, except
-that these callbacks can invoke <code>next('route')</code> to bypass
-the remaining route callback(s). You can use this mechanism to impose pre-conditions
-on a route, then pass control to subsequent routes if there is no reason to proceed with the current route.
+The callbacks are invoked in the order in which they are provided and behave in the same way as middleware. In order for the next callback to be invoked, the current one must call <code>next()</code>. If a callback doesn't call <code>next()</code> and doesn't respond to the client, the request will hang.
```
nfantone:
```suggestion
The callbacks are invoked in the order in which they are provided and behave in the same way as middleware. In order for the next callback to be invoked, the current one must call <code>next()</code>. If a callback doesn't call <code>next()</code> or doesn't respond to the client, the request will hang.
```

en/guide/migrating-5.md
```diff
@@ -126,6 +126,20 @@ In Express 4.7 and Express 5 onwards, the query parser option can accept `false`
 
 <h3>Improvements</h3>
 
+<h4 id="async.request.handlers">async request handlers</h4>
+
+Request handlers and middleware can now be be `async` functions, and they will behave in the same way as non-async handlers when an error occurs.
+
+In Express 4, it was necessary to explicitly handle rejected promises. Without handling the rejection, Node.js would exit and the client would receive an empty response. In Express 5, when middleware returns a rejected promise (for example, by using `await` inside an `async` request handler), Express behaves the same way as when an uncaught error is thrown: That is:
```
nfantone: > Without handling the rejection, Node.js would exit

Probably being too specific here, but Node.js didn't traditionally terminate the process on unhandled promises. Instead, it would print out an `UnhandledPromiseRejectionWarning`

en/guide/migrating-5.md
```diff
@@ -126,6 +126,20 @@ In Express 4.7 and Express 5 onwards, the query parser option can accept `false`
...
+
+1. The client receives an HTTP 500 error,
+
+2. The client is shown provided a stack trace (except when the `NODE_ENV` environment variable is `production`),
```
nfantone:
```suggestion
2. The client is shown the error stack trace (except when the `NODE_ENV` environment variable is `production`),
```

nfantone (commented):
Good stuff. Left some comments that I feel we should address before we hear from @dougwilson et al.

_includes/api/en/5x/app-all.md
euoia: Yes, agreed, I will revert changes to `async` where there's no `await` in the handler body.

_includes/api/en/5x/router-param.md
euoia: No reason. Fixed with 81460a22.

_includes/api/en/5x/routing-args.html
euoia: Hmm, I guess given that it's hard to understand it might be worth rewording to make things even more explicit, for example:

Every callback must either send a response, or call `next()`, allowing one of the subsequent callbacks to send a response. If execution of the series of callbacks ends without sending a response to the client (or erroring), the request will hang.

What do you think?

en/guide/migrating-5.md
euoia: You're right.

I guess I can reword to say:

If the rejection was not handled, the client would receive an empty response (as opposed to an error), and as of Node 15 the `UnhandledPromiseRejectionWarning` would throw an error.

What do you think? Still too specific?

_includes/api/en/5x/routing-args.html
nfantone: Perfect 👍🏼.

_includes/api/en/5x/app-all.md
nfantone: ..._aand_ just noted that you actually addressed this in a comment above. Apologies for missing that 🙏🏼.

_includes/api/en/5x/routing-args.html
euoia: Fixed with ba717a8d.

_includes/api/en/5x/routing-args.html
```diff
@@ -26,18 +26,19 @@ <h4> Arguments</h4>
...
 </ul>
 <p>
-You can provide multiple callback functions that behave just like middleware, except
-that these callbacks can invoke <code>next('route')</code> to bypass
-the remaining route callback(s). You can use this mechanism to impose pre-conditions
-on a route, then pass control to subsequent routes if there is no reason to proceed with the current route.
+The callbacks are invoked in the order in which they are provided and behave in the same way as middleware. In order for the next callback to be invoked, the current one must call <code>next()</code>. Every callback must either send a response, or call next(), allowing one of the subsequent callbacks to send a response. If execution of the series of callbacks ends without sending a response to the client (or erroring), the request will hang.
+</p>
+<p>
+Throwing an error or returning a rejected promise in a middleware function will cause all further route callbacks to be skipped. This allows a middleware function to be defined using the <code>async</code> keyword and to <code>await</code> on promises.</p>
```
xjamundx: I think I'd prefer to call these "async functions" instead of "defined using the async keyword" so like:

> This allows middleware to be defined using async functions

 And then maybe include as part of it some kind of explanation like this: 

> Whenever `await` is called on a promise that is later rejected, the middleware chain will be broken and the default error handler will be called with the rejected value, similar to calling `return next(err)` from inside a middleware function. 

I don't know if it needs to be called out explicitly but because of this architecture it's not necessary to wrap every express middleware or route `await` in `try/catch` which is common in most examples.

en/guide/migrating-5.md
```diff
@@ -126,6 +126,20 @@ In Express 4.7 and Express 5 onwards, the query parser option can accept `false`
 
 <h3>Improvements</h3>
 
+<h4 id="async.request.handlers">async request handlers</h4>
+
+Request handlers and middleware can now be be `async` functions, and they will behave in the same way as non-async handlers when an error occurs.
```
xjamundx: Do you mean "when an error is thrown"?

_includes/api/en/5x/routing-args.html
xjamundx: Okay that last part is covered pretty well below, sorry!

en/guide/migrating-5.md
xjamundx: Is there enough documentation around the default error handler that you can simply reference that it's the same behavior as if you throw an error or call `next(err)` inside of a route or middleware function? Instead of getting into all of these details, not the the details are bad!

nfantone: That's not a bad idea. Unsure about the details on the error handling docs, though.

en/guide/migrating-5.md
nfantone: FWIW, I personally think they both read pretty similar.

_includes/api/en/5x/app-all.md
euoia: Should be fixed with 356be22.

en/guide/migrating-5.md
euoia: There is documentation around error handling here: http://expressjs.com/en/guide/error-handling.html

It covers the Express 5 behaviour, so I think I can link to it. I'll try to reword to make it a bit less verbose.

en/guide/migrating-5.md
euoia: @xjamundx are you happy to resolve this without a change? My intention behind using `when an error occurs` was just to try to keep things simple. I didn't want to confuse the reader by implying they might need to throw the error from their code (as opposed to from some code they're calling). Errors may also come from Node.js (not sure if they would be considered `thrown`).

_includes/api/en/5x/routing-args.html
euoia: @xjamundx thanks for reviewing this. I've pushed an update with b369599 that attempts to simplify the wording, please let me know what you think.

en/guide/migrating-5.md
euoia: I've simplified the wording of the whole block given that error handling is well covered in the error handling guide with 5f3ea21. Please let me know if what you think.

xjamundx: Yeah I think the simplified version works well

_includes/api/en/5x/routing-args.html
xjamundx: Yeah that's good. I think there's a meta discussion about what's a "route" vs a "route callback" vs a "middleware function", but I don't think we need to figure that out right now. This is good

en/guide/migrating-5.md
xjamundx: Here's the case I was concerned wasn't explained properly:

```js
app.get('/data', (req, res) => {
    DB.getData().then(data => { 
        res.send(data.property.that.does.not.exist)); // uncaught exception probably
    });
});
```
vs

```js
app.get('/data', async (req, res) => {
    const data = await DB.getData();
    res.send(data.property.that.does.not.exist); // caught in the normal way
});
```

> Request handlers and middleware can now be be `async` functions, and they will behave in the same way as non-async handlers when an error occurs.

Do you see how it's not really handled the same way. The error before isn't really caught (without domains or async local storage or whatever the way is).

Would any of this additional detail be helpful?

> Request handlers and middleware can now be be `async` functions. When an error is thrown or a rejected promise is `await`ed  inside of an async request handler, those errors will be caught and passed to the default error handler as if calling `next(err)`.

en/guide/migrating-5.md
euoia: Will resolve. If you have any more comments on this section let me know.

_includes/api/en/5x/routing-args.html
euoia: I completely agree. It would be nice if somewhere (maybe the glossary) explains the differences between route callback and middleware - it would also be great if we used consistent vocabulary throughout the documentation. Do you know what the differences are? The only documented difference I'm aware of is that a route callback can call `next('route')` to skip the remaining route callbacks.

en/guide/migrating-5.md
euoia: I think that's a good improvement. I've used your text as provided in 500e0408. If you're happy, thumbs up and I'll resolve.

_includes/api/en/5x/routing-args.html
xjamundx: I think it's simply detected in the code base based on the number of arguments your handler includes. If it's 3 it's a middleware, 4 it's an error handler. Definitely something to call out clearly somewhere.

xjamundx:
What else is needed on this. Everything I looked at now looks great 👍🏻

cityturbo (approved):

xjamundx:
I was shocked to se these docs missing from the express 5 beta launch. Hope they manage to land soon. As far as I'm concerned this is the main reason for people to upgrade :D

dougwilson:
Yep. Only got so much time, haha. I got them rebased and stuff, and will be landing tonight when off work.

dougwilson:
Well, dang it, it looks like `app.param` doesn't support `async` in beta.1. I'll look into that. @euoia would you be ok with me removing those changes to land the rest, or would you want to split it up? Looks like the migration section should probably have a link added to the TOC on that page as well.

euoia:
Hey @dougwilson, thanks for looking at this and congrats on getting beta.1 out!

Looking at it again, it was a very presumptuous to change app.param to use an `async` function without actually checking whether it works. My understanding is that by using an `async` callback, the callback itself will return a promise (instead of undefined). I am curious about why app.param cares about the return type from its callback? I don't see anything in the [docs](http://expressjs.com/en/5x/api.html#app.param) about the return type of `app.param` being significant.

In any case, by all means, modify the PR as required to press on! 🎉

dougwilson (approved):
i distilled it down to just the beta.1 change. we'll get the rest landed as it arrives

PR #1316, opened 2022-01-03
Document path-to-regexp migration

euoia:
Add documentation based on the updates to path-to-regexp in https://github.com/pillarjs/router/pull/102/files.

I also took the opportunity to simplify the intro and remove references to "alpha". I'm happy to revert or extract these to a separate pull request if it's more appropriate.

### DIFF
--- a/en/guide/migrating-5.md
+++ b/en/guide/migrating-5.md
@@ -9,22 +9,20 @@ redirect_from: "/guide/migrating-5.html"
 
 <h2 id="overview">Overview</h2>
 
-Express 5.0 is still in the alpha release stage, but here is a preview of the changes that will be in the release and how to migrate your Express 4 app to Express 5.
+Express 5 is not very different from Express 4. Although the basic API remains the same, there are still breaking changes; in other words an existing Express 4 program might not work if you update it to use Express 5.
 
-Express 5 is not very different from Express 4: The changes to the API are not as significant as from 3.0 to 4.0.  Although the basic API remains the same, there are still breaking changes; in other words an existing Express 4 program might not work if you update it to use Express 5.
-
-To install the latest alpha and to preview Express 5, enter the following command in your application root directory:
+To install Express 5, execute the following command in your application directory:
 
 ```console
-$ npm install express@>=5.0.0-alpha.8 --save
+$ npm install --save express@>=5.0.0
 ```
 
-You can then run your automated tests to see what fails, and fix problems according to the updates listed below. After addressing test failures, run your app to see what errors occur. You'll find out right away if the app uses any methods or properties that are not supported.
+You can then run your tests to see what fails, and fix problems according to the updates listed below. After addressing test failures, run your app to see what errors occur. You'll find out right away if the app uses any methods or properties that are not supported.
 
 <h2 id="changes">Changes in Express 5</h2>
 
-Here is the list of changes (as of the alpha 2 release ) that will affect you as a user of Express.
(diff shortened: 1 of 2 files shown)


euoia:
@dougwilson what do you think about removing references to Express 3 in this migration guide?

> The app.router object, which was removed in Express 4, has made a comeback in Express 5. In the new version, this object is a just a reference to the base Express router, unlike in Express 3, where an app had to explicitly load it.

It seems likely to me that Express 3 is not likely to be relevant to many people at this point.

en/guide/migrating-5.md
```diff
@@ -116,6 +114,14 @@ The `res.sendfile()` function has been replaced by a camel-cased version `res.se
 
 The `app.router` object, which was removed in Express 4, has made a comeback in Express 5. In the new version, this object is a just a reference to the base Express router, unlike in Express 3, where an app had to explicitly load it.
 
+The new version of the router contains **breaking changes** to the way paths are parsed:
```
nfantone: These are great, but we should also include and list [changes from `2.0.0-beta.1`](https://github.com/pillarjs/router/pull/102/files#diff-648afe3d986261d8f2015b2b131b0e4a448d4dc6946cfde1a7a836876cee255eR19). In a nutshell, these are:

* New `?`, `*`, and `+` parameter modifiers (e.g.: `/users/:user+` now matches `/users/1`, `/users/1/2`, etc. but _not_ `/users`)
* Special `*` path segment behavior removed. This means that `*` is no longer a valid path and `/foo/*/bar` now matches a literal `'*'`. For the previous behaviour, a `(.*)` matching group should be used instead.
* Regular expressions can only be used in a matching group (e.g.: `/users/\\d+` is now invalid and should be written as `/users/(\\d+)`).

nfantone (commented):
Many thanks! Apart from the observation on missing `router` changes, this LGTM! Let's wait for maintainers to chime in.

en/guide/migrating-5.md
euoia: Ok, fantastic, I'll update the documentation.

I guess it might be worth stating that Express 5 uses path-to-regexp 6, and linking to the documentation for that. Similar to this page (see "Express uses path-to-regexp"):
http://expressjs.com/en/guide/routing.html#route-paths

Which leads a bit deeper down the rabbit hole - that page itself should possibly be updated, but is not versioned. Any suggestions on how to proceed?

I suppose it may also be worth forking http://forbeslindesay.github.io/express-route-tester/ in order to create a version for Express 5.

nfantone: +1 to all.

> that page itself should possibly be updated, but is not versioned. Any suggestions on how to proceed?

None from my side. Core maintainers should help us here!

euoia: The migration guide has been updated with 14ee27 - @nfantone please have a look and let me know if it's okay.

@dougwilson how should we update the routing guide? If we follow the error handling guide (which is also unversioned) there's a section "Starting in Express v5..." - perhaps we could do that?

I think my preference would be to have separate documentation for version 5 versus version 4, but obviously that's a larger undertaking.

euoia:
@dougwilson I realised that the migration guide is translated into other languages. Can you let me know the process for handling the translated versions?

euoia:
Argh - I changed changed the branch name in GitHub (because I used the wrong date!) and it closed this pull request. 

The updated pull request is here: https://github.com/expressjs/expressjs.com/pull/1317

PR #1317, opened 2022-01-05
Document path to regexp migration

euoia:
Add documentation based on the updates to path-to-regexp in https://github.com/pillarjs/router/pull/102/files.

I also took the opportunity to simplify the intro and remove references to "alpha". I'm happy to revert or extract these to a separate pull request if it's more appropriate.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'github-pages', '110', group: :jekyll_plugins
+gem 'github-pages'
+gem "jekyll"
+gem "webrick"
+gem "kramdown-parser-gfm"
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,3 @@
 source 'https://rubygems.org'
 
-gem 'github-pages'
-gem "jekyll"
-gem "webrick"
-gem "kramdown-parser-gfm"
+gem 'github-pages', '110', group: :jekyll_plugins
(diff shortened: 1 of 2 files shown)


euoia:
See the original version of this pull request if you need more context: https://github.com/expressjs/expressjs.com/pull/1316

euoia:
@dougwilson @crandmck The migration guide is translated into other languages. Can you let me know the process for handling the translated versions?

dougwilson:
Hi @euoia sorry for the delay. I will take a read through this, but first just wanted to respond to your question:

Translation is now only handled by community members who lend a hand; we no longer have a professional translation service. If the English page changes enough to make the translated page misleading or such, we can just delete it so that lanage sees the English version until someone translates it again.

crandmck:
Something very similar was done in https://expressjs.com/en/guide/migrating-5.html so this probably isn't needed any more.  

The only diff is that this has
```
npm install --save express@>=5.0.0
```

And [we](https://expressjs.com/en/guide/migrating-5.html) currently have

```
npm install "express@>=5.0.0-beta.1" --save
```

In practice, I think they're equivalent, but I'm no expert on npm versioning expressions.  
@jonchurch Do you think one is better than the other?

jonchurch:
cc @blakeembrey, is this accurate so far as you know?

blakeembrey:
@jonchurch Accurate, although I intend to re-add `*` back before we do the Express.js 5 release. Also, we could land the regex matching group feature in express 4 when we merge https://github.com/expressjs/express/pull/5603.

crandmck:
Sorry to be dense, but can you clarify which you prefer for "Upgrading to 5.x" in the doc?

- The version as specified in this PR `express@>=5.0.0-beta.1`
- What we have now `express@>=5.0.0`
- Either is OK

bjohansebas:
Closed in favor of #1610

PR #1320, opened 2022-01-28
Fix inccorect word

selrond:


dougwilson:
Thank you, but the wording is correct as it is. It is listing two important things, not one thing that does another thing as you changed it to.

selrond:
@dougwilson thanks for clarification. Although this structure looks really strange to me (not a native speaker, though), I’m glad to see it’s intentional and justified.

PR #1322, opened 2022-02-06
Remove react from view engines (it is unmaintained)

coffeeispower:


dougwilson (changes requested):
Thanks! Can you update your PR to just remove it? We don't list broken, deprecated, or unmaintained engines.

coffeeispower:
Done!

dougwilson (approved):

PR #1331, opened 2022-03-28
Typo?

caub:


dougwilson:
Hello, and thank you for your pull request. I double-checked where your suggestion is, and it is indeed not a typo. The values start at the first untrusted address, skipping all the trusted addresses that were set.

PR #1336, opened 2022-04-18
refactor: replaced deprecated `var` keyword with a `const` one in v4 docs

doichev-kostia:


dougwilson:
Hello, and thank you for your pull request. The 4.x API is meant to work with the versions of Node.js the 4.x API actually supports. You'll find the 5.x API docs have migrated from `var` to `const`/`let` since the supported version of Node.js have changed as well.

PR #1337, opened 2022-04-23
Installing express 5 without piping into a file

ubershmekel:
I was trying to install express version 5 and I couldn't figure out why I was getting version 4, and a file called `=5.0.0-beta.1` was getting created. It had this content:

```

added 1 package, and audited 72 packages in 751ms

2 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
```

After messing around with different terminals (cmd, git bash), and trying it out in a different, empty project, I finally realized the `>` bracket was piping.

### DIFF
--- a/en/guide/migrating-5.md
+++ b/en/guide/migrating-5.md
@@ -14,7 +14,7 @@ Express 5.0 is still in the beta release stage, but here is a preview of the cha
 To install the latest beta and to preview Express 5, enter the following command in your application root directory:
 
 ```console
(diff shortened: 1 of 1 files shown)


dougwilson (changes requested):
Ah, you are right. Guess it depends on the shell. Just needs to be wrapped in quotes.

ubershmekel:
Added quotes instead.

dougwilson (approved):

PR #1338, opened 2022-05-13
a word 'instead' occurring two times

JYDP02:
a word 'instead' occurring two times and making the sentence ambiguous.


3imed-jaberi (approved):

dougwilson (approved):
great catch

PR #1341, opened 2022-05-21
fix navbar line-height in css/style.css

yoomaxxx:
Hello, this commit fix line-height in expressjs.com navbar


dougwilson (approved):
nice catch, thanks!

PR #1349, opened 2022-07-06
Clarification Added: req-fresh.md5

gulgulia17:
More details added a better understanding, of how it works.


_includes/api/en/4x/req-fresh.md
```diff
@@ -2,12 +2,19 @@
 
 When the response is still "fresh" in the client's cache `true` is returned, otherwise `false` is returned to indicate that the client cache is now stale and the full response should be sent.
 
+In order to check weather the response is still "fresh" in the client's cache or not, 2 Tags will be used ETag and If-None-Match, ETag will be fetched from response (saved internally from application) and other tag will be recieved from client's request.
```
srkds:
```suggestion
In order to check whether the response is still "fresh" in the client's cache or not, two tags will be used ETag and If-None-Match. ETag will be fetched from response (saved internally from application) and other tag will be received from client's request.
```

muditsunarctech (approved):

gulgulia17:
Is anyone checking this PR?

dougwilson:
Hi, sorry about that!

I have two main comments so far:

1. We should edit the 5x docs with the same updates as here, as they are the same currently.
2. These seem incomplete, as noteably it leaves out the timestaml checking. Also as a nit the if-none-match is actually a list of tags, not just a single tag.

waleedtariq109 (commented):
Everything looks good to me; feel free to merge. @crandmck

_includes/api/en/4x/req-fresh.md
```diff
@@ -2,12 +2,19 @@
...
+In order to check whether the response is still "fresh" in the client's cache or not, two tags will be used ETag and If-None-Match. ETag will be fetched from response (saved internally from application) and other tag will be received from client's request.
+
 When a client sends the `Cache-Control: no-cache` request header to indicate an end-to-end reload request, this module will return `false` to make handling these requests transparent.
 
 Further details for how cache validation works can be found in the
 [HTTP/1.1 Caching Specification](https://tools.ietf.org/html/rfc7234).
 
 ```js
+
+// Get value for ETag where you saved it initially, In general you
```
jonchurch, line 14: This comment isn't very clear. The added example is setting an etag not reading a value.
users do not need to manage or store timestamps to get etags working, so I think this will confuse people

_includes/api/en/4x/req-fresh.md
```diff
@@ -2,12 +2,19 @@
 
 When the response is still "fresh" in the client's cache `true` is returned, otherwise `false` is returned to indicate that the client cache is now stale and the full response should be sent.
 
+In order to check whether the response is still "fresh" in the client's cache or not, two tags will be used ETag and If-None-Match. ETag will be fetched from response (saved internally from application) and other tag will be received from client's request.
```
jonchurch, line 5: The headers which go into determing fresh are seen here in [jshttp/fresh](https://github.com/jshttp/fresh/blob/206d8c5dbe5022d05f5473481852e393bbb2194e/index.js#L33C1-L85C2)

* If-None-Match
* Etag
* Cache-Control: no-cache
* If-Modified-Since
* Last-Modified

_includes/api/en/4x/req-fresh.md
```diff
@@ -2,12 +2,19 @@
...
 
 Further details for how cache validation works can be found in the
 [HTTP/1.1 Caching Specification](https://tools.ietf.org/html/rfc7234).
 
 ```js
+
+// Get value for ETag where you saved it initially, In general you 
+// may use a file to store timestamps with API request name
+
+res.set('ETag', "foo")
```
jonchurch, line 17: The example that existed wasn't very clear to begin with. Setting an etag here doesn't mean that req.fresh will be true, so the example gets less clear with the addition.

It's easier perhaps to demonstrate that setting a new, highly random etag on every request can guarantee that req.fresh is false. That way we create an example which will always have the demonstrated output.

jonchurch (changes requested):
I think that this doc can be improved, but the proposed changes don't make the subject clearer.

Cache management when it comes to 304s and Etags is handled by default by express. It's a nontrivial job which is abstracted away in the default case, and users still have the tools to manage it themselves if they so choose. `req.fresh` is one of those tools.

An improvement would be to better document what "fresh" means here, and how express determines it. We don't really want to encourage folks to try to set etags themselves via examples unless they know what they're doing.

_includes/api/en/4x/req-fresh.md
chrisdel101, line 5: Hey @gulgulia17, 
@crandmck said I could leave any feedback for your PR. I'm not a maintainer here or anything, so you don't have to listen to me :]. It's all just punctuation related stuff. 
1. Add a colon to this:
`In order to check whether the response is still "fresh" in the client's cache or not, two tags will be used: ETag and If-None-Match.`
2. Add a comma in the middle + missing 'the':
`ETag will be fetched from response (saved internally from application), and the other tag will be received from client's request.`

Issue #1350, opened 2022-07-06
Adjust icon size when screen width is under 900px


y-liang:
URL: http://expressjs.com/en/resources/middleware/session.html

Possible solution:
Change img tags' width in the CSS
```
@media (max-width: 899px)
.content p img {
    width: 98%;
    height: 98%;
}
```

Screenshots:
<img width="927" alt="image" src="https://user-images.githubusercontent.com/45440333/177649020-effb6143-60c6-45f0-bc04-89d8f81f0bc9.png">
<img width="934" alt="image" src="https://user-images.githubusercontent.com/45440333/177648916-701a8480-b7ad-4b8c-ae58-de978899f216.png">

PR #1351, opened 2022-07-06
fix package info icon size on smaller screens

y-liang:
#1350

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -943,11 +943,6 @@ h2 a {
     width: 90%;
   }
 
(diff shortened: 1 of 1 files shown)


dougwilson:
Ah, it looks like this particular rule was for something that used to be on the homepage and is no longer needed.

dougwilson (changes requested):
Please remove the rule rather than commenting it out.

dougwilson (approved):

PR #1355, opened 2022-07-25
Update link to the OWASP page

rakhman-ibrayev:
The old link to the OWASP's page about common web vulnerabilities does not work and responds with 404. Update the old link with the new one.


dougwilson (approved):
thanks

PR #1356, opened 2022-07-28
Offer html-express-js as a new template engine

markcellus:
This adds [`html-express-js`](https://www.npmjs.com/package/html-express-js), a new template engine I've created that I'm now using on a few projects. I thought it would be a great idea to offer this up to the rest of the community.

At a basic level, it renders either static HTML pages or HTML pages with dynamic content that gets interpolated using only native JavaScript (Template literals).


Redyou75 (approved):

dougwilson (approved):
awesome, thank you for this!

PR #1359, opened 2022-08-20
Add api-error-handler docs

mifi:
and example code

I was digging for an hour or so trying to find such a middleware, only to realise that it was provided by express all along!

### DIFF
--- a/en/guide/error-handling.md
+++ b/en/guide/error-handling.md
@@ -189,6 +189,31 @@ function errorHandler (err, req, res, next) {
 Note that the default error handler can get triggered if you call `next()` with an error
 in your code more than once, even if custom error handling middleware is in place.
 
+## JSON API error handler
+
+The default error handler returns HTML, but one might want to instead return JSON when designing a pure API instead of a website. Express provides a package for this, [api-error-handler](https://github.com/expressjs/api-error-handler). It can be used in conjunction with a package like [http-errors](https://github.com/jshttp/http-errors) in order to easily throw HTTP errors with a status code with metadata and have them automatically returned by the API. For example:
+
+```js
+const createError = require('http-errors')
+const errorHandler = require('api-error-handler')
+
+app.get('/api/endpoint', (req, res) => {
+  throw createError(400, 'Invalid data sent')
+})
+
+app.use(errorHandler())
+```
+
+This will return a 400 with the following JSON response:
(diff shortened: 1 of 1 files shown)


dougwilson (changes requested):
There are many modules that do this and similar. The guide here is around how to write your own rather than linking to other middlewares. It should likely be added to this page: https://expressjs.com/en/resources/middleware.html

dougwilson:
Oh, I looked at the history and it was removed because the author no longer maintains it. I will archive the repo in the org. I don't have access to the npm module itself, so nothing I can do to mark it there besides the repo.

mifi:
Ok, thanks for cleaning up. I couldn't find any other npm package that does this, in a simple clean way.

I created a fork [error-handler-json](https://github.com/mifi/error-handler-json) that I upgraded to latest deps and intend to maintain. Any chance you could link to that from the original archived project's readme, so that any lost souls can find an alternative?

Also, do you think I can submit a PR to add this new fork to https://github.com/expressjs/expressjs.com/blob/gh-pages/en/resources/middleware.md

And possibly a line in https://github.com/mifi/expressjs.com/blob/gh-pages/en/guide/error-handling.md under "The default error handler" saying something like "If you want to return errors as JSON instead of HTML, you may look into this module"?

Thanks!

dougwilson:
Ok, that's great!

> I created a fork [error-handler-json](https://github.com/mifi/error-handler-json) that I upgraded to latest deps and intend to maintain. Any chance you could link to that from the original archived project's readme, so that any lost souls can find an alternative?

I'll get the readme edited on the repo here.

> Also, do you think I can submit a PR to add this new fork to https://github.com/expressjs/expressjs.com/blob/gh-pages/en/resources/middleware.md

Feel free to send over a PR adding it to the middleware page ❤️ 

> And possibly a line in https://github.com/mifi/expressjs.com/blob/gh-pages/en/guide/error-handling.md under "The default error handler" saying something like "If you want to return errors as JSON instead of HTML, you may look into this module"?

It's going to be a bit beyond the guide and from experience is going to cause some issues with folks who want to start adding on there. Ideally it should be a guide for how to use Express.js error handling, such that they could _create_ such a module. If anything pointing elsewhere, it could point to the middlewares page so that would also keep a single place where the middleware are listed.

PR #1366, opened 2022-09-09
fix typo in req-query.md

fetsorn:
`app.setting('query parser', ...)` should be `app.set('query parser', ...)`. 
same as #1361, but for both 4x and 5x
@dougwilson


UlisesGascon (approved):

UlisesGascon:
@dougwilson: note that #1361 already cover the v4 example, but not v5.

PR #1369, opened 2022-10-02
Replace fast_star paths with path patterns

aelmardhi:
fast star is no longer supporrted
using path patterns to allow match all


IamLizu (approved):
LGTM 🚀 

Let's get it landed.

Edit: I removed the "needs tech review" label as I can confirm the changes are correct.

crandmck:
Thanks for reviewing/confirming this fix @IamLizu !

crandmck (approved):

PR #1374, opened 2022-11-14
Update error-handling.md

mishrasur7:
Inside the Catching Errors section, the sentence starting from line 64 to 65 it says 'If the callback in a sequence provides no data, only errors, you can simplify this code as follows:'. This is little confusing for me. would it make more sense if the sentence is modified like this: 'If the callback in a sequence provides no data, only errors, you can simplify write code as follows:'. 

I might be wrong but please make me clear incase if so.

### DIFF
--- a/en/guide/error-handling.md
+++ b/en/guide/error-handling.md
@@ -62,7 +62,7 @@ Express regards the current request as being an error and will skip any
 remaining non-error handling routing and middleware functions.
 
 If the callback in a sequence provides no data, only errors, you can simplify
-this code as follows:
+write code as follows:
 
 ```js
 app.get('/', [


en/guide/error-handling.md
```diff
@@ -62,7 +62,7 @@ Express regards the current request as being an error and will skip any
 remaining non-error handling routing and middleware functions.
 
 If the callback in a sequence provides no data, only errors, you can simplify
-this code as follows:
+write code as follows:
```
dougwilson, line 65: Thank you, but unfortunately the phrase "you can simplify write code as follows" is not proper English. Is there another way you you would like to modify this?

dougwilson (changes requested):

en/guide/error-handling.md
mishrasur7: Oh sorry, I read that word simplify as simply, thats my eye error which I realised now. I would modify like this now -> 
'If the callback in a sequence provides no data, only errors, you can simplify the code as follows:' However original line also makes sense. 
Again sorry about my eye error.

crandmck:
I believe the text is OK "as is" and there are many higher-priority tasks at hand.

PR #1377, opened 2022-11-27
Update link: you must use http://localhost

kamil-leys:
You can't translate http://localhost:3000/ in french, you have to let it unchanged, else it won't work.


kamil-leys:
You can't translate an link in french, you have to keep it http://localhost:3000/

PR #1385, opened 2023-01-03
Update error-handling.md corrected mistake with example

WORMSS:
The example supplied example did not `return` a promise, only created a promise and returned nothing.

I suspect this was originally wrote as `(req, res, next) => Promise.resolve().....` but was changed to `=> {}` for more readability, but failed to notice the change in behavior that this caused.

### DIFF
--- a/en/guide/error-handling.md
+++ b/en/guide/error-handling.md
@@ -104,7 +104,7 @@ that return promises.  For example:
 
 ```js
 app.get('/', (req, res, next) => {
-  Promise.resolve().then(() => {
+  return Promise.resolve().then(() => {
     throw new Error('BROKEN')
   }).catch(next) // Errors will be passed to Express.
 })


UlisesGascon:
Thanks for the suggestion @WORMSS, but the code is correct. As express is using callbacks, we should pass `res` and `next` to the promise to handle the async operation.

In this example, the `res` is not used as throw will occur first (check the next snippet).

```
app.get('/', (req, res, next) => {
  Promise.resolve().then(() => {
    throw new Error('BROKEN')
    res.send("hello")
  }).catch(next) // Errors will be passed to Express.
})
```

As you can see the `next` function is passed to the `.catch()` function to handle the errors generated in the previous `.then()`. Maybe we can improve the example to make it more clear, but not sure how.

Feel free to re-open this PR with a different suggestion 🙂

PR #1391, opened 2023-01-28
Fix typo in `req.query` code example

JekRock:
`app.setting` method does not exist. It seemed like a typo that should be changed to `app.settings`. But `app.settings` is an object, not a function. So I believe, it was intended to use `app.set` in that example


AhsanShihab:
Came here to report the same thing. Tagging @dougwilson as he merged the latest commit.

UlisesGascon:
Thanks @JekRock, but I will close this PR as is duplicated (#1361)

Issue #1397, opened 2023-02-12
Font imbalance of "Route Parameters" under the Guide


faisal-mujeeb:
font size of heading "Route Parameters" is slightly low as compared to others respective headings like for eg. Route handle , Route Path etc 
to be more specific visit
 https://expressjs.com/en/guide/routing.html

bjohansebas:
Fixed by #1481

PR #1399, opened 2023-02-16
Standardize header size

chris-kreidl:
Addresses #1397


UlisesGascon (approved):
Thanks @chris-kreidl! Great catch 👍

PR #1402, opened 2023-03-05
Added documentation for calling acceptsLanguages without argument

### LINKED ISSUE #1401: Documentation bug in request.acceptsLanguages
In the documentation (express 4.x) for [`req.acceptsLanguages`](https://expressjs.com/en/4x/api.html#req.acceptsLanguages), it states that:

>Returns the first accepted language of the specified languages, based on the request’s Accept-Language HTTP header field. If none of the specified languages is accepted, returns false.

However, the [code path](https://github.com/expressjs/express/blob/4.x/lib/request.js#L181) delegates this call to [jshttp/accepts](https://github.com/jshttp/accepts/blob/master/index.js#L207) which returns all requested languages from the `Accept-Language` HTTP header field.

We currently depend on this behavior to get the client's first language (we of course guard against `*` and missing header) from a request.

It appears to be equivalent for version 5.x: https://github.com/expressjs/express/blob/5.x/lib/request.js#L174

Can we please correct the documentation? I'm also afraid that this could break our use-case in the future, since returning `false` if no `language` argument is supplied will break our implementation.

PS. We can currently handle an empty array but not `false`. Using express 4.18.2

dotnetCarpenter:
Calling `req.acceptsLanguages` without argument(s) returns an `array` of languages from the HTTP `Accept-Language` header.

This is much easier than calling `req.get('Accept-Language')`, because you do not need to parse the `Accept-Language` header `string`.

Close #1401

### DIFF
--- a/_includes/api/en/4x/req-acceptsLanguages.md
+++ b/_includes/api/en/4x/req-acceptsLanguages.md
@@ -1,7 +1,15 @@
-<h3 id='req.acceptsLanguages'>req.acceptsLanguages(lang [, ...])</h3>
+<h3 id='req.acceptsLanguages'>req.acceptsLanguages([lang, ...])</h3>
 
 Returns the first accepted language of the specified languages,
 based on the request's `Accept-Language` HTTP header field.
 If none of the specified languages is accepted, returns `false`.
 
+If no `lang` argument is given, then `req.acceptsLanguages()`
+returns all languages from the HTTP `Accept-Language` header
+as an `Array`.
+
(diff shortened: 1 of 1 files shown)


UlisesGascon (approved):
Thanks @dotnetCarpenter. It took me a while to follow the change, but seems clear now that I checked the test file from accepts (https://github.com/jshttp/accepts/blob/master/test/language.js#L12) following the expressjs code (https://github.com/expressjs/express/blob/2a00da2067b7017f769c9100205a2a5f267a884b/lib/request.js#L181).

This is a very useful change in the docs! 👍

crandmck (approved):

Issue #1318, opened 2022-01-20
Hello world example contains misleading binding information


lextm:
In `hello-world.md`, the code snippet below
```javascript
const express = require('express')
const app = express()
const port = 3000

app.get('/', (req, res) => {
  res.send('Hello World!')
})

app.listen(port, () => {
  console.log(`Example app listening at http://localhost:${port}`)
})
```
actually binds to all available network interfaces, based on [documentation](https://nodejs.org/api/net.html#serverlistenport-host-backlog-callback) and testing.

Therefore, claiming "Example app listening at http://localhost:xxxx" is then clearly misleading. If the goal is to listen only at localhost, the correct code should be,

```
app.listen(port, 'localhost', () => {
  console.log(`Example app listening at http://localhost:${port}`)
})
```

dougwilson:
Hello, and thank you for your issue. The console log is just there so give you a link to access it; it is not making any statements about what the bindings are. Infact, localhost is a DNS name that may resolve to 127.0.0.1 or ::1 depending on various factors. Unfortunately the change you suggest does not work on all environments, as Node.js and browsers do not always agree on which to use under dual stack.

lextm:
You have the last word, but tons of other tools/frameworks decide to bind to localhost (`ng serve` for Angular, Gunicorn for Python, or Kestrel for .NET Core), so when the sample code binds to `[::]` (and 0.0.0.0), some users might be surprised.

I am also curious on how often "Node.js and browsers do not always agree on which to use under dual stack", because anyway the two resolve localhost using the same DNS configuration or `hosts` file.

dougwilson:
If you want to discuss how Node.js listens by default, you should be opening an issue with Node.js, not here.

PR #1407, opened 2023-03-14
Update hello-world.md

SubhanAliZaidi:
Changed console.log(`Example app listening on ${port}`) to console.log(`Example app listening at http://localhost:${port}`) in app.listen function.


dougwilson:
Hello, unfortunately that leads to confusion and it will not always work given different system configurations see #1318

PR #1423, opened 2023-06-03
Create SECURITY.md

als-alt:
كيفي


dougwilson:
Thank you, though this is just a static web site, so there is no need for a file like this on the website repo.

Issue #1382, opened 2022-12-26
dark mode for website


MehfoozurRehman:
would like to help here with making a dark theme for expressjs.com

dougwilson:
Sounds awesome! I look forward to your contribution!

hugoovf:
Just remove the background image every time u join a page of the website, or make a pr removing the file xd
![image](https://user-images.githubusercontent.com/69076992/212496121-d6cd658c-6a0b-4f19-9e2b-89922eb09383.png)

tabishnaqvi1311:
Hello @MehfoozurRehman , I noticed that you opened this issue three months ago about adding dark mode. I'm interested in contributing to the project and was wondering if you are still working on this issue or if you would be willing to let me take a look at it? Please let me know if there's anything I can do to help.

codesmith25103:
Hey @dougwilson, is this issue still open? If yes, can you please assign me this issue?

faisalsaifii:
What's pending here?

MehfoozurRehman:
Hey everyone,

Apologies for my absence last year due to studies. I'm comfortable with the MERN stack but not Ruby. However, I use Express.js daily and am eager to implement dark mode. Can someone guide me on running the project locally? I'm ready to contribute.

Thanks,

MehfoozurRehman:
ok so last year I tried doing the local run according to repo readme and again today i tried but 

![image](https://github.com/expressjs/expressjs.com/assets/68416499/a1b4adda-1d25-4d82-994c-de4f095f7067)

its giving me this error for some reason to make sure it's not my local pc i am using GitHub code spaces.

![image](https://github.com/expressjs/expressjs.com/assets/68416499/3e30f555-358a-4f23-bbe8-09451e881843)
![image](https://github.com/expressjs/expressjs.com/assets/68416499/b8fcaeea-ecb9-4ef7-b9e3-32caaa03d026)

these is the result of first steps I am not really familiar with the ruby ecosystem so please let me know what i am doing wrong.

MehfoozurRehman:
and i have read the issues created based on this one so what i am thinking once i have this running locally 

![image](https://github.com/expressjs/expressjs.com/assets/68416499/216aece2-b737-411b-bd6a-ffa41261d7c9)

the area where i have marked red i am thinking of adding 

![image](https://github.com/expressjs/expressjs.com/assets/68416499/0b03bead-bfa1-4c71-88eb-cb43351933c1)
[...]

PR #1427, opened 2023-06-14
Fix: Removed background image form style.css

YashNuhash:
**Fixes** #1382

Sometimes it creates eye strain when you are learning form a white backgrounded documentation since 4 to 6 hours a day.  That is why, I removed that white background image and the website is looking very good now.

> One important thing I need to mention that people who don't have enabled **Auto Dark Mode for Web Contents** they will still enjoy the same view as it is. But, like me those who enabled that they would view this website like below ...

### Examples:

image

image

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -14,7 +14,7 @@
 
 body {
   font: 400 14px/1.6 "Open Sans", sans-serif;
-/*   background: url(/images/bg.jpg); */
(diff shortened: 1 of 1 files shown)


dougwilson:
Thank you very much! Can you use a media query to make this conditional for when in dark mode at least? The background image is not white, but more muted to ease reading, but this makes it white again.

PR #1437, opened 2023-08-25
Update routing.md

teboho:
Link clips to the main (head-ed) subsections

### DIFF
--- a/en/guide/routing.md
+++ b/en/guide/routing.md
@@ -34,7 +34,7 @@ app.get('/', (req, res) => {
 })
 ```
 
-<h2 id="route-methods">Route methods</h2>
+<h2 id="route-methods">Route methods <a href="#route-methods">🔗</a></h2>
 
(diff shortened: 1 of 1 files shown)


UlisesGascon:
Hello @teboho,

I appreciate your contribution. However, I have reservations about how much value it will add, considering that GitHub already includes the link in the titles when you hover over them. Additionally, other documents in the repository do not follow this approach.

Please consider these points, and let me know if you have any further thoughts or if there's anything specific you'd like to discuss.

**For reference (on hover effect)**
Screenshot 2023-11-23 at 19 47 45

dougwilson:
I assume it is for the website itself, which doesn't have an easy way like this. But I don't think manually adding it all over the docs is good idea. The rendering should be adjusted to include them, imo.

teboho:
This makes sense. In retrospect 'manual add' of such will make things complicated in future.

PR #1446, opened 2023-11-23
Fix missing character

stevejkang:
Fix missing character in Korean translation document.


UlisesGascon (approved):
Thanks @stevejkang!

LGTM, I checked the change in Google translator as my korean is limited ☺️

dougwilson (approved):

PR #1447, opened 2023-11-25
Fix duplicate link in menu.md

mark-mxwl:
On the API Methods menu, `app.get` appears twice. As the duplicate link only increments the menu by one item, and offers no distinction between `app.get` and `app.get.method`, the duplicate can therefore be removed to improve navigation.

### DIFF
--- a/_includes/api/en/5x/menu.md
+++ b/_includes/api/en/5x/menu.md
@@ -42,8 +42,6 @@
             </li>
             <li><a href="#app.get">app.get()</a>
             </li>
-            <li><a href="#app.get.method">app.get()</a>
-            </li>
             <li><a href="#app.listen">app.listen()</a>
             </li>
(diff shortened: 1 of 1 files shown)


UlisesGascon (approved):

mark-mxwl:
Thx for the quick review, @UlisesGascon !

dougwilson:
The two `app.get` are different, though. One gets a setting and the other adds a route. The 4.x docs are like this too, and this only changes 5.x. Not sure what the best solution is.

mark-mxwl:
Hey @dougwilson ! Yes, the two are different, but the documentation makes the distinction, where the navigation does not. So it appears like a duplicate, and doesn't really serve the user. If you'd prefer for them to be separated, maybe we could disambiguate? `app.get(name)` and `app.get(path, callback)`?

PR #1463, opened 2024-03-02
fix: temporary change to enable CI on review submission

wesleytodd:
This is a change (ideally temporary) to run ci when we approve PRs. There was a conflicting setting to not allow merges when CI hadn't run, but also the branch protections were set to require CI to merge. We could just override, but why have CI if we are not going to run it?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,10 +7,16 @@ on:
   push:
     branches:
     - gh-pages
+  pull_request_review:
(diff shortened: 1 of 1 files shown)


wesleytodd:
And that's what I get from not testing this before pushing. Anyway, I think this added check worked.

crandmck (approved):
Since I'm a GH workflow noob, I'm just going to take your word that this works as it should! :-)

PR #1476, opened 2024-03-22
removed <pre> tag reason syntax hilighting

sabeerbikba:
# before
before

# after 
after


crandmck:
This is intentional, because it's illustrating HOW to create a fenced code block in source, not .  It's meant to look as is.  The subsequent entry illustrates what you've done.

PR #1482, opened 2024-04-10
docs: fixed typos, spaces, and punctuation

grjan7:
Fixes typos, spaces and punctuation.


crandmck:
Thanks for the PR @grjan7 

Based on what I saw so far, this all looks OK, but I stopped at the edits to the `_includes/readmes`. These files live in the corresponding repo and are simply copied here using [`get-readmes.sh`](https://github.com/expressjs/expressjs.com/blob/gh-pages/get-readmes.sh).
So, e.g., https://expressjs.com/en/resources/middleware/body-parser.html comes from https://github.com/expressjs/body-parser/blob/master/README.md.

So the files should be edited in each individual repo, not here.  There is a notice at the top of each such page on the site:
> Note: This page was generated from the xyz README.

We really should explain this somewhere, and it would be way better if it were done in a GitHub workflow instead of manually running a script.  

IAC, please remove these files from the PR, and then I'll continue reviewing.
Also, you should know that we plan to remove some pages that are linked from the "Resources" and "Advanced" menus. cf. https://github.com/expressjs/expressjs.com/wiki/Discuss:-Remove-pages-from-docs.
You might want to remove these files from the PR, too.  If we land that PR (which doesn't yet exist, but soon will) before you re-submit this one, you'll need to remove them from the PR.

crandmck (changes requested):
Remove files in `_includes/readmes` per comment.

grjan7:
Ok, @crandmck.

grjan7:
@crandmck, I have removed those files from PR. You can review them now.

en/starter/installing.md
```diff
@@ -44,7 +44,7 @@ $ npm install express --no-save
 ```
 
 <div class="doc-box doc-info" markdown="1">
-By default with version npm 5.0+ npm install adds the module to the `dependencies` list in the `package.json` file; with earlier versions of npm, you must specify the `--save` option explicitly. Then, afterwards, running `npm install` in the app directory will automatically install modules in the dependencies list.
+By default with version npm 5.0+, npm install adds the module to the `dependencies` list in the `package.json` file; with earlier versions of npm, you must specify the `--save` option explicitly. Then, afterwards, running `npm install` in the app directory will automatically install modules in the dependencies list.
```
crandmck: s/be `npm install` in first use here.

en/advanced/healthcheck-graceful-shutdown.md
```diff
@@ -120,9 +120,9 @@ app.listen(3000, () => {
 
 [http-terminator](https://github.com/gajus/http-terminator) implements logic for gracefully terminating an express.js server.
```
crandmck: terminating an Express server.

en/guide/overriding-express-api.md
```diff
@@ -61,7 +61,7 @@ Object.defineProperty(app.request, 'ip', {
 
 ## Prototype
 
-In order to provide the Express.js API, the request/response objects passed to Express.js (via `app(req, res)`, for example) need to inherit from the same prototype chain. By default this is `http.IncomingRequest.prototype` for the request and `http.ServerResponse.prototype` for the response.
+In order to provide the Express.js API, the request/response objects passed to Express.js (via `app(req, res)`, for example) need to inherit from the same prototype chain. By default, this is `http.IncomingRequest.prototype` for the request and `http.ServerResponse.prototype` for the response.
```
crandmck: Let's use "Express" everywhere rather than inconsistently using "Express", "Express.js", per the glossary entry.

crandmck (changes requested):

en/starter/installing.md
grjan7: Ok. I'll update it.

en/advanced/healthcheck-graceful-shutdown.md
grjan7: Ok.

en/guide/overriding-express-api.md
grjan7: It makes sense.

grjan7: However, I just left them as it was, in external references like blog/article titles.

crandmck: That's OK.

crandmck (approved):

crandmck:
Thanks again @grjan7. 

Corrections to spelling, punctuation, and formatting are valuable.  

Many of the edits in this PR simply removed extra spaces.  I don't feel that adds a lot of value.  It doesn't do any harm either, so I left them in this PR, but just for future reference.

grjan7:
Thanks for your input, @crandmck.

PR #1441, opened 2023-09-24
Update res-render.md

zareefhasan:
Using template engine link was redirecting to Turkish page buy default. Changing it to English version page.

PR #1486, opened 2024-04-12
Fix link to point to the English docs

Miguelrom:
A /en was added to the link for the Performance best practices section in English.

Without it, the link redirects to the docs in some other language.


jonchurch:
Hmm we had another PR like this recently as well, lemme see what we did. 

Do you mind experimenting with `{{page.lang}}` like in here, we may want to update everything to use this. https://github.com/expressjs/expressjs.com/pull/1441/files

Miguelrom:
Ok. I made the change suggested.

crandmck (approved):

crandmck:
Thanks @Miguelrom ... Sorry for the delay.

PR #1484, opened 2024-04-11
Remove ancillary pages and corresponding menu nav items

crandmck:
Per discussion in TC and:
- https://github.com/expressjs/discussions/issues/211
- https://github.com/expressjs/discussions/issues/105

Details are laid out in https://github.com/expressjs/expressjs.com/wiki/Discuss:-Remove-pages-from-docs.

Not previously discussed: The home page has a "Frameworks" tile with a link to one of the pages being removed.
- For the English page, I replaced this with   
```
      <h3>Middleware</h3> 
      Express is a routing framework with minimal functionality of its own.  A lot of
      functionality is contained in Express <a href="{{ page.lang }}/resources/middleware.html">middleware</a> modules.
```
- For non-en pages, I removed the text with no replacement.

PR #1489, opened 2024-04-14
asset optimizations

MehfoozurRehman:
so assets consume less bandwidth.


crandmck:
Many of these assets are no longer used due to landing https://github.com/expressjs/expressjs.com/pull/1484.  But some (e.g. the Express logo) are still used.  

Using less bandwidth is good, so I'm inclined to land it, but can anyone speak to the specific details of the changes in this PR?

NOTE:  We also need a PR to remove all the company logos that used to be on the "Companies using Express" page removed in the above-mentioned PR.  Any takers?

chrisdel101:
What do you need removed? I can do it. Do I need to wait until this PR  #1489 is merged before doing it?

crandmck:
Some of the files modified in this PR are no longer used.  For example, I believe that the files in `images/companies` are no longer used anywhere since #1484 landed.

So modifying them doesn't serve any purpose.  How about removing them in this PR, then we can land it?  
Ideally, we would have some way of determining ALL the images that are no longer used, and just remove them, but that might be asking too much :-)

MehfoozurRehman:
ok i am on it.

chrisdel101:
Sorry, thought you were asking for a diff PR for a separate issue. If this is the case I can do that. But don't want to interfere with this current PR.

PR #1501, opened 2024-04-28
add preview mode for pull requests

### LINKED ISSUE #1491: Expand CI functionality to preview changes from PR
Adding this functionality will make it easier to test and review changes submitted via PR directly in the browser.

lucasraziel:
Added a PR Preview to CI using  rossjrw/pr-preview-action@v1

closes https://github.com/expressjs/expressjs.com/issues/1491

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -1,0 +1,21 @@
+name: Deploy PR previews
+concurrency: preview-${{ github.ref }}
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - closed
+jobs:
+  deploy-preview:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: rossjrw/pr-preview-action@v1
+        with:
+          source-dir: .
+          preview-branch: gh-pages
+          umbrella-dir: pr-preview
+          action: auto


UlisesGascon:
Thanks for the proposal @lucasraziel!

I think we will need to have some capability of running previews from PRs generated in forks, but currently they are not supported:

> This Action does not currently support deploying previews for PRs from forks.
> [See more details here](https://github.com/rossjrw/pr-preview-action/pull/6)

I don't have a lot of experience with Jekyll previews, but maybe something like gostatic can solve the preview problem: [gostatic Jekyll GitHub Action](https://www.gostaticapp.com/jekyll-github-action). Since we will deploy the changes directly in a different environment (gostatic), what do you think about this approach?

I just found gostatic on a google search, so not sure if the new provider is the best option, but the free tier seems to cover out needs.

lucasraziel:
I will check about It on next weekend and/or will look for other alternatives.

jonchurch (changes requested):
accepting this as it is would be destructive, so we cannot

our gh-pages branch is our main/production branch in this repo. this action commits to gh-pages to deploy previews under a preview dir namespace

the action does have the ability to commit the previews to a separate repo, which could work but is extra maintenance burden

jonchurch:
Im planning to add netlify deploy previews since we can get those for free via their open source license. We can't leverage this repo's gh-pages for previews without committing to our production branch. This is going to have to be an integration, requiring admin perms, so I'll set it up.

PR #1511, opened 2024-05-03
add netlify logo as per their OSS license requirements

jonchurch:
> and also test the deploy previews

### DIFF
--- a/_includes/footer/footer-en.html
+++ b/_includes/footer/footer-en.html
@@ -7,7 +7,9 @@
   <br>
   Community translation available for: <a href="/sk/">Slovak</a>, <a href="/uk/">Ukrainian</a>, <a href="/uz/">Uzbek</a>, <a href="/tr/">Turkish</a> and <a href="/th/">Thai</a>.
   </section>
-  <div id="footer-content">
+    <div id="footer-content">
+    <a href="https://www.netlify.com"> <img src="https://www.netlify.com/v3/img/components/netlify-color-accent.svg" alt="Preview Deploys by Netlify" 
+      style="width: 50px;" /> </a>
       <div id="github">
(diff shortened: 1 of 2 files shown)


jonchurch:
Cool, previews work! the logo isn't aligned so this shouldn't land yet, someone else can run with it if they want

crandmck:
The preview capability is great!  Now I don't have to clone every fork/branch to see what it looks like--awesome!

But I'd like to suggest:
- The logo should be a bit smaller so it's not quite so prominent.  Do they require any particular size?
- It should flow better with the existing footer.  

Here's my suggestion:
Screenshot 2024-05-02 at 10 19 14 PM

What do you think?

This is done in `_includes/footer-en.html` like this:

```
  <div id="footer-content">
    <a href="https://www.netlify.com"> <img src="https://www.netlify.com/v3/img/components/netlify-color-accent.svg" alt="Preview Deploys by Netlify" 
      style="width: 50px;" /> </a>
```

jonchurch:
LGTM

And just to be clear (for posterity), we are still hosting the site and deploying via github pages

Only previews are netlify, so "powered by netlify" isnt super true. 

We also have the option to provide a link to netlify if we prefer not to use their badge

crandmck:
OK, I made that change in your fork/branch @jonchurch .  It was also a good trial of the deploy preview functionality.  So nice.

Notes:
- This is only for the English footer.  Someone can add it to the other langs if they wish.  Not required IMO.
- Logo appears in the footer for ALL en pages, not just the home page.  This is a side-effect of putting it in the footer.  If anyone objects, we can look at alternatives.

jonchurch:
good stuff

wrapped it in a div to give it a bit more spacing on mobile like the other footer items:

Screenshot 2024-05-04 at 4 05 52 PM

PR #1515, opened 2024-05-09
Docs: Clarify inheritance behaviour of router params callbacks

mao-sz:
Current documentation specifies router param callbacks are not inherited by mounted apps or routers, but does not also specify they do not inherit from parent routers either.

This lack of explicitness led to false expectations in [Express #5586](https://github.com/expressjs/express/issues/5586).

### DIFF
--- a/_includes/api/en/4x/app-param.md
+++ b/_includes/api/en/4x/app-param.md
@@ -22,7 +22,7 @@ app.param('user', function (req, res, next, id) {
 })
 ```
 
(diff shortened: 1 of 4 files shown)


crandmck (approved):

crandmck:
The changes look fine to me, but I'd like confirmation from @wesleytodd since he was involved in the original discussion in https://github.com/expressjs/express/issues/5586.

wesleytodd (approved):
LGTM! Thanks @MaoShizhong!

PR #1516, opened 2024-05-13
Add express-joi-validations to community middlewares

mattiamalonni:


crandmck:
Thanks @mattiamalonni ...

@expressjs/docs-wg 
In a recent PR we removed from the website a bunch of pages with ancillary information and links to repos/sites outside the expressjs org.   We did this specifically to reduce the ongoing maintenance burden--TC and collaborators don't have time to vet every process manager, blog, book, etc.  The main purpose of the website is to document Express itself and middleware in the expressjs org (i.e. middleware that the team maintains) and we need to keep our focus on that.

However, we kept this section: https://expressjs.com/en/resources/middleware.html#additional-middleware-modules. I guess it was an oversight on my part.  Obviously, middleware is a very important resource and it may be useful to keep it.  However, if we keep it I think we need to have some basic guidelines for modules listed here.  

I would also note that we should be careful because in theory any middleware might come with some security risks, and per recent warnings we should not take this lightly.  Even though there is a disclaimer, listing it on the website might lend some credibility to any module, so I think we should do some kind of vetting.

IMO we have two options:
- Remove this section altogether.
- Keep this section, and define a simple set of criteria for inclusion (e.g. at least version 1.0, at least x number of stars in GitHub, at least x number of downloads in npm). 

I would note that this module proposed to be added in this PR (https://github.com/mattiamalonni/express-joi-validations) is very new (first commit was 2 weeks ago), is on version 0.1.0, has 0 stars, and 133 weekly downloads on npm.

mattiamalonni:
Thanks @crandmck,
I totally understand your point and agree with you. I made this PR to give visibility to the tool but I am aware that it could be rejected for security reasons and the lack of popularity of the middleware.

inigomarquinez:
I also agree with you @crandmck , and from the 2 options you mention I think we should take a decision to move on.

crandmck:
Thanks for your understanding @mattiamalonni ...

@inigomarquinez To make sure I understand, when you say 
> take a decision to move on.

Do you mean to remove the community middleware section altogether?

Issue #1508, opened 2024-05-02
Dark Mode flash


jonchurch:
There's a flash on page load when user has dark mode set

We can inline the css and js to set up the page before it's rendered to the browser

Slowed down w/ connection throttling:
![flash](https://github.com/expressjs/expressjs.com/assets/12915163/1c287d9f-ca7d-416c-abee-9cdeb5869bb2)
related #1490

crandmck:
@chrisdel101

chrisdel101:
Thanks for the feedback. I think there might be an influx of bug reports on this one, which I thought would be the case.

I did notice this but didn't realize it would be such an issue. I can make this fix. Do you have any tips on the best way to fix it?

jonchurch:
No worries @chrisdel101, you made a good change, and issues will come in identifying gaps. Power of the crowd! If you want to hack on this please do, others may want to fix some of the rough edges too. If we didn't like the change we would just revert it.

chrisdel101:
Thank @jonchurch! I will hack.

chrisdel101:
There appears to be some issue specifically with the GH pages build of this site. #1524 should have fixed it but did not. If there is someway to re-build the production branch maybe this would fix it? 

I've setup a fork version of the express site also using GH pages with my local account and the flash does _not_ appear to occur there: https://chrisdel101.github.io/my-expressjs.com/

crandmck:
Can you make sure it's not some kind of caching issue?  Sometimes this happens --- and a hard reload will fix it.  Otherwise, I'm puzzled.

crandmck:
Wow, that's weird.  I honestly don't know how that could happen.

The only thing I can think of is turning off GHP and then turning it back on.  That would bring the site down momentarily... Should I try that?

chrisdel101:
Maybe it will fix on the next PR that gets pushed as this will reload it then? 
I noticed #1526 is fixed but I don't see how it got fixed. Might be the same kind of thing (unless a fix was applied to #1526 that I just can't see). You can see the nav flash still happens on the link above.

chrisdel101:
Just revisiting this since I hate this bug! It is still happening. We determined that it's something to do w/ the GH-pages build of this site. I'm unable to reproduce it when I deploy this same site to my own github (as seen above) but I'm not using any of the scripts, actions, etc, so it might be something there? I really have no idea...

[...]

PR #1524, opened 2024-05-30
Dark Mode Flash Fix

chrisdel101:
This fixes issue #1508. Adjust setting of the dark mode class and moved the order of the head to load to run the JS first and add the `dark-mode class`, then load dark mode style sheet first, if required. The theme class is now on the `html` tag to run this all before DOM load. 

If this is not the final solution, it is significantly better than the current situation, which is extremely jarring. 

I'm noticing a flicker now when loading the page, in the nav bar. It's occurring on my local build with the latest changes, but is not occurring on the production site online. So I'm going to submit this PR so I can see the preview to check for the flicker. 

EDIT: After checking, the flicker is there in preview, but is coming from the main branch for me. Doesn't seem to be caused by this PR.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -5,8 +5,9 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta name="og:image" content="https://expressjs.com/images/express-facebook-share.png">
     <link rel="icon" type="image/png" href="/images/favicon.png" />
+    <script src="/js/theme.js"></script>
+    <link rel="stylesheet" href="/css/dark-theme.css?_={{ site.time | date: '%s' }}">
     <link rel="stylesheet" href="/css/style.css?_={{ site.time | date: '%s' }}">
-    <link rel="stylesheet" href="/css/theme.css?_={{ site.time | date: '%s' }}">
     <link rel="stylesheet" href="/css/dropit.css">
     <link rel="stylesheet" href="/css/prism.css">
     <link rel="stylesheet" href="/css/font-awesome.min.css">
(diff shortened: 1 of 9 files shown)


crandmck (approved):

crandmck:
Also added @jonchurch since he opened the original issue.

chrisdel101:
I had made it a draft PR to check out that preview as I noted before, so I just made it a normal PR. 
~Please do check for the flicker in the nav bar while reviewing this :)~
~(It's possible the issue this is fixing is actually covering up the nav flicker means it's not caused directly by this PR)~

chrisdel101:
RE: previous comments about nav bar flicker. I think this is a separate issue #1526. So disregard that.

chrisdel101:
This has not fixed the problem :( Is this change pushed to production already?

If I look at the preview, the flash is not there, that's why I thought it was fixed. But the flash is there in prod.

chrisdel101:
I do see the code changes are in prod. So I guess I need to take another stab at this. I'm not sure why it's not working as expected. 

I pulled the updated prod changes to my machine and the flash is NOT happening there. So I'm not sure how to debug this since it's only happening on the expressjs.com site, even thought I can see the changes in the markup there.

chrisdel101:
Maybe it's a GH pages issue? Whereas the preview is Netlify and not happening there? 
I can maybe try hosting my own GH pages instance of the site to solve this. I can't think of any other way I'd be able to find out what the issue is here.

chrisdel101:
I tried to push my own instance of the site to GH pages and it's not working. It's not inputting the `expressjs.com` part after my site name. This is a HUGE issue I run into every time I use pages, but I assumed that I'd be able to push this repo _as is_ and it would work just using the settings in the repo.

How do you guys get around this when deploying to pages? 

It's going to take alot of tinkering to get this to run on my own pages if I have to do it all manually

chrisdel101:
FYI: Hacked together a working version.

PR #1527, opened 2024-06-11
dark theme updated

iryanraushan:
update the dark theme color


chrisdel101:
Hi @iryanraushan. Our dark mode is far from perfect! But I'm not seeing a real need to "update the dark theme color." 
Using GitHub dark mode as a model, the range between colors should be very small, blending together almost like it's one color. Obviously there's no single "right way," but that's what we're going for.
 
I'm going to close this (@crandmck can re-open if you disagree :) )

PR #1528, opened 2024-06-20
feat - adjust in layout and theme - v2

carlosstenzel:
Adjust in layout and theme

Before

https://github.com/expressjs/expressjs.com/assets/3890516/e123d4a2-294e-4998-9099-97c09bc592f4

After

https://github.com/expressjs/expressjs.com/assets/3890516/c600615f-05f7-46bb-a7fc-3a5d57d038eb

### DIFF
--- a/_includes/header/header-en.html
+++ b/_includes/header/header-en.html
@@ -1,11 +1,12 @@
 <header>
+  <div class="container">
   <div id="mobile-menu">
       <div id="nav-button" class="fa fa-bars fa-2x button"></div>
   </div>
   <section id="logo"><a href="/" class="express">Express</a>
   </section>
   <div id="navbar">
-      <input id="q" placeholder="🔎 search">
+      
       <ul id="navmenu">
           <li><a href="/" id="home-menu"{% if page.menu == 'home' %} class="active"{% endif %}>Home</a></li>
           <li>
@@ -153,8 +154,11 @@
           </li>
         -->
       </ul>
-      <div id="theme-icon-container"  class="theme-toggle default-theme" title="toggle darkmode">
-        <i class="fa fa-moon-o fa-2x"></i>
+      <input id="q" class="input-search" placeholder="Search ...">
+      <div id="theme-icon-container" class="theme-toggle default-theme" title="toggle darkmode">
+        <i class="fa fa-moon-o fa-lg hidden-dark"></i>
+        <i class="fa fa-sun-o fa-lg hidden-light"></i>
       </div> 
   </div>
+</div>
 </header>
--- a/css/style.css
+++ b/css/style.css
@@ -25,6 +25,31 @@ h1 {
   line-height: 36px;
 }
 
+.container {
+  width: 100%;
+  max-width: 1100px;
+  margin: auto;
+}
+
+.container-api {
+  width: 100%;
+  max-width: 1100px;
+  margin: auto;
+  padding: 0 5%;
+  overflow: auto;
+}
+
+.input-search {
+  border-radius: 8px;
+  border: 1px solid #ddd;
+  padding-left: 8px !important;
+}
+
+.input-search:focus-visible,
+.input-search:focus {
+  border-color: #333 !important;
+}
+
 #intro h3, #api-doc h1 {
   font-weight: normal;
 }
@@ -137,9 +162,9 @@ strong, th {
 }
 
 .content {
-  margin: 153px 3% 7%;
   max-width: 1090px;
-  padding-left: 225px;
+  margin: auto;
+  margin-top: 120px;
 }
 
 span.block-section {
@@ -153,6 +178,11 @@ li code {
 #intro h3 {
   font-size: 25px;
   margin-bottom: 10px;
+}
+
+#api-doc {
+  width: 80%;
+  float: right;
 }
 
 #api-doc section {
@@ -234,13 +264,13 @@ a {
   display: block;
   font: 25px "Helvetica Neue", "Open Sans", sans-serif;
   font-weight: 100;
-  color: #444;
+  color: #0d0d0d;
 }
 
 #description .express {
   display: block;
-  font: 4.5em "Helvetica Neue", "Open Sans", sans-serif;
-  font-weight: 100;
+  font: 5em "Helvetica Neue", "Open Sans", sans-serif;
+  font-weight: 200;
   margin-bottom: .25em;
 }
 
@@ -621,6 +651,8 @@ footer {
   margin: 60px 10px 30px 10px;
   padding-left: 5%;
   max-width: 1090px;
+  margin: auto;
+  margin-top: 20px;
 }
 
 #footer-content {
@@ -723,6 +755,7 @@ footer {
   border: 1px solid #b2b2b2;
   margin: 0 0 0 1px;
   box-shadow: 0px 1px 3px rgba(0,0,0,0.15);
+  border-radius: 5px;
 }
 
 .menu ul.dropit-submenu a {
@@ -740,12 +773,12 @@ footer {
 /* secondary menu */
 
 #menu {
-  position: fixed;
+  width: 20%;
+  float: left;
   margin: 0;
   padding: 0 10px 0 0;
   top: 153px;
   left: 30px;
-  height: 500px;
   text-align: left;
   font-size: 13px;
   overflow-y: auto;
--- a/_includes/footer/footer-en.html
+++ b/_includes/footer/footer-en.html
@@ -1,4 +1,4 @@
-<a id="top" href="#"><img src="/images/arrow.png"></a>
+<a id="top" href="#" class="icon-top-footer"><img src="/images/arrow.png" /></a>
 
 <footer>
   <section id="doc-langs">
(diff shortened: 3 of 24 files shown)


_includes/header/header-en.html
```diff
@@ -153,8 +154,11 @@
           </li>
         -->
       </ul>
-      <div id="theme-icon-container"  class="theme-toggle default-theme" title="toggle darkmode">
-        <i class="fa fa-moon-o fa-2x"></i>
+      <input id="q" class="input-search" placeholder="Search ...">
+      <div id="theme-icon-container" class="theme-toggle default-theme" title="toggle darkmode">
+        <i class="fa fa-moon-o fa-lg hidden-dark"></i>
+        <i class="fa fa-sun-o fa-lg hidden-light"></i>
```
chrisdel101, line 160: Personally I don't like this icon b/c it looks like the "settings" gear icon to me. It's not clear that it means "light mode" immediately.

_includes/header/header-en.html
```diff
@@ -153,8 +154,11 @@
           </li>
         -->
       </ul>
-      <div id="theme-icon-container"  class="theme-toggle default-theme" title="toggle darkmode">
-        <i class="fa fa-moon-o fa-2x"></i>
+      <input id="q" class="input-search" placeholder="Search ...">
```
chrisdel101, line 157: I think the search bar should remain on the left side of the nav bar since moving it is a drastic-ish change. Also there's some overlap issues. Also happens on the Express Logo icon side too
Screenshot 2024-06-23 at 11 55 55 AM

css/dark-theme.css
```diff
@@ -13,27 +13,60 @@
...
+  display: block;
+}
+
+html.dark-mode #theme-icon-container .hidden-dark {
+  display: none;
 }
+
 html.dark-mode > body {
-  background: var(--main_dark_bg);
+  background: #000;
```
chrisdel101, line 42: - Why undo the variable with a hard-coded value? If you want a one-off change I suggest making it variable anyway. 

- And why change the color here anyway? I don't see the need for this.

css/dark-theme.css
```diff
@@ -13,27 +13,60 @@
...
+
+html.dark-mode #logo > a,
+html.dark-mode #navbar li a:hover {
   color: var(--dark_text);
 }
 /* navbar links/drop down menu links */
 html.dark-mode #navbar ul#navmenu li a,
 html.dark-mode #navbar ul#navmenu li.dropit-trigger a {
-  color: var(--dark_text);
+  color: #b6b6b6;
```
chrisdel101, line 69: - Again, I don't see any reason to hardcode this. You should make it a variable. 
- I personally like the white color that is is now, but I could be sold on this color if others like it more.

css/style.css
```diff
@@ -25,6 +25,31 @@ h1 {
...
 }
 
+.container {
+  width: 100%;
+  max-width: 1100px;
+  margin: auto;
+}
+
+.container-api {
+  width: 100%;
```
chrisdel101, line 35: - This is too wide. Since the page is largely text it's awkward to read from edge to edge.
- The width of the styles currently works better.
- Standard I've always used is like a piece of paper, or a tablet size

css/style.css
```diff
@@ -25,6 +25,31 @@ h1 {
...
+  max-width: 1100px;
+  margin: auto;
+  padding: 0 5%;
+  overflow: auto;
+}
+
+.input-search {
+  border-radius: 8px;
+  border: 1px solid #ddd;
+  padding-left: 8px !important;
```
chrisdel101, line 45: - Never use important. I've been skunked a few times on this project by other ppl doing this.

css/style.css
```diff
@@ -25,6 +25,31 @@ h1 {
...
+}
+
+.input-search:focus-visible,
+.input-search:focus {
+  border-color: #333 !important;
```
chrisdel101, line 50: - Again, find a different way to do this. No important.

css/dark-theme.css
```diff
@@ -13,27 +13,60 @@
...
+
+html.dark-mode #theme-icon-container {
   color: var(--dark_text);
+  background-color: var(--second_dark_bg);
 }
+
 html.dark-mode header {
   background-color: var(--second_dark_bg);
   color: var(--dark_text);
+  border-bottom: 1px solid #626262;
```
chrisdel101, line 54: - just personal preference but I like the white border of the current styles more. I just think it pops more.

css/dark-theme.css
```diff
@@ -44,11 +77,17 @@ html.dark-mode #navbar ul#navmenu li a.current:hover {
...
   background: var(--dark_bg);
+  border: 1px solid var(--dark_text-gray);
 }
 /* drop down menu links  */
 html.dark-mode #navbar .menu ul.dropit-submenu a:hover {
   background: var(--dark_hover);
 }
+
+html.dark-mode #navbar a.active {
+  color: var(--dark_text) !important;
```
chrisdel101, line 88: - Don't use important. Please find another way to make this work. You can make it more specific like `html.dark-mode #navbar ul#navmenu li a, html.dark-mode #navbar ul#navmenu li.dropit-trigger a` type of thing

css/style.css
```diff
@@ -234,13 +264,13 @@ a {
...
-  color: #444;
+  color: #0d0d0d;
 }
 
 #description .express {
   display: block;
-  font: 4.5em "Helvetica Neue", "Open Sans", sans-serif;
-  font-weight: 100;
+  font: 5em "Helvetica Neue", "Open Sans", sans-serif;
+  font-weight: 200;
```
chrisdel101, line 273: - This is too thick (I think). The current way it matches the Express icon more closely as well. 
- I like current way better.

chrisdel101:
Apologies for if anything appears overly critical, but there are alot of changes here so it just more room for critique :)

Overall there are some good ideas here, but maybe too many? I'd personally rather have had you just do a PR that was say "Reformat Nav Bar," since the nav is problematic,  rather than trying to tackle the entire site in one PR (and which I'm honestly not seeing a huge need for). If you are open to that, maybe we could just focus on the nav parts to start with? And try the rest in another PR?

Anyway, I did some reviewing of the whole thing, but due to the size I could not check every single thing. I'd strongly suggest we do another pass over this again later.

I made some comments. I put "personally" or "I think" in many of the comments since I don't see the need for them, but @crandmck might have a different outlook. 

The others probably do require the changes. Get rid of all the `important`s for sure. 

Random other things I'd suggest
- I miss the border around the nav bar dropdowns. You've removed this.
- On mobile view there is some strange x-scroll occurring on the API pages. This is not happening in production. Could just be my setup, but please check that out. 
- On a random page the text content stretches all the way to the edge of the page. We can't have this. I made comment about text block width conventions, or you can fall back to the current settings since they work okay.

Screenshot 2024-06-23 at 12 50 32 PM

carlosstenzel:
> Apologies for if anything appears overly critical, but there are alot of changes here so it just more room for critique :)
> 
> Overall there are some good ideas here, but maybe too many? I'd personally rather have had you just do a PR that was say "Reformat Nav Bar," since the nav is problematic, rather than trying to tackle the entire site in one PR (and which I'm honestly not seeing a huge need for). If you are open to that, maybe we could just focus on the nav parts to start with? And try the rest in another PR?
> 
> Anyway, I did some reviewing of the whole thing, but due to the size I could not check every single thing. I'd strongly suggest we do another pass over this again later.
> 
> I made some comments. I put "personally" or "I think" in many of the comments since I don't see the need for them, but @crandmck might have a different outlook.
> 
> The others probably do require the changes. Get rid of all the `important`s for sure.
> 
> Random other things I'd suggest
> 
>     * I miss the border around the nav bar dropdowns. You've removed this.
> 
>     * On mobile view there is some strange x-scroll occurring on the API pages. This is not happening in production. Could just be my setup, but please check that out.
> 
>     * On a random page the text content stretches all the way to the edge of the page. We can't have this. I made comment about text block width conventions, or you can fall back to the current settings since they work okay.
> 
> 
> Screenshot 2024-06-23 at 12 50 32 PM

Thank you for the review, we can divide and treat each point separately.

I couldn't see your review comments, do I need permission?

carlosstenzel:
@chrisdel101 
I moved in another PR, this point about change icon in toggle theme
https://github.com/expressjs/expressjs.com/pull/1532

chrisdel101 (changes requested):

chrisdel101:
I think I needed to submit the review. Hopefully you can see them now.

_includes/header/header-en.html
carlosstenzel: I didn't find any other icon that could replace it :( 

https://fontawesome.com/v4/icons/

carlosstenzel: I added another icon for the sun
https://github.com/expressjs/expressjs.com/pull/1532

carlosstenzel: with fa icons
Captura de Tela 2024-06-24 às 11 42 04

new icon
Captura de Tela 2024-06-24 às 11 45 51

PR #1519, opened 2024-05-18
Express Blog

chrisdel101:
So hoping you can live preview this before making it live. With that in mind, I've not added any photos. 

The feature includes the `/posts` page, the `/write-posts` page, and the `/example-post` page. The last two are very similar. If they are too similiar we can get rid of the latter.
You can visit the individual posts themselves by clicking on the title 
**Review item**: Would you rather have the entire div clickable? Or okay as is?

There are some changes made to the search bar to make it fit with added list items. It's only been tested on mac, so hopefully it works on windows.

The gutter menu is just another listing of the blog links. There is also a menu for tags, but since there are not enough tags, this isn't super useful right now. If it's okay to just leave this unused, great. Else, we can delete it from the code base.

Last, there are 3 dummy posts used to populate the posts page. After preview reviewing is complete, we will need to delete those as they are only for the purposes of review. Those are the ones entitled `sample-post`.

If there are no posts this is what displays
Screenshot 2024-05-18 at 3 07 40 PM

Let me know if you find any bugs :) 
PS - I  have no idea what those build errors mean.
PSS - I added some comments to the files so you know what they are doing (I seriously hope you did not get notifications for all of those! If you did, sorry!) Also, I assigned you guys to this PR. I hope that's okay :)

PR #1529, opened 2024-06-20
Chore/pr only preview 1519

carlosstenzel:
Only preview to PR https://github.com/expressjs/expressjs.com/pull/1519


crandmck:
Thanks @carlosstenzel for doing this... 
It looks like the merge conflicts were resolved in #1519 so I don't think this is needed any longer.  If that's the case, you can go ahead and close the PR.

PR #1534, opened 2024-06-25
Update best-practice-security with TLS Cipher Suites.md

stellastrareviews:


chrisdel101:
Hi @stellastrareviews. As this is an unsolicited PR with no text body description whatsoever I'm inclined to close this. 
I'll ask for @crandmck  to take a look first.

crandmck:
Thanks @chrisdel101 ... 
I'm not a security expert, so I have no idea if this is good advice.  It's important to provide good security guidance, so if it IS, then we should include it.  

Perhaps @jonchurch can advise?   Otherwise, then, yes, we can just close this.

ctcpip:
Suggest closing without merging.  Unsolicited PR recommending a proprietary tool which is redundant for the recommendations already provided.

PR #1541, opened 2024-07-14
Fix links

anitahammer:


crandmck:
I don't think this change is required.  The links in https://expressjs.com/en/4x/api.html#req and https://expressjs.com/en/5x/api.html#req (in the admonition under "Properties") seem to work just fine.

PR #1543, opened 2024-07-19
Update database-integration.md

selimsezr:
add DuckDB integration

### DIFF
--- a/en/guide/database-integration.md
+++ b/en/guide/database-integration.md
@@ -22,6 +22,7 @@ Adding the capability to connect databases to Express apps is just a matter of l
 * [SQL Server](#sql-server)
 * [SQLite](#sqlite)
 * [Elasticsearch](#elasticsearch)
+* [DuckDB](#duckdb)
 
 <div class="doc-box doc-notice" markdown="1">
 These database drivers are among many that are available. For other options,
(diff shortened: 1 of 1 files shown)


en/guide/database-integration.md
```diff
@@ -494,3 +495,32 @@ client.search({
   console.trace(error.message)
 })
 ```
+
+## DuckDB
+**Module**: [duckdb]([https://github.com/elastic/elasticsearch-js](https://github.com/duckdb/duckdb))
```
crandmck, line 500: This is a malformed md link.  There are 2 URLs here--pick one.
Why would Elastic be relevant?

crandmck (changes requested):

crandmck:
@jonchurch I don't see any reason not to add DuckDB.  WDYT?

Separately--and in the spirit of reducing/eliminating links or endorsements of 3rd party packages or products--I wonder if we should either move this page to the "Advanced" menu or just remove it entirely?  Clearly, it has useful info, but...?

selimsezr:
It might be good to move it to the “Advanced” page. I tried for about 4 days to make the DuckDB connection with Nextjs and it didn't work. I did it much faster with Express. So I think it's important to write 3rd party applications. When there is an update or error, a PR/Issue can be created immediately.

I couldn't understand where the error was either, I only made changes in the .md file.

crandmck:
@selimsezr If you would address my comment about the link, we can consider merging this.

crandmck:
This hasn't been update for over a month, so closing.

PR #1546, opened 2024-07-23
Clarified the explanation of Router

tomasz13nocon:
Discussed in expressjs/express#5753

tldr: the current wording of the docs on Router made it seem like it isolates its middleware from the rest of the app, which is not the case.


crandmck (approved):

crandmck:
Thanks @tomasz13nocon !

PR #1550, opened 2024-07-28
feat: add style of scroll

bjohansebas:
styles are added to the scroll, as is done on YouTube or other websites at chrome

<div style="display: flex;">

</div>

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -191,6 +191,28 @@ li code {
   color: #666;
 }
 
+/* scroll */
+
+*::-webkit-scrollbar {
+  background-color: hsla(0, 0%, 100%, .145);
+  width: 16px;
+}
+
+*::-webkit-scrollbar-thumb {
+  border-radius: 8px;
+  border: 4px solid transparent;
+  background-clip: content-box;
+  background-color: hsl(0, 0%, 67%);
+}
+
+*::-webkit-scrollbar-thumb:hover {
+  background-color: gray
+}
+
+*::-webkit-scrollbar-thumb:active {
+  background-color: #616161
+} 
+
 /* links */
 
 a {


crandmck:
In the deploy preview, I can't actually see what this does ... ?

UlisesGascon (approved):
Thanks @bjohansebas! It is more elegant now :)

UlisesGascon:
> In the deploy preview, I can't actually see what this does ... ?



@crandmck (new left, old right) comparing in chrome


Screenshot from 2024-07-31 20-46-42

chrisdel101:
I concur with @crandmck. I don't see any difference. I'm comparing the prod url to the preview, I have them both open, and they appear to be the same? (MacOS: FF & Chrome)

Is there any particular page I should be looking at? This is just the main page scroll right?

UlisesGascon:
> Is there any particular page I should be looking at? This is just the main page scroll right?

I tested on https://deploy-preview-1550--expressjscom-preview.netlify.app/en/guide/routing vs https://expressjs.com/en/guide/routing. I am using Chromium `Version 126.0.6478.182 (Official Build) snap (64-bit)
` in Ubuntu

bjohansebas:
@chrisdel101 @crandmck In Firefox, it doesn’t work because they don’t use WebKit, and also because they handle scrolling a bit differently. In Chrome and its derivatives, it should work fine, at least on Linux and Windows where I can test it. It’s possible that on macOS, Chromium might not allow changing the scroll design due to the operating system's design constraints.

chrisdel101:
I can't test the UI changes then but consensus seems this change is positive. The code changes look fine to me.

crandmck:
Thanks @bjohansebas !

Issue #1206, opened 2020-09-16
Docs & design overhaul


sk1nzz:
Hi there. It's 2020 and Express remains the most popular Node framework. However, we still have an obsolete docs site that doesn't meet today's design trends.

I think it's time to:

- rebuild the docs site with modern static site frameworks like Gatsby;
- update the Express logo and implement a design system for the docs.

Your opinions?

rodion-arr:
Hi, I assume core team is focused now on pushing next major `express` version. If you have something to propose opening a PR is more than welcome!

crandmck:
PRs always welcome, of course, but this proposal would be a significant project.  Does Gatsby have a GitHub integration?  That's one of the main advantages of Jekyll.

hunvreus:
I'd be more than happy to contribute to this, but not sure pushing a massive PR would be helpful.

I'm assuming there are several people involved in maintaining the website, I think they should be involved in the process at the very least with regards to UX and sitemap.

It would likely take several steps to get to an updated version of 

- **Stack**: I don't mind sticking to Jekyll. We could certainly move to another SSG (e.g. Eleventy) but I'm not sure that would address the key issue (UI and UX), although it could help if the current build process is too slow. I do think it'd be helpful to adopt a modern CSS framework though (e.g. Tailwind).
- **Sitemap**: I think there should be a concerted effort to review the current pages/sections and figure out what would make sense for the projects (AND the site contributors). I'm not sure there are analytics on the site, that could be helpful figuring out what folks are actually looking for.
- **Design**: expectations aren't what they used to be when this website was designed, even for Open Source project. I think the current design is a bit damaging to all the hard work that the maintainers are putting in. This could however be the more contentious point for Express.js collaborators (it's always difficult to get a group to agree on UI choices). I would also say that a very light refresh of the wordmark wouldn't hurt, but I wouldn't want it to distract from the UI/UX work that needs to be done.

Happy to put in work there, but again I'm mostly concerned with how we should break this work down.

I'm fine with either:

- **Option 1**: breaking things down in a bunch of small incremental improvements,
- **Option 2**: make a big redesign push we can refine until we're comfortable with the changes.

@dougwilson I'd be curious to know who leads on the expressjs.com front (yourself?) and where contributors hang out (IRC seems dead and Gitter mostly looks like user support).

jonchurch:
There's a lot to be said about a docs and design overhaul.

But let's break it down slightly

### Docs

[...]

PR #1551, opened 2024-07-30
Change from Jekyll to 11ty

bjohansebas:
Currently, Jekyll is a good static site generator, but there are now better tools like 11ty that can greatly improve the development experience. This includes enhancements in compilation time, the implementation of tools like vitejs, and even the minification of pages.

Several of the changes do not require rewriting the documentation, as 11ty supports Liquid, facilitating the transition to newer technologies that enhance the documentation experience such as mdx.

<details><summary>Build time</summary>
<p>

</p>
</details> 

TODO: 

- [ ] Update write posts
- [ ] Update readme


en/blog/write-post.md
```diff
@@ -6,24 +6,23 @@ lang: en
...
+![Blogger](/images/blogger.jpg)
 
 If you have an idea for a blog post, follow these steps to propose it and potentially get it published!
 
 1. ### Propose your post
 Before taking the time to write a post, please confirm that we will be able to publish it. We're looking for topics specifically related to Express, and so we want to pre-approve all posts. For the moment, this means we aren't accepting any unsolicited posts. To propose a blog post, [open an issue](https://github.com/expressjs/expressjs.com/issues) entitled  `Blog post proposal: <your idea>`.
 
 
-1. ### Fork the repository 
+2. ### Fork the repository
```
ctcpip, line 17: these should all remain `1.`

crandmck:
I think it's reasonable to move from Jekyll to a more modern framework, however I think a change of this magnitude requires at least _some_ discussion.  While 11ty did come up in conversation, I don't think we've had any specific discussion and evaluation of other frameworks in an issue... correct me if I'm wrong.

Before landing this, I'd like to at least discuss the framework to which we'll move and ideally get consensus on a decision.
We can use any of these issues:
- https://github.com/expressjs/expressjs.com/issues/1206
- https://github.com/expressjs/discussions/issues/211

UlisesGascon:
I agree with @crandmck! We might want to choose the right tooling first. Something that might be a good fit for us is https://docusaurus.io/. It already has many features we need out of the box, like MDX, i18n, blog, responsive design, dark theme, API Docs versioning, etc. It's widely used by JS libraries like Jest (https://jestjs.io/).

I think one of the keys for us is to try to use a framework that allows us to focus on minor changes and content rather than building a lot of components, pages, etc., from scratch.

bjohansebas:
@UlisesGascon @crandmck Of course, you are right; it is hasty to switch from one tool to another, and it requires discussions to ensure that nothing in the documentation breaks

Although I had also chosen this library to facilitate an easy transition that supports Liquid and doesn’t require much modification to the documentation, with the intention of later migrating to other tools like Starlight, VitePress, or Docusaurus, after having converted the documentation to MDX, which is used by these tools as a content generator.

PR #1552, opened 2024-08-01
Using variable to display the version in migration guide for 5

### LINKED ISSUE #1464: Update 5.0 version in Migration Guide
Docs in en and de (in https://github.com/expressjs/expressjs.com/pull/1371) were updated, but all other langs need update as well to change: 

```
$ npm install express@5.0.0-alpha.2 --save
```
to 
```
$ npm install "express@>=5.0.0-beta.1" --save
```

Perhaps we should consider a better way to address this in the long run that wouldn't require touching so many files.  The `_data/express.yml` file has a `current_version` version for the latest 4.x release version.  We could add `next_version` for the 5.x latest version.  Of course, when 5.x is released, `current_version` would become `5.0.0` or whatever, and `next_version` would be something like `6.0.0-alpha.1` or the like.

IamLizu:
As @crandmck suggested in [expressjs/expressjs.com#1464](https://github.com/expressjs/expressjs.com/issues/1464), we need a better way to address the version change in the long run.

I have kept a variable in `_data/express.yml` as suggested and referenced that variable in the hardcoded npm command in all available language pages.

There was a formatting issue at ko lang, basically removed unnecessary script declaration to fix that.

### Notes
- Closes https://github.com/expressjs/expressjs.com/issues/1464


IamLizu:
The check translation action throwing an error saying,

```js
RequestError [HttpError]: Resource not accessible by integration
```

I am guessing this has something to do with the action itself as nothing else than the required md files and `_data/express.yml` was changed in the PR.

cc: @UlisesGascon @crandmck

UlisesGascon (approved):
LGTM!

UlisesGascon:
Yes @IamLizu! Seems like there is an issue with GitHub actions regarding HTTP requests that is not related to this change, we can ignore that CI check.

crandmck (approved):

crandmck:
Thanks @IamLizu !

PR #1554, opened 2024-08-02
Remove applications page of russian lang and all the images it uses

bjohansebas:
This page no longer exists in the English documentation


chrisdel101:
So this is a good change. I see the applications file is not being used (Myspace! ha)
But, what does removing the screenshots have to do with removing the Russian page though? I'd like to see this PR handle _just_ the Russian page - so move those other changes to another PR (I didn't confirm if they are valid here)
@crandmck let us know if you disagree

chrisdel101:
Ahh I see. All those images are from that page. I'll need to verify those images still then.
@crandmck I assume we don't need this page? Just wanted to double check https://expressjs.com/ru/resources/applications.html

Weirdly this page is not even Russian. Not sure why it's sitting here.
image

crandmck:
Yes, that page is no longer used and can be removed.  All the other langs were removed in a previous PR, and this was overlooked, apparently.

FYI for community translations, pages that need translation are added with their en content initially, with the intent to replace with translated version later. Cf. https://github.com/expressjs/expressjs.com/blob/gh-pages/CONTRIBUTING.md.

I changed the title of the PR to reflect that it also removes the referenced image files, and I double-checked that those files are not used anywhere else.  Since the Netlify deploy preview works, it should be good to merge.

crandmck (approved):

PR #1555, opened 2024-08-02
Update readmes and contributing page

bjohansebas:
Run `./get-readmes.sh` and `./get-contributing.sh`

In the future, it might be possible to remove external README files with links to the repositories themselves to avoid having to run that script.

PR #1559, opened 2024-08-03
feat: add workflow to update external docs

bjohansebas:
To avoid having to update the documentation like in #1555 

see: https://github.com/peter-evans/create-pull-request/tree/v5?tab=readme-ov-file#action-inputs

### DIFF
--- a/.github/workflows/update-external-docs.yml
+++ b/.github/workflows/update-external-docs.yml
@@ -1,0 +1,30 @@
+name: Update External Docs
+on:
+    schedule:
+    - cron:  '0 8 * * 1'
+
+
+permissions:
+    pull-requests: write
+    contents: write
+
+jobs:
+    update-docs:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v3
+            - name: Run scripts
+              run: bash ./get-contributing.sh && bash ./get-readmes.sh
+            - name: Create Pull Request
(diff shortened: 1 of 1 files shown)


.github/workflows/update-external-docs.yml
```diff
@@ -0,0 +1,30 @@
...
+    update-docs:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v3
+            - name: Run scripts
+              run: bash ./get-contributing.sh && bash ./get-readmes.sh
+            - name: Create Pull Request
+              uses: peter-evans/create-pull-request@v5
+              with: 
+                token: ${{ secrets.ACTIONS_BOT_TOKEN }}
```
UlisesGascon:
```suggestion
                token: ${{ secrets.GITHUB_TOKEN }}
```

Is the automatic token an option in this case? https://docs.github.com/en/actions/security-guides/automatic-token-authentication

bjohansebas: Yes, it is the default option for the token, and now that we know it is created automatically, it is a good idea to leave it as you suggested.

UlisesGascon (approved):
LGTM!

crandmck:
I'm relying on review from @UlisesGascon to ensure that this doesn't break anything and works as expected!

Thanks @bjohansebas

PR #1561, opened 2024-08-04
remove unused css

bjohansebas:
To make the page design more maintainable, several styles that served no purpose have been removed since that content no longer exists.


css/style.css
```diff
@@ -6,15 +6,13 @@
 /* general */
 
 * {
-  -webkit-box-sizing: border-box;
-  -moz-box-sizing: border-box;
```
chrisdel101, line 10: I wonder if we don't leave these in just to be safe? The only point would be for very rare backwards compatibility cases.

Maybe with border box there is no point doing this as all browsers support it? 
Edit: I think it's okay to leave your change as is

css/style.css
```diff
@@ -349,15 +328,15 @@ pre code {
 #top {
   line-height: 0;
   background: black;
-  -webkit-border-radius: 2px;
```
chrisdel101, line 374: Also maybe leave in this webkit prefix, but also keep your added non-prefixed change. 

Edit: I can't imagine a single browser would require a prefix for border radius. So forget what I said, and leave you change as is.

css/style.css
```diff
@@ -349,15 +328,15 @@ pre code {
...
+  border-radius: 2px;
   position: fixed;
   bottom: 15px;
   right: 15px;
   padding: 8px;
   text-decoration: none;
   color: white;
   opacity: 0;
-  -webkit-transition: opacity 300ms;
```
chrisdel101, line 382: Keep it (the prefix) and also keep your change too. Of the 3 maybe this is the only one that has a remote chance of ever being needed.

chrisdel101 (commented):
It does seem all these styles are actually unused in the code. I made a few comments that I thought just couldn't hurt to leave in. 

I assume you relied on global search to find these styles? This is fine, but isn't always trustworthy. So...
I would ask that you make sure to test everything: try to visit every page of english site, and at least one or two of each translation. Just to make sure nothing is out of whack.

Nice work!

css/style.css
bjohansebas: Sure, according to [caiuse](https://caniuse.com/mdn-css_properties_box-sizing_border-box), it has good support; we shouldn't worry about that small percentage.

css/style.css
bjohansebas: The same, according to [caniuse](https://caniuse.com/border-radius), it has good support.

css/style.css
bjohansebas: This property has good support, just like the previous two. I think it would be better to remove the prefix. [ref](https://caniuse.com/mdn-css_properties_transition)

chrisdel101: Yeah okay.

bjohansebas:
@chrisdel101 I've checked several pages and it doesn't seem that removing that code would have any impact.

crandmck (approved):

crandmck:
Thanks for collaborating on this @bjohansebas and @chrisdel101 ...  I did a cursory check of the deploy preview and it looks good, but really I'm trusting that you two did full testing.

PR #1563, opened 2024-08-07
remove NodeInteractive css

bjohansebas:
It seems that this file was forgotten to be deleted when the event was completely removed several years ago, or does it have some importance for a future event?


chrisdel101:
I also looked and could not find any files that use these styles. Don't think they're used for anything.

crandmck (approved):

chrisdel101:
Thanks @bjohansebas

Issue #1556, opened 2024-08-02
Remove express v2 docs


bjohansebas:
Just like other documentation practices, keeping this version on a separate branch with a different subdomain would be beneficial for historical purposes and to avoid cluttering the main documentation that is subject to changes.

UlisesGascon:
I think that is a good idea! We can have a look in our infra and try to do the settings for this. What do you think @crandmck?

crandmck:
By separate sub-domain, do you mean that all the 5.x doc would be at something like v4.expressjs.com?  This is an interesting approach, but differs from what we've been doing, where each version's API reference is maintained  separately, e.g. https://expressjs.com/en/4x/api.html and https://expressjs.com/en/3x/api.html while the rest of the documentation is for the current version (4.x now, and once v5 is released, it would be 5.x).

> ...to avoid cluttering the main documentation that is subject to changes.

To be clear, changes are primarily in the latest/upcoming version (5.x). The older version(s) that are in bug-fix/maintenance mode typically doesn't require doc updates so the docs docs are basically "frozen" (except for stuff like correcting existing errors).

This idea is worth considering, since of course we need to maintain 4.x docs to help people maintain existing v4 apps.  However, once 5.x is released, we don't want to encourage anyone to install or use v4 as a "new user" so we'd want to remove all the v4 "Getting started" material.  If there's an argument for keeping it, I'd like to understand it.  

All the info under the "Resources" menu is not version-specific, and duplicating it is not a good idea.  Exception: I'm not sure about the [Middleware](https://expressjs.com/en/resources/middleware.html). I assume most middleware will be usable by either v4 or v5, but perhaps there are special cases?

Most of the info under the "Advanced" menu is best practices (e.g. security and performance) which are not likely to differ much between v4 and v5.  I'm not sure about [Developing template engines](https://expressjs.com/en/advanced/developing-template-engines.html) -- it may need updates for v5, but I presume we don't want to encourage people to create new template engines for v4, so I suspect we should keep only a version for v5.

So, for the most part, the only v4 doc that should be maintained/duplicated is the API doc and everything under the "Guide" menu (modulo the migration guides).  Since we already version the API docs, this proposal boils down to also versioning most of the "Guide" material.  Does it make sense to have a subdomain just for that?

bjohansebas:
[...]

PR #1565, opened 2024-08-10
Adding contents from Wiki of expressjs:express in en language

### LINKED ISSUE #251: Deprecate the Express Legacy Wiki
Hello,

The link to template engines here in the [wiki](https://github.com/expressjs/express/wiki#template-engines) is broken. The page on expressjs.com (https://expressjs.com/en/resources/template-engines.html) was removed in this [PR](https://github.com/expressjs/expressjs.com/pull/1484). So probably the link should be removed as well?

I'm not sure if it is possible to create a PR to change the wiki, so I created this issue instead. Maybe someone can help me with how make a change request for the wiki or just update it directly.

Thank you!

IamLizu:
### Overview
As we have discussed here in https://github.com/expressjs/discussions/issues/251, we are keeping docs only in the [website](https://expressjs.com/). It is easier to maintain a single place instead. This PR will make sure that anything that is in the [wiki](https://github.com/expressjs/express/wiki) but not in the website is being added. 

### New files

- [x] New features in 3x
- [x] Migrating from 2x to 3x
- [x] New features in 4x
- [x] Migrating from 3x to 4x

### Notes
1. There's pages for Migrating from 2x to 3x and New features in 3x in Wiki. However, 3x is already deprecated. So, I just added these files and referenced in `en/3x/api.md`.
2. Added the _Other modules_ from wiki home to the `en/resources/utils.md`.
3. Only focused on `en` as of now.

### Linked Issues
- Closes https://github.com/expressjs/discussions/issues/251

**Edit**: Rewrote description in order to give a better idea. Added few links so that relevant things can be navigated to from here.


IamLizu:
Before I proceed with other contents, I have a few questions,

1. I have written these two files in markdown. Other guides such as migrating-4 or migrating-5 uses HTML. Can we be lose on this one or should I rewrite?
2. I should add for other languages as well or `en` is okay for now? Perhaps we could include everything in `en` first, then move to other languages in different PR?
3. Is this okay so far?
4. Do we need to include this https://github.com/expressjs/express/wiki/ESON-JSON-Configuration?

cc: @wesleytodd @UlisesGascon

bjohansebas:
It would be better if the new pages were translated in different PRs, as it could happen like in https://expressjs.com/ru/resources/applications.html where the page is not translated into the corresponding language.

en/guide/migrating-4.md
```diff
@@ -100,11 +101,12 @@ In version 4 you can use a variable parameter to define the path where middlewar
 For example:
 
 ```js
-app.use('/book/:id', (req, res, next) => {
-  console.log('ID:', req.params.id)
-  next()
-})
+app.use("/book/:id", (req, res, next) => {
```
bjohansebas: Perhaps it is not necessary to keep changing the quotation marks in the code examples.

chrisdel101: Agreed. You should recommit it without the prettyifiing. 
We don't need to commit those.

IamLizu: Yeah, should have been careful there. I will try to recommit migrating-4.md. If it's still not clean I guess I will go back to bb564a99.

Edit: There's so much changes due to semicolon as well. I think its better if I just go back and make only the necessary changes in migrating-4.md.

Edit: Since the unnecessary prettier changes are discarded, I am resolving this one.

IamLizu:
There's pages for Migrating from 2x to 3x and New features in 3x. However, 3x is already deprecated. So, I just added these files and referenced in `en/3x/api.md`. 

Is that okay?

cc: @bjohansebas @chrisdel101

UlisesGascon:
>1. I have written these two files in markdown. Other guides such as migrating-4 or migrating-5 uses HTML. Can we be lose on this one or should I rewrite?

I think that markdown is fine, but you might require html if you need additional features that are not included in the markdown support.

>2. I should add for other languages as well or en is okay for now? Perhaps we could include everything in en first, then move to other languages in different PR?

Don't worry, we can focus on English only in this PR.

>3. Is this okay so far?

Yeah! Thanks for leading this :) 

>4. Do we need to include this https://github.com/expressjs/express/wiki/ESON-JSON-Configuration?

I don't think so, seems outdated in 2024 context :+1:

IamLizu:
Hi 👋 

Regarding the items which were included as _Pending_ in the PR description,

1. Template Engines: It was just linked to website, so nothing to add here.
2. Frameworks built on top of Express: Links to website, no other content, it seems it was removed from the website intentionally at some point or moved into different page. In any case, seems nothing to do here.
3. Boilerplate: Most of them are old, but added those in the Express examples page.

And with that, I am marking this PR available for review 🤚 

**Edit**: Forced pushed after a rebase in order to update the translation workflow file.

wesleytodd:
I am generally 👍 on this, but just want to call out there is a bunch of outdated info in this content from long ago. So it might be that we never even want to translate much of this and should take some time to clean it up first.

IamLizu:
> I am generally 👍 on this, but just want to call out there is a bunch of outdated info in this content from long ago. So it might be that we never even want to translate much of this and should take some time to clean it up first.

Make sense. However, these _outdated contents_ are in their respective file such as "migrating to 3x". **Those information are supposed to be outdated but should be kept for historical reason** in my opinion. 

Readers coming to the website will definitely read the new contents and I assume no is looking to migrate to 3.x for example, so we can safely ignore their translation as well maybe?

bjohansebas:
Is it useful to bring in the boilerplate when Express@5 is almost ready?

If it's for historical purposes, it might be better to handle it, for example, with a URL like `/v4/examples` and add the boilerplate there while keeping the main examples page as it was. What do you think about this idea?

crandmck:
Sorry I'm coming to this discussion late, as I've been out on vacation for a few weeks.

I think we should be cautious about what material we port from the wiki to the doc site, because:
- Every page in the docs potentially adds work for maintenance and translation.
- Our focus should be on 5.x docs.  We have a LOT of work to do there, without taking on additional efforts that have less value.

Let's think carefully about whether it's worth investing the time and effort to move/maintain/translate these topics:

> New features in 3x

"New" since when, 2014?  How much value does this really have for real-world users?

> Migrating from 2x to 3x

If someone (who?) still has a v2 app, do we really want to encourage them to migrate to v3?  Wouldn't it make more sense to just rewrite it for v5, or even v4?  This seems to me to have little or no real world value.

> New features in 4x
> Migrating from 3x to 4x

Again, this stuff is from around 10 years ago. The info should be covered in https://expressjs.com/en/guide/migrating-4.html but again, what's the best practice for someone (presumably not that many ppl) with a 3.x app?  Is it to move to 4.x?  Or rewrite it for 5.x?  If we really want to spend time documenting how to upgrade from 3x -> 4x, then we should just update https://expressjs.com/en/guide/migrating-4.html.

Also re this comment
> information are supposed to be outdated but should be kept for historical reason

IMHO, we should consider keeping the wiki around specifically for outdated historical/archival info.  It's far less visible than the website, and does't ever need to be translated.  These pages should clearly be marked as "OUTDATED ARCHIVED".  Yes, we can move them to the website, but is that really how we want to spend our precious time and effort?

IamLizu:
Hey @crandmck 👋 

Sorry, I guess I also somehow missed this one. 

I hear you, the _new_ files are not new at all. But they are new in their own context. For example, in docs of 2.0, features of 3.0 are new. And that is where those said _new_ features are mentioned, in their respective files (https://github.com/expressjs/expressjs.com/pull/1565#issuecomment-2303571030).

Funny thing I recently watched an conference talk of yours about how wiki should focus for developers and website for consumers. My bad, I would have at least approached a bit different way if I known better when we were discussing about deprecating the wiki in https://github.com/expressjs/discussions/issues/251 where @wesleytodd and @UlisesGascon had a affirmative indication to what I said in https://github.com/expressjs/discussions/issues/251#issuecomment-2249792768.

That said, I think no one is going to read the docs for 2.0 or 3.0 so perhaps we can just simply ignore translation for those old files?

crandmck:
Yes, honestly, now that 5.0 is released, I would advocate for removing 2.x doc altogether, and perhaps even 3.x.

IamLizu:
Yes, we could remove the old version(s) docs. However, I am being sentimental 😄 

Personally, I would keep the old ones, maybe just ignore their translation but keep en for historical record.

cc: @expressjs/express-tc

bjohansebas:
Another option would be to route deprecated APIs through subdomains, as I suggest in #1556

IamLizu:
What is our take on this?

bjohansebas:
After a long time since this PR was created,I'm not in favor of adding these resources to the page, for several of the reasons @crandmck mentioned.

Also, there are middlewares being added, and the policy since I've been here is not to add external resources, as it leads to discussions about what should and shouldn't be added. There are more important things to do here than debating whether to add such tools.

As @crandmck said, those wiki pages should be marked as "OUTDATED/ARCHIVED" This work can be done by a member of the triage team, as long as those permissions are given back to the triage team, since it seems they were disabled in the Express repository.

bjohansebas:
Thanks for the PR, but for now we won't be including these pages because they are too outdated

PR #1562, opened 2024-08-06
Refactor to es6

bjohansebas:
Using jQuery in much of this code is unnecessary. 

The menu was modified in this #1568.

PR #1568, opened 2024-08-10
new design and logic for the menu #1590

bjohansebas:
The new menu logic is designed, and the new styles are added to work correctly.

## Changes
- Remove dropit and its styles.
- The responsive design is fixed(#1589)
- New styles

### desktop
image

### mobile
image


chrisdel101:
I'm personally somewhat ambivalent about all the JS refactors (also #1562, #1569). Is it refactoring for refactoring's sake?
Yes it's older tech, but IMO it still works fine for the use cases on a simple site. 
Guess it's really up to @crandmck

bjohansebas:
> Yes it's older tech, but IMO it still works fine for the use cases on a simple site.

I agree with that, although on the other hand, wouldn't it be better to remove these dependencies that are only adding more load to the website? It might also be worth evaluating if these changes could optimize performance and maintenance in the long run.

crandmck:
Sorry for the delay in responding, I was on vacation for the last two weeks :-) 

I'm ambivalent about this.  Since @bjohansebas has already done the work, I don't know that there's a good reason NOT to accept this PR--ASSUMING that it doesn't introduce any new bugs or glitches.  Since I'm just getting back and catching up after vacay, I haven't had time to look more closely at this to ensure that's the case.  @chrisdel101 have you?

We certainly appreciate the work you put in @bjohansebas but in the future it would be great to open an issue to discuss BEFORE doing the work for a PR.  If you already did this and I just missed it, apologies, as I said I've been out for a couple weeks.  I'm not saying that we won't accept a PR without prior discussion in an issue, but that's the preferred approach :-)

bjohansebas:
@crandmck Sorry for rushing, collaborating on an open-source project is new to me. On the other hand, as you mentioned in #1562, it might be better to focus on more important things; whatever is best for the project.

chrisdel101:
I agree with you @crandmck 100% about per-discussing PRs, especially ones like this that are ethos shifting,  whereas maybe small bug fixes could slip by without?

This Improvements like this do have a legit place in web development. My thoughts are:

- It doesn't seem worth the effort to have redone this for this project. It's not like this is an application where benchmarks are essential. 
- Only 1 file using JQ (I think) App.js, plus dropit. 
- **Pro**: JQuery is coming via CDN so we're not using tons of space storing it...**Con**: although nowadays people are starting to hate on CDNs. **Con**: And the version 2.1.1 CDN could go away, sometime and then we'd need to do maintenance like this anyway.

**So, since you did all this work already it seems like a waste to throw it out.** 

The main problems are:
- There 45 files in the PR (which I've glanced at briefly and saw things I had questions about already). Yes the changes might be very similar for each file, but this is alot at once to review.
- Changing ALL the JS at once means alot of testing. 
- We need to be 100% sure the new JS works in all places before we merge this. It seems like this all might be pretty simple JS, but would take time just to double check every page since these are potentially breaking changes.

SO...
1. I'd be willing to review some of this stuff but I can't manage it across multiple PRs like it is. And, a single giant PR is not my preference either. 
2. My thought is that you might have to redo these :( (not the work, just the arrangement)
3. If it's possible could you make us a simple listing (like a proposal) of the content you want to include across the PRs? Then implelement the first, like just redo drop-it in one PR, merge, then move to the next. This way they are not relying on each other.
4. If the 3 PRs _already_ satisfy this, then having a proposal type-of -thing describing their order would help us see what is happening across all three.
5. Your "proposal" could just be an ordered list added to this thread.

This response got WAY too long and rambling. Did this make sense?

bjohansebas:
**First PR (Refactor Menu):**
- Remove dropit
- Remove functions related to dropit
- Refactor the menu according to the new structure
- Move menu styles to a new file
- Add new menu functionality in a new file

**Second PR (Remove Unused Functions):**
- Remove ismobil.js (
- Remove logic for page editing (no longer used since the footer was updated)
- Remove logic for [`window.location.pathname === '/3x/api.html'`](https://github.com/expressjs/expressjs.com/blob/83cf6069eca7670210c856fca51d6db4dded2c0d/js/app.js#L121) (no longer used since almost all URLs now go to `/en/*`)
- Remove function related to [`#tags-side-menu`](https://github.com/expressjs/expressjs.com/blob/83cf6069eca7670210c856fca51d6db4dded2c0d/js/app.js#L155)

**Third PR (Adapt get elements):**
- Update scroll top logic (@chrisdel101, should we keep the scroll logic to add a style when not at the top of the page?)
- Adapt element retrieval and events
- Add overlay logic to the menu
- Adapt code highlighting

**Fourth PR (Adapt Functions):**
- Update logic for the i18n banner
- Update logic for the closer title (used in the API)
- Adapt cookies logic to ES6

**Fifth PR (Remove jQuery Dependency):**
- Remove link to jQuery

Note: While writing the proposal, I realized that more PRs are needed to make the review easier.

@chrisdel101, do you agree with the proposal?

chrisdel101:
Yes this looks good to me overall. Starting w/ the menu as you said is a good plan. 
I think we can go ahead with PR # 1 as a separate PR, if you okay with that, at which point we'd prob just close this PR. But it'll be nice to have for this list and discussion. 

Some things to consider:
- please try your best not to commit any prettifying changes (not saying you did but there are tons of changes in the header files and I can't tell if they are legit, or just prettifying). It's way easier to read and see the code changes without huge blocks of green and red. 
- I like menu.css having a sep file too! But then I can't easily tell what has been changed. Not sure what the answer to this is...but just saying. 
     -  Maybe making that separate file a single change in one of the future PRs? Unless this is too gross.
- I like the CSS/markup menu fixes since it seems to fix things that were broken, at least on FF, and this is why I'd like to see clearly exactly what's been changed.

Is this okay @crandmck?

crandmck:
Yes, this is a good approach.  Thanks for taking the lead on this @chrisdel101.  And, of course, thanks @bjohansebas for all the refactoring effort.

To help coordinate and simplify reviews, I would suggest opening an issue with the content of the [above comment](https://github.com/expressjs/expressjs.com/pull/1568#issuecomment-2316523369) and then link that to all the PRs.

js/ismobile.js
```diff
@@ -1 +1 @@
...
...undefined"!=typeof s[1]&&(r=s[0]),s=r.split("Twitter"),"undefined"!=typeof s[1]&&(r=s[0]),this.apple={phone:q(b,r),ipod:q(c,r),tablet:!q(b,r)&&q(d,r),device:q(b,r)||q(c,r)||q(d,r)},this.amazon={phone:q(g,r),tablet:!q(g,r)&&q(h,r),device:q(g,r)||q(h,r)},this.android={phone:q(g,r)||q(e,r),tablet:!q(g,r)&&!q(e,r)&&(q(h,r)||q(f,r)),device:q(g,r)||q(h,r)||q(e,r)||q(f,r)},this.windows={phone:q(i,r),tablet:q(j,r),device:q(i,r)||q(j,r)},this.other={blackberry:q(k,r),blackberry10:q(l,r),opera:q(m,r),firefox:q(o,r),chrome:q(n,r),device:q(k,r)||q(l,r)||q(m,r)||q(o,r)||q(n,r)},this.seven_inch=q(p,r),this.any=this.apple.device||this.android.device||this.windows.device||this.other.device||this.seven_inch,this.phone=this.apple.phone||this.android.phone||this.windows.phone,this.tablet=this.apple.tablet||this.android.tablet||this.windows.tablet,"undefined"==typeof window)return this},s=function(){var a=new r;return a.Class=r,a};"undefined"!=typeof module&&module.exports&&"undefined"==typeof window?module.exports=r:"undefined"!=typeof module&&module.exports&&"undefined"!=typeof window?module.exports=s():"function"==typeof define&&define.amd?define("isMobile",[],a.isMobile=s()):a.isMobile=s()}(this);
```
bjohansebas, line 1: It doesn't contain any changes, it's just that when I copied the file again, GitHub recognized it as if I had changed it. This file would be deleted in the second PR.

bjohansebas:
@chrisdel101, you can review now. I fixed some bugs I found in Chrome, so everything should work fine now.

css/dark-theme.css
```diff
@@ -44,6 +43,11 @@
   height: 27px;
   cursor: pointer;
 }
+
+html.dark-mode .button {
+  color: #e5e5e6;
```
chrisdel101: Can you make this a variable? Might as well.

js/menu.js
```diff
@@ -0,0 +1,73 @@
...
+	if (event.matches) {
+		isSmallScreen = true;
+	} else {
+		document.body.classList.remove("no-scroll")
+	}
+});
+
+// Desktop Menu
+
+const $itemsMenu = document.querySelectorAll(".submenu");
```
chrisdel101: I don't know about the $ sign in the variables in this file. I've always seen this as a JQuery convention. I don't see the point of starting with a $. Suggest you remove, but this is not mandatory.

js/menu.js
```diff
@@ -0,0 +1,73 @@
...
+		if (isSmallScreen || 'ontouchstart' in document.documentElement) {
+			e.preventDefault();
+		}
+	});
+}
+
+document.querySelector("#nav-button").addEventListener("click", () => {
+	$menu.classList.toggle("opens");
+	$overlay.classList.toggle("blurs");
+	document.body.classList.toggle("no-scroll")
```
chrisdel101: If `$menu` is undefined then calling `$menu.classList.toggle` will cause an error.  I'd suggest using the optional chaining functionality here: `$menu.classList?.toggle.`

js/menu.js
```diff
@@ -0,0 +1,73 @@
...
+});
+
+// close mobile menu
+$overlay.addEventListener("click", () => {
+	$menu.classList.remove("opens");
+	$overlay.classList.remove("blurs");
+	document.body.classList.remove("no-scroll")
```
chrisdel101, line 70: Suggest optional chaining here too

js/menu.js
```diff
@@ -0,0 +1,73 @@
+// It uses `window.matchMedia` instead of `window.addEventListener('resize')` because `matchMedia` performs better by not changing its value with every screen resize.
+const mobileScreen = window.matchMedia("(max-width: 1064px)");
+let isSmallScreen = false;
+
+mobileScreen.addEventListener("change", (event) => {
```
chrisdel101: It might be useful to have errors show up like this when they occur (so this means not using optional chaining)
VM1798:1 Uncaught TypeError: Cannot read properties of undefined (reading 'addEventListener')
    at <anonymous>:1:15

But, If you don't want to the errors to show up, consider adding optional chaining to this, and to `$itemsMenu` listener below too.

css/dark-theme.css
```diff
@@ -24,9 +24,8 @@
 }
 #theme-icon-container {
   position: relative;
-  float: right;
-  color: white;
-  min-width: 50px;
+  min-width: 27px;
+  min-height: 27px;
```
chrisdel101: These lines are no longer needed. `sun-icon` has the settings needed.

css/dark-theme.css
```diff
@@ -69,27 +73,47 @@ html.dark-mode #logo > a {
...
+
+html.dark-mode #navmenu > li {
+  background: var(--second_dark_bg);
 }
-/* drop down menu links  */
-html.dark-mode #navbar .menu ul.dropit-submenu a:hover {
-  background: var(--dark_hover);
+
+html.dark-mode #q {
+  border-color: #979797 !important;
```
chrisdel101: **We don't want to use important.** 
Make the element more specific using `div#q`, something like this might work
`html.dark-mode section#logo ~ div#navbar  input#q`

I remember using it here b/c someone else used it in style.css so I had to use to overide that. We only want important as a last resort. 

If it wasn't used in the style.css already, I'd say we should NEVER use it.

css/dark-theme.css
```diff
@@ -69,27 +73,47 @@ html.dark-mode #logo > a {
...
+}
+
+html.dark-mode .algolia-autocomplete #q:focus-visible,
+html.dark-mode .algolia-autocomplete #q:focus {
+  border-color: var(--dark_border) !important;
```
chrisdel101: Again, **We don't want to use important.**

css/dark-theme.css
```diff
@@ -69,27 +73,47 @@ html.dark-mode #logo > a {
...
+
+/* drop down menu links  */
+
+html.dark-mode #navbar #navmenu li ul {
   background: var(--second_dark_bg);
 }
+
+@media all and (max-width: 1064px) {
+  html.dark-mode #navbar #navmenu li ul {
+    background: var(--main_dark_bg) !important;
```
chrisdel101: Get rid of important

css/dark-theme.css
```diff
@@ -69,27 +73,47 @@ html.dark-mode #logo > a {
...
+  }
+
+  html.dark-mode .submenu-content li a:hover {
+    background: var(--dark_hover) !important;
+    color: var(--dark_main_text) !important;
```
chrisdel101: These are already specific but I'd suggest trying more even, using important ONLY when no other choice.

css/style.css
```diff
@@ -679,67 +674,82 @@ footer {
...
   display: block;
   color: #777;
-  padding: 0 8px 2px 8px;
+  padding: 2px 20px 2px 20px;
 }
 
-.menu ul.dropit-submenu a:hover {
+.submenu-content a:hover {
   background: #259dff;
   color: #fff !important;
```
chrisdel101: This is the important that causes problems. Maybe this can go? 
I suggest we remove this and never use important (see dark-theme.css for more)

css/style.css
```diff
@@ -850,13 +860,22 @@ h2 a {
...
+  border-radius: 8px;
+  border: 1px solid #ddd;
+  transition: color .3s ease;
+  padding-inline: 12px;
+  outline: none;
 }
+
+.algolia-autocomplete #q:focus-visible,
+.algolia-autocomplete #q:focus {
+  border-color: #333 !important;
```
chrisdel101: Ge rid of important if possible. Looks like it might have been there already, but would be good go get rid of it now!

css/style.css
```diff
@@ -1347,4 +1256,133 @@ html.dark-mode .hidden-light {
...
+    column-gap: 12px;
+    padding-inline: 32px;
+  }
+
+  #mobile-menu {
+    display: block;
+  }
+
+  .algolia-autocomplete {
+    display: none !important;
```
chrisdel101, line 1284: Remove important

css/style.css
```diff
@@ -1347,4 +1256,133 @@ html.dark-mode .hidden-light {
...
+    font-size: 19px;
+    margin: 0;
+    width: 100%;
+    height: 100%;
+    padding-left: 5%;
+  }
+
+  .submenu.open {
+    flex-direction: column;
+    align-items: initial !important;
```
chrisdel101: remove important

css/style.css
```diff
@@ -1347,4 +1256,133 @@ html.dark-mode .hidden-light {
...
+  .submenu-content li:last-child a {
+    border-radius: 0;
+  }
+
+  .submenu-content.open {
+    display: inline-block;
+  }
+
+  .submenu-content li a {
+    font-size: 16px !important;
```
chrisdel101: remove importants

css/style.css
```diff
@@ -1347,4 +1256,133 @@ html.dark-mode .hidden-light {
...
+    padding-left: 10% !important;
+  }
+
+  .submenu-content li {
+    border-bottom: 1px solid #cdcdcd;
+  }
+
+  .submenu-content li a:hover {
+    background: #cccccc;
+    color: #666 !important;
```
chrisdel101: remove important

js/ismobile.js
chrisdel101: Thanks for this comment.

chrisdel101 (changes requested):
I wasn't a super reviewer on the header files since those are tricky to review. If there's any typos there hopefully we get them over time.
I left a bunch of comments to fix, and some suggestions too. 
Great job with this. The navbar is finally a flex box! And the French nav is the correct way around.

js/menu.js
bjohansebas: The use of `$` is to make it easier to detect DOM elements; it's a coding style I've had. If you don't like it, I can change it.

js/menu.js
bjohansebas: done

`$itemsMenu` wouldn’t be necessary since if it doesn’t find any elements, no event will be added.

css/dark-theme.css
bjohansebas: done

css/dark-theme.css
bjohansebas: done

js/menu.js
chrisdel101: It's definitely not the end of the world. It just makes me think that it's jQuery when it isn't. You can probably leave it.

css/dark-theme.css
bjohansebas: done

css/dark-theme.css
bjohansebas: done

css/style.css
bjohansebas: This `!important` is necessary since the styles for the search are injected by an external source.

chrisdel101:
Was this one good to go? It's been awhile. @bjohansebas This ones fixes the nav bar squishing issue if I recall?
If the issues were all fixed I can review again.
@crandmck - can we look at this one to get rid of the nav squishing error?
I've been less active lately, but just saw this still happens on the site so wanted to follow up.

bjohansebas:
Yes @chrisdel101, it should be ready now. I was planning to add the language selector, but that will be for another PR.

bjohansebas:
I just noticed that this PR has conflicts. I can fix it next week.

bjohansebas:
I prefer not to impose a coding style. @chrisdel101, can you give it one last review so we can release it?

chrisdel101 (commented):
I think maybe the express logo is too far in in large screens? If you agree, maybe this could stick the left corner more? I put the screen measurements in the corner so you can see where it's happening, or don't have a large screen.

express_logo

I glanced at all the translations home pages and they all look good. For these, the logo is even further in, like https://deploy-preview-1568--expressjscom-preview.netlify.app/zh-tw/.   

Unless this is what you are going for?

I got into testing this, and found a solution. I'll put it here, but you can do it your own way. I just spent the time playing around so thought I'd put it here. Or just ignore my code here. It's just an idea.
remove the header col gap
col_gap
Add flex grow and margin-left
logo
nav
And then it seems to fix the translations too. 
tn_chin

I do personally agree with removing the `$` in the JS too just to keep the style consistent. 
Everything else looks good that I checked. Once again, nice work. And thanks for making the first round of changes.

bjohansebas:
@chrisdel101, thanks for the suggestion. I've opted for a different solution, let me know if you like it or not.

js/menu.js
```diff
@@ -0,0 +1,76 @@
+// It uses `window.matchMedia` instead of `window.addEventListener('resize')` because `matchMedia` performs better by not changing its value with every screen resize.
+const mobileScreen = window.matchMedia("(max-width: 1110px)");
```
chrisdel101, line 2: What about moving these into css variables and grabbing from there? Same with `ismobile.js` so you have it all in one place? 
Just a suggestion and not a requirement. Having css in a file that is not being used there can get messy, but this might be a way to have a single spot for the media settings, in case there being used in multiple places.
Totally optional.

chrisdel101: After looking again I'm seeing this is the only time you set something like this. I guess I was just seeing multiples due to the changes.

chrisdel101:
I like it. Just maybe a few adjustments to the nav styles? For me there is too much space btw the logo and the search bar. What if you move the `theme-icon-container` out of the `nav-menu`? The only down side to this is that it requires changing every single header file
So like this
Screenshot 2024-10-22 at 9 45 42 PM

Then you don't need to change any css
You get this 
Screenshot 2024-10-22 at 9 32 05 PM
Instead of this
Screenshot 2024-10-22 at 9 32 14 PM
If you dont want to change all the headers, you could also just flip to a `space-around` on the header. But I'm on only on a laptop screen rn so can't test a large screen to see if that fixes the original prob.
At the very least, I think there needs to be some space btw the "blog" nav item and dark/light button on the far right. And also maybe some space btw the search bar and  "Home".

js/menu.js
bjohansebas: Can CSS variables be brought into JavaScript?

bjohansebas:
I liked separating the `theme-icon-container` from the `nav-menu` more, this would help to add a language selector in the near future. I'll make the change later when I have some time

js/menu.js
chrisdel101: Yeah. With `getComputedStyle`. It'd be something like this, which could be made into a single helper function. All constants stored in the CSS.
  
```
# in CSS :root  { --max-width: 1110px} 

const style = getComputedStyle(document.body)
const maxWidth= style.getPropertyValue('--max-width') 
const mobileScreen = window.matchMedia($`(max-width:${maxWidth)`);
```

I just thought having constants in one place would be nice. This not a best practice I've seen used before though. Just an idea.

On another note, how about a new PR where we put the style.css re-used vals into variables? Since that file does not use vars. I could work on this after my current issue is open? I'll open an issue if you like the idea.

bjohansebas: Thank you, I'm going to try it.

> how about a new PR where we put the style.css re-used vals into variables?

Are you referring to the reused values like the colors? If that's the case, i think it's a great idea, i recently opened an issue to define this, see #1628.

bjohansebas: Can CSS variables be used for the condition of a media query? I tried to add them, but I think it's not possible, I have to do things that I don't think are a good practice for it to work.
```
@media all and (max-width: var(--nav-mobile)) {
```

Maybe you can do it in another pr when we launch this?

chrisdel101: I think it’s good like u have it too.  So we can leave this idea for now.

chrisdel101 (changes requested):
Looks great! Sorry I missed this before.
So the only thing I notice is that the nav menu breakpoint kicks in too early which makes it awkward to use on a laptop. I def want to see the nav links still at the same size in the new format. Can you lower the breakpoint to kick in at like 899 (like it was before?)
This is the current way. I can see the nav menu which I def want at this size.
Screenshot 2024-10-27 at 9 28 57 AM
And this is changed way
Screenshot 2024-10-27 at 9 30 36 AM
I guess it's this rule but I can't seem to change it in fire fox so I'm not able to see what it looks like. Change this back?
Screenshot 2024-10-27 at 9 36 31 AM
Old setting (with nav hidden) - can u see the breakpoint to this value?
Screenshot 2024-10-27 at 9 40 44 AM

bjohansebas:
@chrisdel101 Yeah, that's true, but the problem is that there are many items in the menu that, if it doesn't start from that point, it breaks the layout and this issue(#1589) would occur.

bjohansebas:
I believe all the issues are already resolved, we can always make changes later. 

I need the approval of a person with write permissions to merge it (@UlisesGascon, @expressjs/docs-captains).

chrisdel101:
Yeah your right. Maybe we can hide the last few links or something, since I don't want to have to click the hamburger all the time. 
But yeah, can deal with this later then. 

LGTM.

bjohansebas:
I've brought the changes made in #1674 and #1667 to avoid conflicts

UlisesGascon (approved):

bjohansebas:
@chrisdel101 thanks for helping me get this onto the website

PR #1569, opened 2024-08-10
Remove jQuery 

bjohansebas:
Using jQuery is unnecessary

This PR is part of #1562 and #1568


crandmck:
Can this https://github.com/expressjs/expressjs.com/pull/1562 and https://github.com/expressjs/expressjs.com/pull/1568 be merged separately, or do they all need to be merged at the same time?

bjohansebas:
I did it so they can be merged separately and thus tested better.

Issue #1566, opened 2024-08-10
Netlify deploy error


bjohansebas:
As has happened in the following PRs (#1565, #1563, #1562), the deploy of these pages continues to encounter the same installation error.

```shell
11:59:12 AM: build-image version: ecdc8b770f4a0193fd3f258c1bc6029e681813a4 (focal)
11:59:12 AM: buildbot version: 778f625377bd173cd306eb0b71ebeb25fdfff620
11:59:12 AM: Fetching cached dependencies
11:59:12 AM: Starting to download cache of 217.4MB
11:59:14 AM: Finished downloading cache in 1.381s
11:59:14 AM: Starting to extract cache
11:59:16 AM: Finished extracting cache in 1.939s
11:59:16 AM: Finished fetching cache in 3.376s
11:59:16 AM: Starting to prepare the repo for build
11:59:16 AM: Preparing Git Reference pull/1565/head
11:59:18 AM: Starting to install dependencies
11:59:18 AM: Python version set to 3.8
11:59:18 AM: Attempting Ruby version 2.7.2, read from environment
11:59:19 AM: Using Ruby version 2.7.2
11:59:19 AM: Started restoring cached ruby gems
11:59:19 AM: Finished restoring cached ruby gems
11:59:19 AM: Installing gem bundle
11:59:19 AM: [DEPRECATED] The `--path` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set --local path '/opt/build/cache/bundle'`, and stop using this flag
11:59:19 AM: [DEPRECATED] The --binstubs option will be removed in favor of `bundle binstubs --all`
11:59:21 AM: Fetching gem metadata from https://rubygems.org/...........
11:59:21 AM: Resolving dependencies...
11:59:22 AM: Failed during stage 'Install dependencies': dependency_installation script returned non-zero exit code: 1
11:59:22 AM: Could not find compatible versions
11:59:22 AM: Because nokogiri = 1.16.2 depends on Ruby >= 3.0.0
11:59:22 AM:   and nokogiri >= 1.16.2, < 1.16.3 depends on Ruby >= 3.0, < 3.4.dev,
11:59:22 AM:   nokogiri >= 1.16.2, < 1.16.3 requires Ruby >= 3.0.0.
11:59:22 AM: So, because nokogiri = 1.16.3 depends on Ruby >= 3.0.0
[...]

PR #1570, opened 2024-08-12
Sun icon missing everywhere except En Header - Add it

chrisdel101:
The sun icon was only added to English. Might as well add it to the entire site. 

Also moon icon was 2x size originally and was changed. Re-add the 2x since it looks better. 

Changes here should only be the addition of the sun icon to every header file.


chrisdel101:
All the tests are failing.  Looks like ruby version error?
I've never had this happen here before. 

This PR should only be a single commit but those 5 merges are tacked on for some reason. Maybe it's related to those....
Looks like a ruby versioning error. Not even where to set this?

Edit: I see it's actually #1566

crandmck:
#1566 was merged, but tests are still failing.  @chrisdel101  I wonder if a trivial commit to this fork/branch might re-run the tests so they would now pass?

chrisdel101:
Seems fixed after a full re-merge with main.

carlosstenzel (approved):

crandmck (approved):

crandmck:
Thanks again @chrisdel101 !

PR #1574, opened 2024-08-16
fix ruby version in Netlify

bjohansebas:
After several attempts to fix this error and considering that it could be a version issue, as tried in PR (#1573) but in Docker, I noticed that Netlify uses an outdated default version. Although this change can be managed through environment variables, I don’t think it's best to handle it that way. Instead, it should be specified in a file so that administrators don’t have to manually change the version through the settings.

Resources:
https://answers.netlify.com/t/cant-deploy-with-ruby-3-1-2/55728
https://docs.netlify.com/configure-builds/available-software-at-build-time/
https://docs.netlify.com/configure-builds/manage-dependencies/#ruby

Close:
#1566 

Edit: I just realized that the Docker configuration also needs to be fixed.
Edit: I'm not sure if these changes will work on GitHub Pages. If they don't work, it will be necessary to change the version using environment variables.
Edit: I have been testing in my repository, and GitHub Pages seems to work fine, so setting a specific Ruby version wouldn't be a problem.


UlisesGascon (approved):
Thanks @bjohansebas for working on this!

UlisesGascon:
I will merge it as this will unblock previews and CI pipelines failing in many PRs

PR #1578, opened 2024-08-20
Branding content

bjohansebas:
There are several questions about what the logo for Express (https://github.com/expressjs/discussions/issues/262) is, its usage, or whether they plan to have an official logo. To clarify these doubts, it would be helpful to have a dedicated page for the project's branding and to inform users that the Express brand is under the domain of the OpenJS Foundation.


crandmck:
I'm curious whether the @expressjs/express-tc thinks this is a good idea.  

If so, I suggest adding a section to the existing [Community](https://expressjs.com/en/resources/community.html) page rather than add a whole new page.

wesleytodd:
I actually think this is a GREAT addition. I do agree it might be best to add to the existing page, but the fact that the mark is owned by the foundation and that there are polices around it's use is something we should for sure include.

crandmck:
This looks good, but merging other PRs has led to merge conflicts.  Can you pls address @bjohansebas ..?

bjohansebas:
@crandmck, you can review now.

crandmck (approved):

PR #1583, opened 2024-08-23
Change wildcard to catch routes

### LINKED ISSUE #1408: express 5 router wildcard does not work as documented
Hi,

The following does not work as written in the documentation:
https://expressjs.com/en/5x/api.html

```
app.all('*', requireAuthentication, loadUser)

```

This is because you are at a newer version of path-to-regexp

https://github.com/pillarjs/path-to-regexp#compatibility-with-express--4x

so please update the documentation because this works

```
app.all('(.*)', requireAuthentication, loadUser)
```

I think that also the migration doc should reflect this change.

Thanks

IamLizu:
As #1408 points out, the wildcard defined no longer works. So, replaced `*` with `(.*)` in 5.x specific files. I have also installed the 5.x from npm and verified the wildcard change for root level and api level routes present in the docs.

#### Notes

- Closes #1408

### DIFF
--- a/_includes/api/en/5x/app-all.md
+++ b/_includes/api/en/5x/app-all.md
@@ -25,14 +25,14 @@ can perform a task, then call `next()` to continue matching subsequent
 routes.
 
 ```js
-app.all('*', requireAuthentication, loadUser)
+app.all('(.*)', requireAuthentication, loadUser)
 ```
 
 Or the equivalent:
 
 ```js
-app.all('*', requireAuthentication)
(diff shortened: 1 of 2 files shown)


bjohansebas:
@IamLizu It seems that this PR has the same changes as #1369, Which one should we keep?

IamLizu:
> @IamLizu It seems that this PR has the same changes as #1369, Which one should we keep?

We should definitely land the existing one. My bad I didn't check for this. I am closing this in reference of #1369

Issue #1480, opened 2024-04-02
Proposal to map out outdated translations


UlisesGascon:
In our recent TC meeting, we discussed the need for a more efficient method of detecting outdated translated versions of our documentation. Below is a proposal aimed at addressing this issue and enhancing the user experience for our contributors.

### Assumptions

- Typically, changes are made to the English version of the documentation first, without waiting for translations to be completed. (For example, see: [PR #1474](https://github.com/expressjs/expressjs.com/pull/1474))
- Our community is eager to assist with translations ([Issue #1237](https://github.com/expressjs/expressjs.com/issues/1237)), but it's challenging for them to initiate the process.
- Identifying which parts of the documentation have changed and need updating in their respective translated versions is difficult, leading to outdated translations.

### Workflow

I propose creating a GitHub Action similar to the one discussed in [Issue #1479](https://github.com/expressjs/expressjs.com/issues/1479). This action would detect changes made in any version using regex for the paths. It would then add tags in the PR for other versions that need translation, such as `needs fr translation` and `needs sp translation`. This tagging system would enable us to easily identify closed PRs requiring translation. For instance, we could filter closed PRs for the French version using a tag-based search: [French Version PRs](https://github.com/expressjs/expressjs.com/pulls?q=is%3Apr+is%3Aclosed+label%3A%22needs+fr+translation%22).

If a change doesn't require translation, such as fixing a typo in the English text, we can simply remove the corresponding tags before merging the PR.

### Next Steps
- [ ] Develop a GitHub Action to support this workflow.
- [ ] Update the `CONTRIBUTING.md` guide to incorporate this workflow.

crandmck:
Great idea!  Keep in mind that the API doc is not translated, so this only applies to the rest of the doc.

carlosstenzel:
@crandmck  @UlisesGascon  @bjohansebas  @chrisdel101 

Hello team,

I developed a script that creates a page with translation statuses.

But :´( the action doesn't work

![Captura de Tela 2024-08-30 às 21 02 22](https://github.com/user-attachments/assets/689353a8-ee73-4a46-808a-58bbfe417a84)

I think it could be a configuration for action

![Captura de Tela 2024-08-30 às 21 02 04](https://github.com/user-attachments/assets/eb845aba-5b58-4917-a056-331e97aaab37)

I forced the manual creation of the page, to view

https://deploy-preview-1586--expressjscom-preview.netlify.app/en/resources/status-translate

[...]

PR #1586, opened 2024-08-30
feat - create page by action to verify status - (TEST)

carlosstenzel:
Captura de Tela 2024-08-30 às 20 23 52

Ref https://github.com/expressjs/expressjs.com/issues/1480


.github/workflows/status-translation.yml
```diff
@@ -0,0 +1,50 @@
...
+       contents: write
+       pull-requests: write
+    steps:
+      - uses: actions/checkout@v3
+        
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+      - run: | 
+          npm install
```
bjohansebas: Why do we need to install the dependencies?

translationStatusScript.js
```diff
@@ -0,0 +1,109 @@
...
+
+verifyLanguage.forEach(() => {
+  content = content + '---|'
+})
+
+content = content + '\n'
+
+const ignoreFiles = ['resources/status-translate']
+
+const languageBaseFiles = ['guide', 'resources', 'resources/middleware', 'starter', 'advanced', '3x', '4x', '5x']
```
bjohansebas, line 40: It is not necessary to keep checking if the API versions are translated, and I don't think it will be necessary to keep looking at 'resources/middleware' either, as it is an external resource that is extracted by running a script.

bjohansebas:
It would be better if this workflow were a task that repeats periodically and checks the labels of the PRs as proposed in #1480.

carlosstenzel:
> It would be better if this workflow were a task that repeats periodically and checks the labels of the PRs as proposed in

we can create the schedule in main (gh-pages), but I don't know if the repository is configured to allow direct commits, so I want to do the same thing on every pull request

ref - https://docs.github.com/pt/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule

.github/workflows/status-translation.yml
carlosstenzel: Thanks for the feedback, it really doesn't make sense, I must have added it for testing and forgot to remove it

translationStatusScript.js
carlosstenzel: in the language (th) there are translated 'resources/middleware'

https://deploy-preview-1586--expressjscom-preview.netlify.app/en/resources/status-translate

Captura de Tela 2024-09-01 às 22 09 38

the API page has been translated into other languages

crandmck:
This is potentially useful, but the whole area of localization (translation) probably needs to be reviewed and discussed more holistically... 

Anyway, I have a few questions & comments on this PR specifically:
- How does this information get generated? 
- Simply determining if a page has been translated is not that helpful (it's pretty obvious if the page doesn't exist in the locale dir or if it's still in en). What would be more useful would be determining if the en version matches the localized version, which is likely quite difficult without translation AI.  Even date-checking the localized version vs. en version migt not be sufficient.
- This probably doesn't need to have its own page (it's only of interest to a small subset of contributor), but rather should be linked from a Documentation section in the Contributing page.

TBH, I'd like to discuss localization with the @expressjs/express-tc in general, and since we may have access to IBM translation AI, consider re-translating the whole site once we get the docs ready for 5.x.

bjohansebas:
Moving to Crowdin — thanks for this PR, you did a great job, but I think it's no longer needed.

I'm contacting Crowdin support, and you'll have more updates about the integration soon

PR #1587, opened 2024-08-31
Fix support link

bjohansebas:
This fixes the links in version 2 that were returning 404 errors.


crandmck:
I really think we should get rid of the 2.x docs, and maybe even the 3.x docs, with assent from the TC.

 I guess there's no harm in landing this in the meantime, though.

bjohansebas:
@crandmck While deciding whether to remove those documentations, it's better to merge this change to avoid 404 errors on those pages.

UlisesGascon (approved):

PR #1588, opened 2024-09-02
"Using template engines": Remove outdated info and broken links

### LINKED ISSUE #1290: URL Not Working - Comparing JavaScript Templating Engines: Jade, Mustache, Dust and More
https://expressjs.com/en/guide/using-template-engines.html

Comparing JavaScript Templating Engines: Jade, Mustache, Dust and More

https://github.com/expressjs/expressjs.com/blob/57d0dc447717bb5e3f24bf44dfc9c9923ac4f20f/en/guide/using-template-engines.md?plain=1#L20

crandmck:
Closes #1290 

Currently, this is en-only, but at some point the same changes should be done in other langs.

### DIFF
--- a/en/guide/using-template-engines.md
+++ b/en/guide/using-template-engines.md
@@ -11,19 +11,9 @@ A _template engine_ enables you to use static template files in your application
 variables in a template file with actual values, and transforms the template into an HTML file sent to the client.
 This approach makes it easier to design an HTML page.
 
-Some popular template engines that work with Express are [Pug](https://pugjs.org/api/getting-started.html),
-[Mustache](https://www.npmjs.com/package/mustache), and [EJS](https://www.npmjs.com/package/ejs).
-The [Express application generator](/{{ page.lang }}/starter/generator.html) uses [Jade](https://www.npmjs.com/package/jade) as its default, but it also supports several others.
+The [Express application generator](/{{ page.lang }}/starter/generator.html) uses [Pug](https://pugjs.org/api/getting-started.html) as its default, but it also supports [Handlebars](https://www.npmjs.com/package/handlebars), and [EJS](https://www.npmjs.com/package/ejs).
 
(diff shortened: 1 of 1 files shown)


en/guide/using-template-engines.md
```diff
@@ -36,21 +26,21 @@ $ npm install pug --save
 ```
 
 <div class="doc-box doc-notice" markdown="1">
-Express-compliant template engines such as Jade and Pug export a function named `__express(filePath, options, callback)`,
-which is called by the `res.render()` function to render the template code.
+Express-compliant template engines such as Pug export a function named `__express(filePath, options, callback)`,
+which `res.render()` calls to render the template code.
 
 Some template engines do not follow this convention. The [Consolidate.js](https://www.npmjs.org/package/consolidate)
```
bjohansebas: The author of `consolidate.js` says that in the future the package will stop receiving updates in favor of `@ladjs/consolidate`. It might be a good idea to update the link.

bjohansebas:
I have an off-topic question, why isn't `Developing template engines for Express` in the advanced topics menu?

en/guide/using-template-engines.md
crandmck: Updated

bjohansebas:
LGTM

bjohansebas (approved):

chrisdel101:
Just wondering if we should list _which_ templates are currently supported by express? I guess we'd need a core person to let us know that. 

When I read "but it also supports...among others" I was like, which others?

I found this old link https://www.expressjs.com.cn/resources/template-engines.html, but not sure if all of these are still valid. Also, I think this page should be removed (I'll make an issue for it).

I'm being nit picky here, so this is obviously not urgent :)

bjohansebas:
@chrisdel101 Maybe you want to put it in an issue, your comment could get lost over time, and I think this could be a good idea.

chrisdel101:
That link above must be some site clone, so the link is not valid. But this issue persists. Thanks for heads up.

PR #1593, opened 2024-09-03
Dark Mode Flash fix (#2)

### LINKED ISSUE #1508: Dark Mode flash

chrisdel101:
Attempt to fix #1508. Moves the theme.js code into the head file and removes the js file. There is one small change in the JS that I will highlight in a comment. 

- After merging this PR, check if the flash is fixed in prod.
- If it is not fixed, then that means this will likely not solve the issue. 
- Revert the commit and undo the merge - we don't want the JS in the head unless it actually solves the problem, so this needs to be reverted.

For Reference: Dark mode flash fix (# 1) #1524 failed.


_includes/head.html
```diff
@@ -4,7 +4,103 @@
...
+                    }
+                }
+                document.addEventListener('DOMContentLoaded', () => {
+                    const themeToggleElement = document.querySelector('.theme-toggle');
+
+                    if (themeToggleElement) {
+                        themeToggleElement.addEventListener('click', toggleStorageTheme);
+                    } else {
+                        console.error('Error: document object is not available');
+                    }
```
chrisdel101, line 45: These few lines are the only new lines added. The rest is just straight copy.

Added these just to add an error message if the attaching fails.

bjohansebas:
Another approach could be to add the script right after opening the `<body>` tag, so we wouldn’t have to wait for the document to fully load. Just a thought.

chrisdel101:
I don't think it'll work b/c it requires the document obj to add a listener to. It cannot add that if the document is not loaded (it's null) So it's not possible to load _this_ script without a document obj.

This version _should_ load right after the page tags loads and attach the classes to the html tag, which is available way before the body tag. And it does this in every instance expect our prod version. I know this b/c I tried it on body and it was way slower.

I don't know of a faster loader than DOMcontentLoaded to access the `document`. 

That said, there's some of debugging I can think of to try like running only the css without any JS, and see if the flash is there even then, but I'd need a staging environment where non admins could deploy stuff to test, and GH pages doesn't support this.

If this PR doesn't work, which I'm almost certain it will not, I'm open to your idea though. If you mean putting the script inside the top of the body? I've never done that, but who knows? Does that allows us to acess the document sooner?
At the same time, this is low priority I'd think, so you don't have so spend time on it unless you want to.

bjohansebas (approved):
LGTM, hopefully it works

bjohansebas:
Please see https://github.com/expressjs/expressjs.com/issues/1508#issuecomment-2336869550.

chrisdel101:
Wow. Nice sleuthing. I actually never did check the connections panel during this debug! :(
AND I saw those tags around and had no idea what they were (before my time), and never looked them up.  

I've also never heard of rocket thing, but good catch. It's almost certainly that!

Closing this PR and remaking with that change.

PR #1599, opened 2024-09-09
Remove confusing note from the 4x documentation

### LINKED ISSUE #1431: Incorrect default value for dotfiles in docs for express.static
I think the default value mentioned at https://github.com/expressjs/expressjs.com/blob/gh-pages/_includes/api/en/4x/express.static.md#expressstaticroot-options is incorrect. It says in the table that the default value is 'ignore', but in the detailed docs at https://github.com/expressjs/expressjs.com/blob/gh-pages/_includes/api/en/4x/express.static.md#dotfiles which are linked in the table, it says this:

> With the default value, it will not ignore files in a directory that begins with a dot.

The actual default value seems to be 'allow'.

(The docs for v5 have the same table with the same default value but I don't know whether the behavior changed for v5.)

bjohansebas:
The removed note was incorrect. Thanks @IamLizu for the clarification

closes #1431

### DIFF
--- a/_includes/api/en/4x/express.static.md
+++ b/_includes/api/en/4x/express.static.md
@@ -16,7 +16,7 @@ See also the [example below](#example.of.express.static).
 
 | Property      | Description                                                           |   Type      | Default         |
 |---------------|-----------------------------------------------------------------------|-------------|-----------------|
-| `dotfiles`    | Determines how dotfiles (files or directories that begin with a dot ".") are treated.  <br/><br/>See [dotfiles](#dotfiles) below. | String | [See below](#dotfiles) |
+| `dotfiles`    | Determines how dotfiles (files or directories that begin with a dot ".") are treated.  <br/><br/>See [dotfiles](#dotfiles) below. | String | `undefined` |
 | `etag`        | Enable or disable etag generation <br/><br/>NOTE: `express.static` always sends weak ETags. | Boolean | `true` |
 | `extensions`  | Sets file extension fallbacks: If a file is not found, search for files with the specified extensions and serve the first one found. Example: `['html', 'htm']`.| Mixed | `false` |
 | `fallthrough`  | Let client errors fall-through as unhandled requests, otherwise forward a client error. <br/><br/>See [fallthrough](#fallthrough) below.| Boolean | `true` |
@@ -37,12 +37,7 @@ Possible values for this option are:
 - "allow" - No special treatment for dotfiles.
 - "deny" - Deny a request for a dotfile, respond with `403`, then call `next()`.
 - "ignore" - Act as if the dotfile does not exist, respond with `404`, then call `next()`.
-
-{% capture dotfiles %}
-When this property is not set and so uses the default value, the behavior is similar to 'ignore', with the exception that files in a directory that begins with a dot are **NOT** ignored. This behavior is based on the 1.x version of [serve-static](https://expressjs.com/en/resources/middleware/serve-static.html) middleware.
-{% endcapture %}
-
-{% include admonitions/caution.html content=dotfiles %}
+- `undefined` - Act as ignore, except that files in a directory that begins with a dot are **NOT** ignored.
 
 <h5 id='fallthrough'>fallthrough</h5>
 
--- a/_includes/api/en/4x/express.static.md
+++ b/_includes/api/en/4x/express.static.md
@@ -38,8 +38,6 @@ Possible values for this option are:
 - "deny" - Deny a request for a dotfile, respond with `403`, then call `next()`.
 - "ignore" - Act as if the dotfile does not exist, respond with `404`, then call `next()`.
 
-**NOTE**: With the default value, it will not ignore files in a directory that begins with a dot.
-
 <h5 id='fallthrough'>fallthrough</h5>
 
 When this option is `true`, client errors such as a bad request or a request to a non-existent
--- a/_includes/api/en/4x/express.static.md
+++ b/_includes/api/en/4x/express.static.md
@@ -16,7 +16,7 @@ See also the [example below](#example.of.express.static).
 
 | Property      | Description                                                           |   Type      | Default         |
 |---------------|-----------------------------------------------------------------------|-------------|-----------------|
-| `dotfiles`    | Determines how dotfiles (files or directories that begin with a dot ".") are treated.  <br/><br/>See [dotfiles](#dotfiles) below. | String | "ignore"|
+| `dotfiles`    | Determines how dotfiles (files or directories that begin with a dot ".") are treated.  <br/><br/>See [dotfiles](#dotfiles) below. | String | [See below](#dotfiles) |
 | `etag`        | Enable or disable etag generation <br/><br/>NOTE: `express.static` always sends weak ETags. | Boolean | `true` |
 | `extensions`  | Sets file extension fallbacks: If a file is not found, search for files with the specified extensions and serve the first one found. Example: `['html', 'htm']`.| Mixed | `false` |
 | `fallthrough`  | Let client errors fall-through as unhandled requests, otherwise forward a client error. <br/><br/>See [fallthrough](#fallthrough) below.| Boolean | `true` |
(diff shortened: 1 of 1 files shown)


IamLizu (approved):

UlisesGascon:
cc: @expressjs/express-tc

_includes/api/en/4x/express.static.md
```diff
@@ -38,8 +38,6 @@ Possible values for this option are:
 - "deny" - Deny a request for a dotfile, respond with `403`, then call `next()`.
 - "ignore" - Act as if the dotfile does not exist, respond with `404`, then call `next()`.
 
-**NOTE**: With the default value, it will not ignore files in a directory that begins with a dot.
```
blakeembrey, line 41: It's definitely confusing, but this was changed to ignore in a 2.0 beta: https://github.com/expressjs/serve-static/blob/87c5f092ed7cd23202f699b08c7f407c2b536df0/HISTORY.md?plain=1#L27

However, I believe Express 4.x still uses the 1.x release of `serve-static`: https://github.com/expressjs/express/blob/7e562c6d8daddff4604f8efaaf9db2cf98c6dcff/package.json#L56

I think that means this is still an issue?

wesleytodd: Yeah, this sounds right to me. Should be fixed in v5 though, so if this line is present in the 5.x docs we should remove it there.

bjohansebas:
I find it difficult to use the command you mentioned because I use Docker, and I can't access the container since it fails to start and I can never access it:
```
Error response from daemon: container 6963d14ae8d27b7774a3ccfa1474c03316e7f8dcd5fd2904137aad76e795c883 is not running
```

crandmck:
@bjohansebas Based on the comments from @blakeembrey and @wesleytodd it sounds like this statement is accurate for 4.x, but not for 5.x, and indeed it's not in https://expressjs.com/en/5x/api.html#express.static so nothing needs to be done there (I think).

However, the doc  4.x could be clearer.  I suggest:

1. In the "Default" column for this property, replacing `"ignore"` with something like ("See below" with a link to the note in https://expressjs.com/en/4x/api.html#dotfiles).
1. Clarifying the wording of the note; something like this:
> NOTE: When this property is not set and so uses the default value, the behavior is similar to 'ignore', with the exception that  files in a directory that begins with a dot are NOT ignored.  This behavior is based on the 1.x version of [serve-static](https://expressjs.com/resources/middleware/serve-static.html) middleware.

bjohansebas:
I have made the changes suggested by @crandmck

IamLizu:
Hi @bjohansebas 👋 

Let's look at this code,

```js
const express = require('express');
const app = express();

app.use(
    express.static('public', {
        fallthrough: false,
    })
);

app.get('/', (req, res) => {
    res.send('Hello World');
});

app.listen(3000, () => {
    console.log('Server is running on port 3000');
});
```

Here is the package.json,
```json
{
  "dependencies": {
    "express": "^4.21.0"
  }
}
```

So, as you can see, the setup is with Express v4 and I have disabled `fallthrough` so that you can see the actual error. And I have not provided any value for `dotfiles`, it means it will use the default value `ignore`.

Consider there's a directory called `public` and as you can see, inside you have two files,

- text.txt
- .text.txt

Start the server with `node index.js` and try accessing the contents respectively. You will see an error while trying to access `.text.txt`,

```console
NotFoundError: Not Found
    at createHttpError (D:\work\personal\express-tests\dotfiles\node_modules\send\index.js:978:12)
    at SendStream.error (D:\work\personal\express-tests\dotfiles\node_modules\send\index.js:270:31)
    at SendStream.pipe (D:\work\personal\express-tests\dotfiles\node_modules\send\index.js:579:14)
    at serveStatic (D:\work\personal\express-tests\dotfiles\node_modules\serve-static\index.js:125:12)
    at Layer.handle [as handle_request] (D:\work\personal\express-tests\dotfiles\node_modules\express\lib\router\layer.js:95:5)
    at trim_prefix (D:\work\personal\express-tests\dotfiles\node_modules\express\lib\router\index.js:328:13)
    at D:\work\personal\express-tests\dotfiles\node_modules\express\lib\router\index.js:286:9
    at Function.process_params (D:\work\personal\express-tests\dotfiles\node_modules\express\lib\router\index.js:346:12)
    at next (D:\work\personal\express-tests\dotfiles\node_modules\express\lib\router\index.js:280:10)
    at expressInit (D:\work\personal\express-tests\dotfiles\node_modules\express\lib\middleware\init.js:40:5)
```

As we know `ignore` means it will throw 404 and then call `next()`. So it appears the default is `ignore` which i_ndeed ignores the dot files._

I hope it makes sense 🎉 

@blakeembrey @wesleytodd Please help me understand if I am wrong.

blakeembrey:
Looks like @IamLizu might be right, I reviewed the 1.x and 2.x code and I don't think it materially changed this behavior. The `send` and `serve-static` docs/changelog are incorrect. `'ignore'` appears to be the default in both releases _and_ it ignores both directories/files making this line incorrect too:

> When this property is not set and so uses the default value, the behavior is similar to 'ignore', with the exception that files in a directory that begins with a dot are **NOT** ignored.

It should be fixed in both the respective package READMEs too.

_includes/api/en/4x/express.static.md
```diff
@@ -16,7 +16,7 @@ See also the [example below](#example.of.express.static).
 
 | Property      | Description                                                           |   Type      | Default         |
 |---------------|-----------------------------------------------------------------------|-------------|-----------------|
-| `dotfiles`    | Determines how dotfiles (files or directories that begin with a dot ".") are treated.  <br/><br/>See [dotfiles](#dotfiles) below. | String | "ignore"|
+| `dotfiles`    | Determines how dotfiles (files or directories that begin with a dot ".") are treated.  <br/><br/>See [dotfiles](#dotfiles) below. | String | [See below](#dotfiles) |
```
IamLizu: Default  value is `ignore`

_includes/api/en/4x/express.static.md
```diff
@@ -30,15 +30,19 @@ See also the [example below](#example.of.express.static).
...
 
 Possible values for this option are:
 
 - "allow" - No special treatment for dotfiles.
 - "deny" - Deny a request for a dotfile, respond with `403`, then call `next()`.
 - "ignore" - Act as if the dotfile does not exist, respond with `404`, then call `next()`.
 
-**NOTE**: With the default value, it will not ignore files in a directory that begins with a dot.
+{% capture dotfiles %}
+When this property is not set and so uses the default value, the behavior is similar to 'ignore', with the exception that files in a directory that begins with a dot are **NOT** ignored. This behavior is based on the 1.x version of [serve-static](https://expressjs.com/en/resources/middleware/serve-static.html) middleware.
```
IamLizu: Actually it does ignore the dot files even on 1.x version.

IamLizu (changes requested):

bjohansebas:
Hi @IamLizu, I have also been testing this behavior. You can have a directory called `.static` inside the `public` folder, and the files will be obtained as long as they don't start with a `.`. However, if you set the ignore, the `.static` folder will no longer work and will throw an error.

In other words, the default behavior does not ignore folders that start with a `.`.

You can clone this repo and give it a try if you want! https://github.com/bjohansebas/test-dotfiles-express

_includes/api/en/4x/express.static.md
blakeembrey: Does it ignore the directory in 1.x? E.g. `.foo/test.js`?

IamLizu:
Ah I see the confusion now. 

I only tested with files but not with directory. Its clear now. 

Thanks @bjohansebas 🎉 

Edit: I am hiding my earlier comment as "off-topic" to avoid confusion.

IamLizu (approved):
LGTM 🎉

blakeembrey:
OK, I see what's confusing. `serve-static` doesn't do anything with `dotfiles`. The option is passed through to `send`. In 1.x, we're using 0.x of send. In 2.x, we're using 1.x.

This is the commit that fixed the default: https://github.com/pillarjs/send/commit/85dfcd7bc86b5661f92582ba86ce3fbfa9800dce.

It was tricky to find this in the history because it come _before_ the version 0.18 release, and because I couldn't find any other commit that sounded like it changed this I assumed incorrectly that our docs were wrong. I think instead our merge commit is a bit of a mess and could have been squashed to make this clearer.

TL;DR the new doc changes look right.

WDYT about specifying that default is `undefined`, and how `undefined` works differently to `ignore`?

blakeembrey (approved):

blakeembrey (approved):

IamLizu (approved):

crandmck (approved):

Issue #1479, opened 2024-04-02
Review and Update v5.x Documentation against v4.x


UlisesGascon:
In our recent TC meeting, we discussed the necessity to review the `v5.x` documentation due to discrepancies with changes made in `v4.x` that were not ported over.

Additionally, we need to implement a GitHub Action that will automatically add a comment to any PR that modifies files within the `../v4/..` folder, reminding contributors to review if these changes also affect `v5.x`. (see: #1514 )

crandmck:
Thanks for captuing this @UlisesGascon

UlisesGascon:
I will start to work on this

UlisesGascon:
I just invested some time to check what is missing potentially in [the migration guide](https://expressjs.com/en/guide/migrating-5.html) and [the 5.x API](https://expressjs.com/en/5x/api.html) based on [the changelog](https://github.com/expressjs/express/blob/5.x/History.md)


### Legend
- 🍿 it is ignored as it is not relevant to Express 5 Docs or migration
- ✅ it was confirmed that is present in Docs and/or Migration
- 🚩 it will require some work potentially


### Changelog digestion


5.0.0-beta.3 / 2024-03-25
----

This incorporates all changes after 4.19.1 up to 4.19.2. **🍿 (Ignored)**

5.0.0-beta.2 / 2024-03-20
----

This incorporates all changes after 4.17.2 up to 4.19.1. **🍿 (Ignored)**

5.0.0-beta.1 / 2022-02-14
----

This is the first Express 5.0 beta release, based off 4.17.2 and includes
changes from 5.0.0-alpha.8.

  * change:
    - Default "query parser" setting to `'simple'` **🚩 (Requires reference in the Migration Guide?)**
    - Requires Node.js 4+ **✅ (Was included)**
    - Use `mime-types` for file to content type mapping **🍿 (Ignored)**
  * deps: array-flatten@3.0.0 **🍿 (Ignored)**
  * deps: body-parser@2.0.0-beta.1 
    - `req.body` is no longer always initialized to `{}` **🚩 (Requires reference in the Migration Guide?)**
    - `urlencoded` parser now defaults `extended` to `false` **🚩 (Requires reference in the Migration Guide?)**
    - Use `on-finished` to determine when body read **🍿 (Ignored)**
  * deps: router@2.0.0-beta.1 
    - Add new `?`, `*`, and `+` parameter modifiers **✅ (Was included)**
[...]

PR #1602, opened 2024-09-10
docs: change current and next version for the docs

IamLizu:
As we have [Express 5 stable released](https://github.com/expressjs/express/releases/tag/v5.0.0),

- changed the next version to _5.0.0_
- changed current version to _4.20.0_ as of current latest

### DIFF
--- a/_data/express.yml
+++ b/_data/express.yml
@@ -1,2 +1,2 @@
-current_version: "4.20.0"
+current_version: "4.21.0"
 next_version: "5.0.0"
--- a/_data/express.yml
+++ b/_data/express.yml
@@ -1,2 +1,2 @@
-current_version: "4.19.2"
-next_version: "5.0.0-beta.1"
+current_version: "4.20.0"
+next_version: "5.0.0"


IamLizu:
Perhaps we have a lot to do in terms of website doc change as _5.0.0_ is not in beta anymore.

We should,

- [ ] Prepare a list of files that has to be changed in accordance with the _5.0.0_ release.
- [ ] Create separate issues from the list so that people from the community can take ownership and contribute.

cc: @expressjs/docs-wg

UlisesGascon (approved):

_data/express.yml
```diff
@@ -1,2 +1,2 @@
-current_version: "4.19.2"
-next_version: "5.0.0-beta.1"
\ No newline at end of file
+current_version: "4.20.0"
```
bjohansebas: We should promote version 5 instead of version 4
```suggestion
current_version: "5.0.0"
```

IamLizu: We could, but our current version is not 5 yet. We are still calling it next until we polish our documentation or other things first. 

In short, `4.20.0` will be _lastest_ and `5.0.0` will be _next_ for a while.

bjohansebas: Sure, I have more context now, so it seems good to me.

bjohansebas:
@IamLizu, there’s already an issue that provides quite a good overview of all the changes that need to be made to the API and the documentation. Maybe it’s missing some minor changes introduced after beta 3, but I think it’s pretty good for that. see #1479

_data/express.yml
bjohansebas:
```suggestion
current_version: "4.21.0"
```

IamLizu:
Hi @crandmck 

I think its good to go now, requesting your feedback.

bjohansebas (approved):

carlosstenzel (approved):

crandmck (approved):

PR #1606, opened 2024-09-11
fix rules issues for update_external_docs workflow

bjohansebas:
Changes are made to run the workflow correctly and prevent this workflow from running in a forked repository.

## Changes:
- Allow the workflow to run manually
- Now the workflow will only run in the main repository (expressjs/expressjs.com).
- The tool is updated to use an external action allowed by the organization.
- Automatically add the review for the @expressjs/docs-wg team.
- Update the checkout action to its latest version

### DIFF
--- a/.github/workflows/update-external-docs.yml
+++ b/.github/workflows/update-external-docs.yml
@@ -1,7 +1,8 @@
 name: Update External Docs
 on:
+    workflow_dispatch:
     schedule:
-    - cron:  '0 8 * * 1'
+        - cron:  '0 8 * * 1'
 
 permissions:
     pull-requests: write
--- a/.github/workflows/update-external-docs.yml
+++ b/.github/workflows/update-external-docs.yml
@@ -12,7 +12,7 @@ jobs:
         runs-on: ubuntu-latest
         if: github.repository_owner == 'expressjs'
         steps:
-            - uses: actions/checkout@v3
+            - uses: actions/checkout@v4
             - name: Run scripts
               run: bash ./get-contributing.sh && bash ./get-readmes.sh
             - name: Create Pull Request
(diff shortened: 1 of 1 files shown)


bjohansebas:
ping

.github/workflows/update-external-docs.yml
```diff
@@ -11,20 +11,20 @@ permissions:
...
         runs-on: ubuntu-latest
+        if: github.repository_owner == 'expressjs'
         steps:
-            - uses: actions/checkout@v3
+            - uses: actions/checkout@v4
             - name: Run scripts
               run: bash ./get-contributing.sh && bash ./get-readmes.sh
             - name: Create Pull Request
-              uses: peter-evans/create-pull-request@v5
+              uses: gr2m/create-or-update-pull-request-action@v1
```
UlisesGascon, line 20: `peter-evans/create-pull-request` also allows to do team_reviewers (https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#action-inputs). Does `gr2m/create-or-update-pull-request-action` provides any additional feature or improvement? I am just curious.

bjohansebas: This tool is being changed not because it has other functionalities, but because when executing the workflow, the following error occurred

```
peter-evans/create-pull-request@v5 is not allowed to be used in expressjs/expressjs.com. Actions in this workflow must be: within a repository owned by expressjs, created by GitHub, verified in the GitHub Marketplace, or matching the following: pkgjs/*, gr2m/create-or-update-pull-request-action@*, ulisesgascon/openssf-scorecard-monitor@*.
```

UlisesGascon: Oh! That makes sense... those are rules that we define at the org level :+1: 

Thanks for the clarification

UlisesGascon (approved):

PR #1615, opened 2024-09-13
feat - add support and blog in menu

carlosstenzel:
Add in menu
 - page support
 - page blog


juliogarciape (commented):
I think it's good to add a new tab, but if there is no translation for each content on the blog or support page, I think it's unnecessary.

carlosstenzel:
@bjohansebas 
what do you think ?

bjohansebas:
I think this would help maintain consistency between pages even if the content is not translated

bjohansebas:
I would prefer to launch this PR when #1568 is approved, as adding this PR may break mobile responsiveness.

bjohansebas:
hi @carlosstenzel, Can you add the support page to each of the languages, please?

bjohansebas (approved):
LGTM, thanks @carlosstenzel

PR #1572, opened 2024-08-16
improve open graph metadata

bjohansebas:
### Changes:
- Update the OpenGraph image (If you have any ideas to improve the image, you can check the following issue #1564)
- Improve the metadata by adding the description field (Gradually, we can add these descriptions so that Google or other search engines don’t impose them).

PR #1620, opened 2024-09-15
Suggest description

bjohansebas:
Since #1572 was merged into the main branch, the page now supports editing the description metadata, which is important for the page's SEO.

edit: I'm going to open an issue in the future to report that the description needs to be added to the pages

### DIFF
--- a/en/blog/write-post.md
+++ b/en/blog/write-post.md
@@ -30,6 +30,7 @@ Create a new file in the `_posts` directory named using following the format: `Y
     ---
     title: <your-title>
     sub_title: <your-optional-sub-title>
+    description: <description-of-your-post>
     tags: <white-space-separated-topics>
     author: <your-name>
     ---


chrisdel101:
We used to have `description`, at least in development, but it was removed. I'm all for having it, but I do recall it was specifically removed, and not forgotten about. This was for a in template type of thing.
Is this just the meta? Or a template addition too?
What's the point of having it when the page does not display it?

bjohansebas:
Although the page doesn't display it directly, search engines do use it and show it to users when they search for Express. Also, now when a link is shared on any social media, the description will appear below the title. It may have lesser importance, but it can be improved to make the page a `first-class citizen`.

This would be a change that adds just one line, and the page infrastructure already supports it, so it wouldn’t hurt to add it. Making this change across all pages would be a long-term task (and of low relevance), but it could be useful for those who want to get started in the project by making these changes.

chrisdel101:
LGTM.

chrisdel101 (approved):

UlisesGascon (approved):

PR #1626, opened 2024-09-21
Update route guide

bjohansebas:


carlosstenzel (approved):

en/guide/routing.md
```diff
@@ -68,19 +68,21 @@ app.all('/secret', (req, res, next) => {
...
 Route paths, in combination with a request method, define the endpoints at which requests can be made. Route paths can be strings, string patterns, or regular expressions.
 
-The characters `?`, `+`, `*`, and `()` are subsets of their regular expression counterparts. The hyphen (`-`) and the dot (`.`) are interpreted literally by string-based paths.
+{% capture caution-character %} In express 5, the characters `?`, `+`, `*`, `[]`, and `()` are handled differently than in version 4, please review the [migration guide](/{{ page.lang }}/guide/migrating-5.html#path-syntax) for more information.{% endcapture %}
 
-If you need to use the dollar character (`$`) in a path string, enclose it escaped within `([` and `])`. For example, the path string for requests at "`/data/$book`", would be "`/data/([\$])book`".
+{% include admonitions/caution.html content=caution-character %}
 
-<div class="doc-box doc-info" markdown="1">
+If you need to use the dollar character (`$`) in a path string, escape it with `\`. For example, the path string for requests at `/data/$book`, would be `/data/\$book`.
```
blakeembrey: You can just write a `$` in Express 5, there's no need to escape. The only characters needed to escape are values used by `path-to-regexp` specifically, and they're here: https://github.com/pillarjs/path-to-regexp/blob/d63f44bc54f2c6fe7d35dc77f8515622006bf7cb/src/index.ts#L105

en/guide/routing.md
```diff
@@ -282,20 +293,20 @@ The methods on the response object (`res`) in the following table can send a res
...
+| [res.end()](/{{ page.lang }}/5x/api.html#res.end)               | End the response process.                                                             |
+| [res.json()](/{{ page.lang }}/5x/api.html#res.json)             | Send a JSON response.                                                                 |
+| [res.jsonp()](/{{ page.lang }}/5x/api.html#res.jsonp)           | Send a JSON response with JSONP support.                                              |
+| [res.redirect()](/{{ page.lang }}/5x/api.html#res.redirect)     | Redirect a request.                                                                   |
+| [res.render()](/{{ page.lang }}/5x/api.html#res.render)         | Render a view template.                                                               |
+| [res.send()](/{{ page.lang }}/5x/api.html#res.send)             | Send a response of various types.                                                     |
+  | [res.sendFile()](/{{ page.lang }}/5x/api.html#res.sendFile)     | Send a file as an octet stream.                                                       |
```
blakeembrey:
```suggestion
| [res.sendFile()](/{{ page.lang }}/5x/api.html#res.sendFile)     | Send a file as an octet stream.                                                       |
```

en/guide/routing.md
bjohansebas: Thanks, I didn't know that this had changed. It might be good to inform about this change in the migration guide and in the README of path-to-regex

en/guide/routing.md
```diff
@@ -68,19 +68,24 @@ app.all('/secret', (req, res, next) => {
...
+{% include admonitions/caution.html content=caution-character %}
 
-<div class="doc-box doc-info" markdown="1">
+{% capture note-dollar-character %}In express 5, you don't need to escape the dollar character (`$`) with `\`.
+{% endcapture %}
+
+{% include admonitions/note.html content=note-dollar-character %}
+
+{% capture note-path-to-regexp %}
   Express uses [path-to-regexp](https://www.npmjs.com/package/path-to-regexp) for matching the route paths; see the path-to-regexp documentation for all the possibilities in defining route paths. [Express Route Tester](http://forbeslindesay.github.io/express-route-tester/) is a handy tool for testing basic Express routes, although it does not support pattern matching.
```
blakeembrey: Although not part of this edit, I'd probably remove the link to the route tester or open an issue to try and get it updated to the version used by Express 5. We could also fork it and update it as part of the website potentially.

en/guide/routing.md
```diff
@@ -68,19 +68,24 @@ app.all('/secret', (req, res, next) => {
 
 Route paths, in combination with a request method, define the endpoints at which requests can be made. Route paths can be strings, string patterns, or regular expressions.
```
blakeembrey, line 70:
```suggestion
Route paths, in combination with a request method, define the endpoints at which requests can be made. Route paths can be strings or regular expressions.
```

en/guide/routing.md
```diff
@@ -68,19 +68,24 @@ app.all('/secret', (req, res, next) => {
...
 
-The characters `?`, `+`, `*`, and `()` are subsets of their regular expression counterparts. The hyphen (`-`) and the dot (`.`) are interpreted literally by string-based paths.
+{% capture caution-character %} In express 5, the characters `?`, `+`, `*`, `[]`, and `()` are handled differently than in version 4, please review the [migration guide](/{{ page.lang }}/guide/migrating-5.html#path-syntax) for more information.{% endcapture %}
 
-If you need to use the dollar character (`$`) in a path string, enclose it escaped within `([` and `])`. For example, the path string for requests at "`/data/$book`", would be "`/data/([\$])book`".
+{% include admonitions/caution.html content=caution-character %}
 
-<div class="doc-box doc-info" markdown="1">
+{% capture note-dollar-character %}In express 5, you don't need to escape the dollar character (`$`) with `\`.
```
blakeembrey: May be better worded as "In Express 4, regular expression characters such as `$` need to be escaped with a `\`."? Not strongly opinionated, but may be easier to explain it away as Express 4 everything could be a regex and Express 5 there are no regex supported (which is noted elsewhere already).

blakeembrey (approved):

en/guide/routing.md
blakeembrey: It is under:

> Regexp characters are not supported

And this has been true since at least 1.x of `path-to-regexp` IIRC. I haven't had an issue with anyone misunderstanding how `path-to-regexp` works since then because it hasn't been a literal "path is a regex" for a long time, so it's easier to avoid confusing notes like not needing to escape certain characters. Easier to just stick with documenting the DSL that is used.

blakeembrey: Of course, that note could be clarified, since maybe it isn't clear that `^`, `$`, etc are regex characters?

en/guide/routing.md
bjohansebas: I was thinking about removing that part when I edited this file for the first time. Although it doesn’t work in Express 5, string patterns are present in version 4

blakeembrey: Makes sense, I think "string pattern" is confusing... they're both strings. It's just Express 4 and 5 support strings of different syntax within the strings. I was thinking it'd be clearer to just note that the strings are different DSLs, and Express 4 happened to support regex characters, and if you wanted those characters literally they needed to be escaped.

blakeembrey: It also depends on whether this doc is v5 first or v4 first I guess, but since the doc doesn't touch much on things that changed between the versions (i.e. `*` -> `*name`, etc) it might be easier to have it be things supported by each overall with caveats afterward for 4 vs 5 specifics?

en/guide/routing.md
bjohansebas: It's true, it's not updated. I will remove it in the meantime. I would opt to fork the repository so we're not dependent on the author. Maybe we can ask him to transfer it to the organization and we can maintain it

en/guide/routing.md
bjohansebas: I like your suggestion more

en/guide/routing.md
bjohansebas: Although a note was added below stating that string patterns are not supported in version 5, so I'm not sure if it would be good to make the change you suggested.

If you want to make any further changes that you think are important to improve the router guide, it's enabled for maintainers to make edits. You know more about how the router works

bjohansebas: cc: @expressjs/docs-wg

juliogarciape (approved):
LGTM

UlisesGascon (approved):

UlisesGascon:
Once the discussions are solve, I think that we can land it :)

wesleytodd (approved):

en/guide/routing.md
bjohansebas: @expressjs/docs-wg

bjohansebas:
I've updated the playground on my end to bring it back, as I think it's something good to have on the page. If you'd like, it can be moved to this organization.

Also, this hasn't been moved in a long time, so if there are no objections, I'll merge it by the end of next week.

cc: @expressjs/express-tc @expressjs/docs-wg

PR #1630, opened 2024-09-24
fix: Erase a sentence about discussion

LuiGeeDev:
In English version, the sentence 'You can read the discussion that led to the deprecation of the `app.del` function here.' is deleted 9 years ago([commit](https://github.com/expressjs/expressjs.com/commit/2b103e4be2278fc76f46ac9d9e60cc4281f4fe19#diff-380ce9fbefe3250db0463611f085b37c748e503a54e6576c9a7bbc20a27adadeL71)), but it still exists in a Korean migration guide. So this commit erases it.


UlisesGascon (approved):
Thanks for the explanation and the context in the PR @LuiGeeDev! :heart:

bjohansebas (approved):

PR #1631, opened 2024-09-28
remove multipart example

bjohansebas:
Deleted in https://github.com/expressjs/express/pull/5195


bjohansebas:
I just remembered that this file is brought in by the script

PR #1634, opened 2024-09-30
Fix: remove non-existing link sentence

wilyJ80:
One of the sentences in the advanced security section had a non-existing link I couldn't find. If someone does find, they can inform me below, and edits are welcome. Otherwise, this pull request contains changes to remove the sentence.

Also, the previous sentence was incomplete. This was fixed, although changes are welcome: the semantics of the original sentence were vague. Hope this makes the docs even more clear and professional.

### DIFF
--- a/en/advanced/best-practice-security.md
+++ b/en/advanced/best-practice-security.md
@@ -51,8 +51,7 @@ Also, a handy tool to get a free TLS certificate is [Let's Encrypt](https://lets
 ## Do not trust user input
 
 For web applications, one of the most critical security requirements is proper user input validation and handling. This comes in many forms and we will not cover all of them here.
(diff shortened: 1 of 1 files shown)


crandmck:
Good catch!

bjohansebas:
@crandmck below that paragraph, there's a title 'Prevent open redirects,' wouldn't that be an example?

crandmck:
> below that paragraph, there's a title 'Prevent open redirects,' wouldn't that be an example?

Yes, for sure.  

But the sentence removed, "Here are a few examples of validating user input specifically using a few `express` apis." was inaccurate, since there's only _one_ example, and also the formatting/spelling was off.  

It seems obvious that the next section is an instance of validating user input, but if you like you could add a transition, like this: "Ultimately, the responsibility for validating and correctly handling the types of user input your application accepts is yours; for example, applications shouldn't allow open redirects."

Issue #1504, opened 2024-05-02
Add GH action for routine version updates


crandmck:
It would be awesome if we could automate docs for routine minor releases. I'm imagining a GH action that would create a PR similar to https://github.com/expressjs/expressjs.com/pull/1492 for any new release. 

At a minimum, each release requires:
- Updating the release change log https://github.com/expressjs/expressjs.com/blob/gh-pages/en/changelog/4x.md with changes for the releases shown in https://github.com/expressjs/express/releases
- Updating the version number in https://github.com/expressjs/expressjs.com/blob/gh-pages/_data/express.yml (used in a couple of places in the docs) with the version number of the release.

Of course, larger releases might require further changes, but I believe this is sufficient for minor versions.

bjohansebas:
A workflow could be created to detect changes in the history of the Express repository. However, to avoid pulling information about updated dependencies (or irrelevant information), two separate histories could be managed: one for the website and the existing one.

@crandmck I could give it a try if that sounds good to you

PR #1643, opened 2024-10-05
Automate the version change of Express

bjohansebas:
This script automates the process of changing the version number in the documentation, see #1504.

To ensure the workflow functions correctly, please review #1606

### DIFF
--- a/get-express-version.mjs
+++ b/get-express-version.mjs
@@ -1,0 +1,22 @@
+import { writeFileSync } from 'node:fs'
+import { readFile } from 'node:fs/promises'
+import path from 'node:path'
+
+const NPMURL = 'https://registry.npmjs.org/express'
+
+const response = await fetch(NPMURL).then((res) => res.json())
+
+const { next, latest } = response['dist-tags']
+
+const contentFile = await readFile(
+  path.resolve(path.join('_data', 'express.yml')),
+  'utf8'
+)
+  .then((content) =>
+    content.replace(/current_version: ".*"/, `current_version: "${latest}"`)
+  )
+  .then((content) =>
+    content.replace(/next_version: ".*"/, `next_version: "${next}"`)
+  )
+
+writeFileSync('_data/express.yml', contentFile)
--- a/get-express-version.mjs
+++ b/get-express-version.mjs
@@ -1,22 +1,20 @@
-import { writeFileSync } from 'node:fs'
-import { readFile } from 'node:fs/promises'
+import { readFile, writeFile } from 'node:fs/promises'
 import path from 'node:path'
 
 const NPMURL = 'https://registry.npmjs.org/express'
 
-const response = await fetch(NPMURL).then((res) => res.json())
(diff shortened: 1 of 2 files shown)


UlisesGascon (approved):

get-express-version.mjs
```diff
@@ -0,0 +1,22 @@
...
+import { readFile } from 'node:fs/promises'
+import path from 'node:path'
+
+const NPMURL = 'https://registry.npmjs.org/express'
+
+const response = await fetch(NPMURL).then((res) => res.json())
+
+const { next, latest } = response['dist-tags']
+
+const contentFile = await readFile(
```
IamLizu: Should we add error handling here?

get-express-version.mjs
```diff
@@ -0,0 +1,22 @@
...
+  path.resolve(path.join('_data', 'express.yml')),
+  'utf8'
+)
+  .then((content) =>
```
IamLizu: It's better to use `await` consistently instead of mixing with `.then()`.

get-express-version.mjs
```diff
@@ -0,0 +1,22 @@
+import { writeFileSync } from 'node:fs'
+import { readFile } from 'node:fs/promises'
+import path from 'node:path'
+
+const NPMURL = 'https://registry.npmjs.org/express'
+
+const response = await fetch(NPMURL).then((res) => res.json())
```
IamLizu: It's better to use `await` consistently instead of mixing with `.then()`.

IamLizu (changes requested):
Hey @bjohansebas 👋 

Great job on automating this 💯, I am generally 👍 on this. However, would you be open to something like this? Perhaps we should make consistent use of `await`.

I have also added error-handling just to _feel safe_.
```mjs
import { readFile, writeFile } from 'node:fs/promises';
import path from 'node:path';

const NPMURL = 'https://registry.npmjs.org/express'

const response = await (await fetch(NPMURL)).json();
const { next, latest } = response['dist-tags']

try {
  const filePath = path.resolve(path.join('_data', 'express.yml'));
  let content = await readFile(filePath, 'utf8');
  
  content = content.replace(/current_version: ".*"/, `current_version: "${latest}"`);
  content = content.replace(/next_version: ".*"/, `next_version: "${next}"`);
  
  await writeFile(filePath, content, 'utf8');
} catch (error) {
  console.error('Error updating versions in _data/express.yml:', error);
}
```

juliogarciape (approved):

bjohansebas:
Thanks @IamLizu for the suggestions.

get-express-version.mjs
bjohansebas: done, thanks

get-express-version.mjs
bjohansebas: done, thanks

get-express-version.mjs
bjohansebas: done, thanks

carlosstenzel (approved):

PR #1656, opened 2024-10-18
General changelog

bjohansebas:
The changelog is generalized so that it is not only for version 4


bjohansebas:
@UlisesGascon @wesleytodd, what is the possibility of managing two histories in the Express repository, meaning:

- HISTORY.md: As it has been functioning
- HISTORY-REMOTE.md: The changelog adapted for the website

This would help keep the page updated with the latest changes in the library; we would just need to create the script to fetch HISTORY-REMOTE.md, now that external documentation can be automated.

carlosstenzel (approved):

UlisesGascon:
@bjohansebas this might we relevant: https://github.com/expressjs/discussions/issues/293

bjohansebas:
@UlisesGascon While the future of history.md is being decided, we could release this to keep the page updated

UlisesGascon (approved):

wesleytodd:
I have not been able to keep up with all this work, but I am not sure if I agree this should have landed. It is in the 4.x directory, and I think the idea from the past was that we would create a new one in the 5.x directory for the new major. I dont think we should back this out, but I do think we should consider if this was their right direction to take and revisit it after we do that.

PR #1658, opened 2024-10-21
Update community page

bjohansebas:


UlisesGascon (commented):
I think that we follow alphabetical order if I am not mistaking here :thinking:

bjohansebas:
I don't think so, but if you want it to be in alphabetical order, should it be by username or by first name?

blakeembrey:
> if you want it to be in alphabetical order, should it be by username or by first name?

IMO probably username? That way you can just "sort lines" in your editor.

bjohansebas:
ping

PR #1667, opened 2024-10-25
Add multiple social media links to the footer

### LINKED ISSUE #1123: Express Youtube Channel and Twitter Visibility
We should link somewhere more visible on the site to the [Express Youtube channel](https://www.youtube.com/channel/UCYjxjAeH6TRik9Iwy5nXw7g) and the [Express Twitter account](https://twitter.com/useexpressjs). 

There is a link to the youtube channel on the [community page](https://expressjs.com/en/resources/community.html) already, and a link to the twitter account can join it somewhere on there.

We discussed as an aside off the air at the last TC meeting making them a bit more visible, or possibly adding social badges.

It would be really cool eventually to do something like put a notice on the Express site, or a banner on the docs, when a live TC meeting is scheduled to start or is live. That's kind of a shower thought, but if we were set up to handle the amount of participation that could bring, it would be a really sweet way to get the word out.

bjohansebas:
Express has several social media channels that can be promoted from the website.

cc: @expressjs/express-tc 

closes: #1123


carlosstenzel (approved):
LGTM

_includes/footer/footer-en.html
```diff
@@ -23,9 +24,35 @@
...
+            <span class="hidden-dark">
+              {% include icons/x-light.svg %}
+            </span>
+            <span class="hidden-light">
+              {% include icons/x-dark.svg %}
+            </span>
+          </a>
+        </div>
+        <div>
+          <a href="https://openjs-foundation.slack.com/archives/C02QB1731FH" aria-label="Go to the repository on Slack">{% include icons/slack.svg %}</a>
```
UlisesGascon, line 51:
```suggestion
          <a href="https://slack-invite.openjsf.org/" aria-label="Go to the repository on Slack">{% include icons/slack.svg %}</a>
```

Maybe is better to provide a link to the slack invitation rather than the channel? Ideally we can have a small section in the web where we explain how to join slack and what channels to monitor. :thinking:

UlisesGascon (approved):
I love this idea! Thanks @bjohansebas for taking the initiative here :heart:

_includes/footer/footer-en.html
bjohansebas: I think we'd prefer them to join the specific Slack channel instead of just Slack itself, as Slack is more general and includes all OpenJSF projects. The idea here is to strengthen Express

bjohansebas: @expressjs/docs-wg, what do you all think about this?

carlosstenzel: I agree with @bjohansebas , we should direct to the channel for the express

Issue #1668, opened 2024-10-25
Centralizing the site


carlosstenzel:
What do you think about centralizing the site ?

today
![Captura de Tela 2024-10-24 às 21 40 49](https://github.com/user-attachments/assets/958eeaaa-30a5-435b-840b-23bb12da0f00)

centralizing

![Captura de Tela 2024-10-24 às 21 55 38](https://github.com/user-attachments/assets/1bc07780-f870-4a1a-8aef-e523191853cd)

bjohansebas:
That's the direction I want to take for the page design. It's not necessary to center the navigation bar if you want to open a PR

bjohansebas:
fixed by #1670

PR #1670, opened 2024-10-26
feat: improve home layout

carlosstenzel:
Align content

by https://github.com/expressjs/expressjs.com/issues/1668


bjohansebas (commented):
Please fix the mobile responsiveness

carlosstenzel:
> Please fix the mobile responsiveness

@bjohansebas I made some responsiveness adjustments for mobile

chrisdel101:
I think the original issue you put up is good idea, but this is what I get now at this width.  
Screenshot 2024-10-27 at 6 24 37 PM
Obv the nav squish has nothing to do with your changes

Looking at is a bit more the page setup is pretty weird - this is from before our time. 
In my opinion the main page content should have a wrapper where the max-width is set, so like `home-conent`, but without the nav content in there, `announcments` and `intro` should go inside wrapper. Then you can just `margin:auto` or flex box the content inside the wrapper.

Or however you want to do it.

carlosstenzel:
> I think the original issue you put up is good idea, but this is what I get now at this width.
> Obv the nav squish has nothing to do with your changes

I'm not looking at the menu right now at the request of the @bjohansebas 

@chrisdel101  I will analyze the other screen sizes, tks

bjohansebas:
it could be fixed by adding padding to the elements you changed, and please remove the changes to the footer, as this was already done in another PR.

bjohansebas:
Hi @carlosstenzel, I’ve made some changes to fix some issues with certain screen measurements, and I’ve used grid for the boxes. This improves the responsive design. 

Feel free to change it if you don’t think it’s the best approach.

bjohansebas (approved):

carlosstenzel:
@bjohansebas 

Perfect, thank you.

css/style.css
```diff
@@ -425,6 +423,10 @@ html[xmlns] .clearfix {
 /* boxes */
 
 #boxes {
+  display: grid;
```
chrisdel101, line 426: The middleware box seems a bit out of place to me. Maybe we could center that if this looks good? It will need a media query though since it does not work on mobile.
`grid-column-start: 2;`
Screenshot 2024-11-20 at 10 43 31 AM

chrisdel101 (changes requested):

css/style.css
chrisdel101: The grid looks nice. Also I like:

- The page text container does not flow to the page edges on large size
- the links are centred, esp on mobile.

bjohansebas: I am looking for a way to implement it

bjohansebas: What if we make it two columns? So it doesn't stay floating in the middleware box.

chrisdel101: Yeah could work. Just don't let it expand from 2 to 3?

bjohansebas: Maybe now?

chrisdel101: Nice. Yes now

chrisdel101 (approved):

PR #1675, opened 2024-10-29
remove open collective of footer

bjohansebas:
This removes the open collective from the footer while we decide when to promote it

Relate: https://github.com/expressjs/express/pull/6064#issuecomment-2444511668


wesleytodd:
So I believe that we have gotten 👍 from most folks now that we are good to move forward with promoting the OC. We have a small proposal for a "policy" to make sure folks don't feel like they are giving money to us without any plan at all, I will check in to see if we can post that somewhere publicly today. And if so, then we can close this.

carlosstenzel (approved):

PR #1690, opened 2024-11-21
add 404 page

bjohansebas:
The 404 page is added to stop using the default GitHub page when there is a 404 error.

Before:
image

after:

image

### DIFF
--- a/404.md
+++ b/404.md
@@ -1,0 +1,12 @@
+---
+layout: 404
+title: Page not found
+description: The page you requested could not be found.
+lang: en
+---
+
+# Not Found
+
+We couldn’t find the page you’re looking for. Check the address or [head back home](/).
+
+If this is a mistake, [let us know](https://github.com/expressjs/expressjs.com/issues/new/), and we will try to fix it!
--- a/404.md
+++ b/404.md
@@ -1,11 +1,11 @@
 ---
 layout: 404
 title: Page not found
-description: The page you requested could not be found.
+description: We couldn’t find the page you’re looking for.
 lang: en
 ---
 
-# Not Found
+# {{ page.title }}
 
 We couldn’t find the page you’re looking for. Check the address or [head back home](/).
 
--- a/css/style.css
+++ b/css/style.css
@@ -883,7 +883,6 @@ h2 a {
   padding: 153px 32px 7%;
   text-align: center;
   display: flex;
-  align-items: center;
   justify-content: center;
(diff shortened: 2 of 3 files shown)


carlosstenzel (approved):

404.md
```diff
@@ -0,0 +1,12 @@
+---
+layout: 404
+title: Page not found
+description: The page you requested could not be found.
+lang: en
+---
+
+# Not Found
```
chrisdel101: 1. Could you use the title here?
2. I'd like the wording "Page Not Found" better than "Not Found."
3. Also use front matter liquid unless there is a reason not to.

`# {page.title} `

`{page.description}`

css/style.css
```diff
@@ -878,6 +878,20 @@ h2 a {
   border-color: #333;
 }
 
+.content-404 {
+  height: 70vh;
+  padding: 153px 32px 7%;
+  text-align: center;
+  display: flex;
+  align-items: center;
```
chrisdel101: Maybe just make sure this rule does anything. Seems to center without it. 
` align-items: center;`
Optional remove at your discretion.

chrisdel101 (changes requested):

404.md
bjohansebas: sure

css/style.css
bjohansebas: good catch

404.md
bjohansebas: I can't use liquid for the description because we are using link, so i'm going to leave it as it is.

chrisdel101: Oh really? Okay then.
I'd still ask for that change in the title wording unless you are against it.
This might be outdated now....

chrisdel101: Also I misread. U did change the title. LGTM.

chrisdel101 (approved):

bjohansebas:
btw, I couldn't find a way to translate this page into other languages, it always loaded the English version.

bjohansebas:
Can someone with write permissions review it? @expressjs/docs-captains @UlisesGascon

carlosstenzel:
> btw, I couldn't find a way to translate this page into other languages, it always loaded the English version.

I don't think it makes sense to translate a 404 page, as we believe that everyone who is searching the site has some knowledge about what a 404 would be and what it means.

UlisesGascon (approved):

Issue #1596, opened 2024-09-05
Use a code formatter


bjohansebas:
Implementing a formatter like Prettier or Biome would help maintain better consistency in code style. Although standard.js already handles the style, it's only for JavaScript. It would be better to have a tool capable of maintaining consistency in HTML, CSS, or any other language in the repository as well.

cc: @crandmck

chrisdel101:
I'd be open to this since we are the only main contributors working right now, and it'd make for a better developer experience. But its a non trivial amount of work to maintain formatting across a team of many people.

It would either help us get rid of prettifying junk in our PRs, or it could make it worse.

Not a bad suggestion though!

crandmck:
This is reasonable, but I'm not sure it's worth the effort.  I have some other suggestions that might have more value--will try to find time to put them in an issue.

bjohansebas:
btw, we’re not actually using a linter or a formatter for the .js files. ESLint is only being used for the code blocks inside markdown files. Additionally, in tests, it didn’t work well to use Prettier for general use because we’re using Liquid, which makes Prettier act a bit erratically. Hopefully, we can solve this issue in the future.

bjohansebas:
@expressjs/docs-wg what do you think about using Prettier to initially format the CSS?

chrisdel101:
I like the idea of using a formatter, and for main contributors it should be pretty straightforward since we're using this code frequently, and can debug it.
But it can be annoying when someone clones the repo and the formatter doesn't work and needs a bunch of debugging. Happens to me sometimes. 

How much effort would it take to maintain it being in the workflow? I'm pro-formatter overall though.

bjohansebas:
The redesign will include it

Issue #1166, opened 2020-06-11
Is the copyright and license info on the footer still accurate? 


mastermatt:
The footer currently lists this info (in English, but most of the data is consistent across locales):
- Documentation translations provided by StrongLoop/IBM.
- Express is a project of the OpenJS Foundation.
- Edit this page on GitHub.
- Copyright © 2017 StrongLoop, IBM, and other expressjs.com contributors.
- Creative Commons License This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 United States License.

I'm fuzzy on the details of the StrongLoop/IBM/OpenJS F./Linux F. transition. Does StrongLoop still hold the copyrights to some/all of Express? Does this cover the lib or just the website? 

And I see that this repo and the site hasn't updated its license since 2016. I don't have reason to question this expect that it was before the transition and Express itself is MIT. Can someone give me the layman's tl;dr on having different licenses?

jonchurch:
Most of these are before my time, but the only one that stands out to me here is the Copyright notice of StrongLoop/IBM. I'm not sure how that works exactly, but I think the notice is probably still accurate in regards to the website content.

For the Creative Commons license, there's some discussion here https://github.com/expressjs/expressjs.com/issues/634 and here https://github.com/expressjs/expressjs.com/issues/413

The TLDR is that CC license is intended for content (works of art, text, etc) vs MIT which is for software specifically. It covers the website, not the library.

Express is an OpenJS project, so I know that's accurate.

crandmck:
Since I was there, maybe I can shed some light.  Caveat: I'm not a lawyer or legal expert by any means....

I agree, as @jonchurch said, I think the copyright applies to the website content... It's there because IBM/StrongLoop paid salaries for several people (including myself) to completely overhaul the documentation over the course of a couple months.  We started with what TJ wrote, which had some good technical info, but was all in one giant page, not well organized or written, nor easy to navigate, and was somewhat outdated.  We converted it into the site you see today, and added quite a bit of material (e.g. "Getting Started", lots of "Guides", etc).  IBM also paid to translate the content (professional technical translation is very expensive).  Since then, there has been _extensive_ improvement and maintenance by @dougwilson and many others in the community; but IIUC that doesn't change the original copyright.

[...]

PR #1695, opened 2024-11-23
feat: add language picker

### LINKED ISSUE #1687: Add language selector
To improve the experience between i18n pages, adding a language selector would be very helpful. Right now, it's difficult to switch from one language to another without leaving the page you're on.

cengizcmataraci:
Adding language picker to header
closes #1687


bjohansebas (commented):
Maybe we don't want it to be removed in mobile until it really breaks the mobile responsiveness.

With the selector, I think this part of the page can be removed now.
image


I really love that this is being added.

cengizcmataraci:
I don't add the picker in the mobile, but I'll add it and remove the language part if you folks want. And I also like the picker. 😄

carlosstenzel:
@cengizcmataraci 
good job ;)

cengizcmataraci:
Hello, sorry about the delay. Could you review again, please? I added the language picker for mobile. I tested for various devices in browser dev tools. @bjohansebas @carlosstenzel

carlosstenzel (approved):

js/menu.js
```diff
@@ -31,46 +31,88 @@ for (const el of itemsMenu) {
...
+for (const el of languageItems) {
+	el.addEventListener("click", (e) => {
+		const href = el.getAttribute("href");
+
+		if (href && href !== "#") {
+			languagePickerMenu?.classList.remove("opens");
+			overlay?.classList.remove("blurs");
+			document.body.classList.remove("no-scroll");
+
+			window.location.href = href;
```
bjohansebas, line 74: Why is it necessary to do this?

cengizcmataraci: In this for loop, I am iterating over the languageItems elements, and if the clicked item’s href attribute exists, the page’s URL is updated accordingly. For example, if a user is on a German page and clicks the Turkish option, the page is directly redirected to the Turkish version.

bjohansebas: Yep, but wouldn't the href of the link do the same?

cengizcmataraci: Yes, you’re right, but in this case, I wouldn’t be able to handle the extra conditions above. I can change it if you’d like.

bjohansebas: I think this could indeed be removed, I’ve tested it and didn’t notice any strange behavior. Or, @cengizcmataraci did you encounter any unusual behavior while developing? If not, should this line be added?

_includes/language-picker-mobile.html
```diff
@@ -0,0 +1,18 @@
+<div id="mobile-menu">
+  <div id="language-picker-button" class="fa fa-language fa-2x button"></div>
```
bjohansebas: I’d prefer to change this icon for one that looks more modern, maybe one from [Heroicons](https://heroicons.com/)? I could make the change tomorrow.

js/menu.js
cengizcmataraci: No, you are correct. I will remove this line now, thank you.

_includes/language-picker-mobile.html
cengizcmataraci: Allright, could you check it again, please? @bjohansebas

bjohansebas: Great, thanks for the change, it looks modern

js/menu.js
bjohansebas: Okay, now it doesn't work when it's on mobile.

bjohansebas:
I will keep the icon for all screen sizes

bjohansebas (approved):
Thank you for working on this @cengizcmataraci, it really looks great

_includes/footer/footer-en.html
```diff
@@ -1,21 +1,15 @@
 <a id="top" href="#"><img src="/images/arrow.png"></a>
 
 <footer>
-  <section id="doc-langs">
-    Documentation translations provided by StrongLoop/IBM:
```
ctcpip: we can't remove this

_includes/footer/footer-id.html
```diff
@@ -1,21 +1,15 @@
 <a id="top" href="#"><img src="/images/arrow.png"></a>
 
 <footer>
-  <section id="doc-langs">
-  Dokumentasi versi terjemahan yang disediakan oleh StrongLoop/IBM:
```
ctcpip: we can't remove this

ctcpip (changes requested):
don't the other footers need to be updated?

_includes/footer/footer-en.html
bjohansebas: @ctcpip why?

_includes/footer/footer-id.html
bjohansebas: why?

_includes/footer/footer-en.html
ctcpip: They likely agreed to do the work and specifically asked that this attribution/credit appear there.

Some background [is here](https://github.com/expressjs/expressjs.com/issues/1166#issuecomment-664030084).

_includes/footer/footer-id.html
ctcpip: https://github.com/expressjs/expressjs.com/pull/1695#discussion_r1913814532

_includes/footer/footer-en.html
bjohansebas: Interesting context, so we will keep it.

bjohansebas: The reason I was thinking of removing it, in case it serves as context in the future, was because the following banner was already displayed in each language.

image

Which displayed their respective acknowledgment, where they translated it

ctcpip: sure, but that only appears on the translated pages. the main documentation site in English would not display the attribution/credit

css/style.css
```diff
@@ -1032,6 +1043,33 @@ h2 a {
...
+
+  .header-right {
+  display: flex;
+  align-items: center;
+  gap: 8px;
+  margin-right: -10px;
+  }
+
+  #language-picker-menu #navmenu>li:first-child {
+    display: flex !important;
```
bjohansebas: We try not to use `!important`, please find a way to remove it.

_includes/footer/footer-en.html
bjohansebas: @ctcpip Back added

bjohansebas (changes requested):
Please remove the !important

bjohansebas:
On mobile, I think it would be better to handle it the same way as on desktop, meaning the menu doesn't take up the full width, similar to how Node.js does it

image

And maybe always keep the icon instead of showing the language name

css/style.css
cengizcmataraci: Sure, I will look at it.

cengizcmataraci:
> On mobile, I think it would be better to handle it the same way as on desktop, meaning the menu doesn't take up the full width, similar to how Node.js does it
> 
> image
> 
> And maybe always keep the icon instead of showing the language name

Are you referring to the desktop version? Because we don’t display the language name in the mobile picker. And do we really think the Node.js website offers a better UX for the mobile picker? I find it a bit confusing. In our design, the user opens the picker solely to change the language, and once that’s done, it closes automatically. @bjohansebas

bjohansebas:
I'm not an expert in UX/UI to determine if Node.js has the best mobile user experience, it was just a suggestion, so I'm okay with leaving it as it is, which I also think is great.

carlosstenzel:
In my analysis, I believe it is good for this first version.

Captura de tela 2025-01-16 081427
Captura de tela 2025-01-16 081804

cengizcmataraci:
Okay, I removed the important. Could you review again, please? @bjohansebas @carlosstenzel @ctcpip

bjohansebas (approved):
LGTM

ctcpip (approved):

css/style.css
```diff
@@ -1018,6 +1029,33 @@ h2 a {
   #doc-langs {
     font-size: 11px;
   }
+
+  .header-right {
+  display: flex;
+  align-items: center;
+  gap: 8px;
+  margin-right: -10px;
```
dpopp07: This block is missing indentation:

```suggestion
    display: flex;
    align-items: center;
    gap: 8px;
    margin-right: -10px;
```

bjohansebas: That doesn't really matter much, we need to fix the styling issue at some point to make it uniform, you can check https://github.com/expressjs/expressjs.com/issues/1596.

dpopp07: Agreed. Still, it felt worth noting as the rest of the PR was internally consistent.

bjohansebas: fixed, thanks @dpopp07

bjohansebas (approved):

bjohansebas:
awesome work @cengizcmataraci,

Issue #1590, opened 2024-09-02
refactoring from jQuery to modern JavaScript


bjohansebas:
As discussed in #1568, these will be the next steps to stop using jQuery in favor of the browser's API:

- [x] **First PR (Refactor Menu):** #1568
  - Remove dropit
  - Remove functions related to dropit
  - Refactor the menu according to the new structure
  - Create styles according to the new structure
  - Add new menu functionality in a new file

- [x] **Second PR (Remove Unused Functions):** #1698
  - Remove ismobil.js
  - Remove logic for page editing (no longer used since the footer was updated)
  - Remove logic for [`window.location.pathname === '/3x/api.html'`](https://github.com/expressjs/expressjs.com/blob/83cf6069eca7670210c856fca51d6db4dded2c0d/js/app.js#L121) (no longer used since almost all URLs now go to `/en/*`)
  - Remove function related to [`#tags-side-menu`](https://github.com/expressjs/expressjs.com/blob/83cf6069eca7670210c856fca51d6db4dded2c0d/js/app.js#L155)

- [ ] **Third PR (Adapt get elements):**
  - Update scroll top logic
  - Adapt element retrieval and events
  - Add overlay logic to the menu
  - Adapt code highlighting

- [ ] **Fourth PR (Adapt Functions):**
  - Update logic for the i18n banner
  - Update logic for the closer title (used in the API)
  - Adapt cookies logic to ES6

- [ ] **Fifth PR (Remove jQuery Dependency, #1569):**
  - Remove link to jQuery

bjohansebas:
The search bar in dark mode is broken in terms of design. @chrisdel101 Should I fix it in the first PR or better in a separate PR?

![image](https://github.com/user-attachments/assets/5b93173b-37c4-45e7-bbac-b62bac744cbe)

chrisdel101:
I'm just looking at #1590 since I think you fixed it over there right? Sorry
 for delay.

bjohansebas:
@chrisdel101 don't worry, that issue is not fixed yet.

chrisdel101:
Oh is this a different issue? I see. Maybe a small separeate PR would work, but things are getting kind of backed up. Still, it's not really related to the refactor, so separate issue would be my vote.

ShubhamOulkar:
@bjohansebas If you allow me, I like to work on this.

bjohansebas:
@ShubhamOulkar Yes, please finish it.

[...]

PR #1698, opened 2024-11-25
feat: remove unused functions (#1590)

bjohansebas:
The second PR, which is part of #1590, removes unused functionalities from the page.


js/app.js
```diff
@@ -26,34 +25,6 @@ $(function(){
...
-
-  if (lang === 'en') {
-    if (pathName == '/') editLink = '<a href="' + branchPath + '">Fork the website on GitHub</a>.';
-    else editLink = '<a href="' + editPath + '">Edit this page on GitHub</a>.';
-    $('#fork').html(editLink);
-  }
-
   // code highlight
 
   $('code.language-js').each(function(){
```
chrisdel101, line 30: More descriptive:
Something like this maybe "Add class for Prism JS code highlight"

js/app.js
```diff
@@ -68,9 +39,6 @@ $(function(){
 
   // menu bar
 
-  var prev;
-  var n = 0;
-
   var headings = $('h2, h3').map(function(i, el){
```
chrisdel101, line 42: Is there any reason this can't be CSS? There might be a good reason. If not then CSS-ify it.

js/app.js
```diff
@@ -5,7 +5,6 @@
```
chrisdel101, line 5: RE: comments. I'm an all-in supporter of the "comments good" school. I think this can be commented better.

The comments that exist are not that helpful, but better than nothing. 
Each non-obvious func, which is most, should have a short but useful comment, so I suggest adding and updating. W/o it's more time consuming for others to analyze down the line - I literally have no idea what these funcs do and have to pull the branch to figure it out. 

ELSE remove the comments. None is better than bad or missing.

Can we change copyright too? 
`Copyright Express and expressj.com Contributors` since no strongloop or IBM now? + this JS has not been redone by this team.

js/app.js
```diff
@@ -99,50 +67,25 @@ $(function(){
   });
 
   $(document).scroll(function() {
```
chrisdel101, line 69: Comment what this does.

js/app.js
```diff
@@ -5,7 +5,6 @@
 
 $(function(){
   var doc = $(document);
-  var lang = document.location.pathname.split('/')[1]
 
   // top link
```
chrisdel101, line 9: Top link what?

▼▼▼▼

Scrolling links what?

js/app.js
```diff
@@ -99,50 +67,25 @@ $(function(){
...
+    lastApiPrefix = currentApiPrefix.split('.')[0];
   })
-  $('#tags-side-menu li').on('click', function() {
-      // Remove prev 'active's 
-      $(this).next().siblings().removeClass('active');
-      $(this).next().addClass('active')
-    })
 
   // i18n notice
   if (readCookie('i18nClose')) {
```
chrisdel101, line 91: `Show/hide i18n notices on translated pages`

chrisdel101 (commented):
I don't see any bugs in the site working. I don't really know what mobile js does but I never used it so happy to see it go. 
I added some ideas. These are suggestions only. I'd recommend but up to you if you agree. 
Good idea doing this.

js/app.js
bjohansebas: I’m not really sure if there’s a reason for it not to be CSS. I’ll try it in the next PR. This one focused on removing unused stuff to avoid noise in the other two PRs.

js/app.js
bjohansebas: This will be modified in the other two PRs, i’ll make the change to the comment there.

js/app.js
bjohansebas: This will be modified in the other two PRs, I’ll make the change to the comment there.

js/app.js
bjohansebas: The comments will be changed as I progress through the PRs, although I see your point, and it makes the review more difficult not having descriptive comments. I'll go ahead and add them.  

As for the copyright, it would be better to handle it in a separate PR to avoid blocking this one, since it's not clear what the TC's opinion is on this.

chrisdel101 (approved):
Okie if you're doing these later than I guess overall I'd just like a one line comment on all non-super obvious JS funcs in the next iteration. 
Otherwise I don't see these changes break anything.

bjohansebas:

wesleytodd (approved):

PR #1699, opened 2024-11-25
fix: fix menu text align

bjohansebas:
Fix a bug that was accidentally introduced in #1670

before:
image

after:
image

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -961,7 +961,6 @@ h2 a {
   #home-content {
     margin: 60px 0 0 5%;
     padding-right: 5%;
-    text-align: center;
   }
 
   #description {


chrisdel101:
I like the new way with the links centered, better than original left-positioned. But it's buggy. 
Any chance of just fixing up this method? I saw this in the PR and liked it, but missed the bugginess. 
Screenshot 2024-11-25 083032

bjohansebas:
I also liked that it was centered, so I’m going to center the menu.

carlosstenzel (approved):

bjohansebas:

ctcpip:
we should not be centering text as this is bad for [a11y](https://www.w3.org/TR/2008/REC-WCAG20-20081211/#visual-audio-contrast-visual-presentation).

suggest undoing the newest commit

ctcpip (changes requested):
.

bjohansebas:
@ctcpip  The link you provided does not mention anything about centering the text, only about justifying the text, which is indeed a bad practice in terms of accessibility.

Oops, I was mistaken with the nickname, sorry.

ctcpip:
in that context, centered is included. these are well-established accessibility guidelines

https://www.w3.org/WAI/GL/low-vision-a11y-tf/wiki/Supplemental_Guidance:_Text_Justification

ctcpip:
clear usability issue here

Screenshot 2025-01-13 at 2 44 58 PM

bjohansebas:
Okay, thanks for the resource, I will remove it

css/style.css
```diff
@@ -961,6 +961,9 @@ h2 a {
   #home-content {
     margin: 60px 0 0 5%;
     padding-right: 5%;
+  }
+  
+  #homepage-leftpane {
     text-align: center;
   }
```
ctcpip:
```suggestion
```

ctcpip (changes requested):
left pane should not have centered text

css/style.css
bjohansebas: 😞 
image

bjohansebas: fixed

ctcpip (changes requested):
this is the wrong change...

css/style.css
ctcpip: you changed the wrong CSS selector/rule

bjohansebas:
Nothing will be centered anymore if #1732 is merged

ctcpip:
@bjohansebas that's... not true at all

edit: I'm not sure what you meant, but the layout will still be centered as done in https://github.com/expressjs/expressjs.com/issues/1668

ctcpip (approved):

PR #1700, opened 2024-11-26
Remove retina.js and convert icons to SVG

bjohansebas:
retina.js is removed, as it was a script used to load larger images, adding unnecessary additional load. therefore, i have opted to change those icons to svg.


carlosstenzel (approved):

bjohansebas:

UlisesGascon (approved):
LGTM!

PR #1701, opened 2024-11-27
update copyright

bjohansebas:
The copyright of the site is updated. I’m not sure what sense it makes to add them to the files, but it’s good to update it.

As suggested in https://github.com/expressjs/expressjs.com/pull/1698#discussion_r1858643033

cc: @expressjs/express-tc

### DIFF
--- a/_layouts/3x-api.html
+++ b/_layouts/3x-api.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <!---
- Copyright (c) 2016 StrongLoop, IBM, and Express Contributors
+ Copyright (c) 2024 OpenJs Foundation and Express Contributors
  License: MIT
 -->
 <html lang="{{ page.lang }}">
--- a/_layouts/4x-api.html
+++ b/_layouts/4x-api.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <!---
- Copyright (c) 2016 StrongLoop, IBM, and Express Contributors
+ Copyright (c) 2024 OpenJs Foundation and Express Contributors
  License: MIT
 -->
 <html lang="{{ page.lang }}">
--- a/_layouts/5x-api.html
+++ b/_layouts/5x-api.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <!---
- Copyright (c) 2016 StrongLoop, IBM, and Express Contributors
+ Copyright (c) 2024 OpenJs Foundation and Express Contributors
  License: MIT
 -->
 <html lang="{{ page.lang }}">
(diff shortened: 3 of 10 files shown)


UlisesGascon (commented):
I like the idea! Here are some suggestions:
- We should respect the previous copyright and add additional lines when is required, like in [this file](https://github.com/expressjs/express/blob/master/lib/request.js):
	```
	/*!
	 * express
	 * Copyright(c) 2009-2013 TJ Holowaychuk
	 * Copyright(c) 2013 Roman Shtylman
	 * Copyright(c) 2014-2015 Douglas Christopher Wilson
	 * MIT Licensed
	 */
	```
- 2024 is not the correct year, probably is 2016 ([see](https://nodejs.org/en/blog/announcements/foundation-express-news)) back then it was the Node.js foundation, and just recently OpenJS foundation (2019? [see](https://en.wikipedia.org/wiki/OpenJS_Foundation)). But not sure if the copyright is now owned by them or by the project (cc: @wesleytodd ) :thinking:

bjohansebas:
As I understand it, the copyright would belong to OpenJS. If they recommend adding the following text in the footer:

image

ref: https://github.com/openjs-foundation/artwork#copyright-notices-for-project-website-footers

and maybe the year should be from 2019 to the present, perhaps it would be good to clarify this in the CPC?

wesleytodd:
Yeah I agree I think we need to append not update the existing lines. For the date, I am not actually sure how much it matters to be accurate? This is more of a question for the foundation folks. I will link to this issue in slack and see if there is a clear guidance we can follow.

ctcpip (changes requested):
## edit: see my [newest response](https://github.com/expressjs/expressjs.com/pull/1701#issuecomment-2583677907) below

there's not really a need for this change, and copyright notices are not even necessary for copyright protection

that being said, my suggestions for this PR would be:

- don't add new notices where they don't exist -- they are not necessary at all, and especially not in source code files
- you are supposed to at least indicate the earliest year for the copyright, not the newest
- it is only necessary to list the earliest year -- there is no value in providing a range
- consider removing all of these notices -- again, they are not necessary and just take up code comment space for no benefit

ctcpip:
see also:  https://github.com/openjs-foundation/cross-project-council/blob/main/governance/IP_POLICY_GUIDANCE.md#2-copyright-notices

importantly:  

> Please note that you must not change or remove existing copyright lines unless you put them there and have the right to do so.

bjohansebas:
Really, this would be eliminated anyway if we achieve https://github.com/expressjs/discussions/issues/211, and +1 on removing them completely

Since we won't need to modify style or logic with JavaScript

ctcpip:
ok so on further consideration of the IP guidance, my new suggestion is:

1. don't add any new notices
2. do not change any of the existing lines
3. append to existing lines the following:

> Copyright The Express Contributors.

ctcpip:
closing as unnecessary based on our discussion in slack

Issue #1676, opened 2024-10-30
heads up: Potential Shift to Machine Translations on the Horizon


UlisesGascon:
In the last TC meeting, we discussed the future of translations. Here are some notes:

**Current Situation**
- Translations are quite outdated across all languages.
- Currently, we rely on the community for translation efforts.
- There are instances where users encounter outdated information on the website, which can lead to frustration.
- We are planning to migrate the website, and updating/moving translations will add a significant workload when we release the content.

**Proposed Solution**

The idea is to explore using automated translations by default (machine translations). This would require some effort to build the CI system and ensure that the English version of the documentation is optimized for machine translation (plain English, simplified structure, etc.).

This approach could greatly simplify the migration, as it allows us to focus on i18n support rather than handling both translation and content migration.

Of course, we would still enable the community to tweak and improve translations as needed. This way, we can continue providing high-quality documentation for everyone.

**Let’s Discuss**

What do you think? (@crandmck, @expressjs/docs-wg, and @expressjs/express-tc)

Additionally, I wonder if we should consider freezing translation efforts for now (such as issues like [#1480](https://github.com/expressjs/expressjs.com/issues/1480), [#1553](https://github.com/expressjs/expressjs.com/issues/1553)) and redirecting that team’s efforts toward planning the future website so we can kick that off in 2025.

chrisdel101:
Assuming the translations themselves are adequate this seems like a good idea since many translations are missing items, and as you said, some might have incorrect info. I'd be nice to have continuity across the translations, with all of them having the same links and pages in their languages. 

At least the look of the home pages are consistent due to #1568. 

Was there any idea of what changes would be needed to make this work? I wonder if this work can be bundled with any site restructuring that @bjohansebas has mentioned wanting to do in passing. (Thinking he meant like more flexbox and page containers if I understood?)

What format would the translations provide? Could it be like Json? If it's running in the CI and would run every build, and gave json (or some other data) it would be amazing if it was possible to have only a single site markup! Instead of a separate one for each language. And inject the json into the page markup at build time!
Just my two cents...

bjohansebas:
I'll leave my thoughts and questions here later, but I think #1553 could move forward since the [PR](https://github.com/expressjs/expressjs.com/pull/1642) is ready and we just need to add the PAT, while #1480 needs a bit more research, so it would be better to freeze it for now, if that sounds good to you.

crandmck:
[...]

PR #1642, opened 2024-10-04
fix permissions for translation workflow

bjohansebas:
This should fix the workflow for people who do not have write permissions by adding a new custom token

The requested token should have the following permissions:

```
permissions:
  contents: read
  pull-requests: write
  issues: write
```

ref: https://github.com/actions/github-script/tree/v7?tab=readme-ov-file#using-a-separate-github-token

close: #1553

PR #1710, opened 2024-12-09
ci: delete translation workflow

bjohansebas:
Unlike what #1642 does, this workflow is removed due to permission issues. 

Maybe what we want to do is create a bot on GitHub that adds these labels like Node.js does. Or maybe this won't be necessary either if we manage to automate the translations.

ref: 
- #1676
- https://github.com/nodejs/github-bot
- https://github.com/expressjs/discussions/issues/309


carlosstenzel (approved):

UlisesGascon (approved):
Following with the last discussions around this, seems like this pipeline won't be necessary as we move to machine translations :+1:

UlisesGascon:
I will land it as this pipeline is broken anyway (with the permissions issue). This will reduce the noise in new and existing PR(s) cc: @expressjs/docs-captains

Issue #1707, opened 2024-12-04
typo causing broken link redirection in spanish docs 


1fbr:
### Description

On the **Spanish version** of the "Static Files" page ([https://expressjs.com/es/starter/static-files.html](https://expressjs.com/es/starter/static-files.html)), the "documentación en inglés" link points to an incorrect URL:

- **Incorrect URL**: [https://expressjs.com/en/starter/static-filen.html](https://expressjs.com/en/starter/static-filen.html)
- **Expected URL**: [https://expressjs.com/en/starter/static-files.html](https://expressjs.com/en/starter/static-files.html)

This typo ("**filen**" instead of "**files**") results in a 404 error.

---

### Supporting Evidence
1. Screenshot of the Spanish page:
   ![Spanish Page Issue](https://github.com/user-attachments/assets/d26127ee-0692-4e55-9117-e1a31f0d7c38)

2. Screenshot of the broken link:
   ![Broken Link](https://github.com/user-attachments/assets/d1a7fc1b-3013-46cb-af54-45273ea4d71f)

---

### Attempts to Debug
I tried searching for the typo `"filen"` in the repository but couldn't locate it. I guess the URL is being dynamically generated, but I have no clue on how to fix it.

---

### Expectations

Should redirect to this page: [https://expressjs.com/en/starter/static-files.html](https://expressjs.com/en/starter/static-files.html)

bjohansebas:
Wow, that's very strange, that shouldn't happen

1fbr:
Wow, seeing your response, it gave me an idea, out of the blue, and I realised what's going on! It seems that when replacing the language code *es* with *en*, it’s matching the *es* in the filenames too.

To verify, I checked for similar cases, and the same issue occurs on pages like:

 https://expressjs.com/es/guide/behind-proxies.html
 https://expressjs.com/es/guide/using-template-engines.html

This pattern happens with any file whose name ends in *es*. So, yeah, it’s not a simple typo but a bug.

 I hope this helps @bjohansebas !

bjohansebas:
@1fbr Thanks for the information, this is really helpful. 

PRs are always welcome if anyone has a possible solution in code.

PR #1711, opened 2024-12-10
fix -update notice-{x}.md replace link

carlosstenzel:
fix reported bug

Changing replace logic, to ensure only the language in url is changed

https://github.com/expressjs/expressjs.com/issues/1707

The error occurs when changing a string in the URL.

For example

`../es/...files.html`

In the language **es** to change to **en**, it ends up changing the name of the file and not just the language, becoming:

`/en/..filen.html`


_includes/notice/notice-de.md
```diff
@@ -1,3 +1,3 @@
 <p>Diese Übersetzung zur Verfügung gestellt von StrongLoop / IBM.</p>
 
-Dieses Dokument kann im Vergleich zur englischen Dokumentation veraltet sein. Aktuelle Updates finden Sie in der <a href='{{ page.url | replace: page.lang, "en" }}'>englischen Dokumentation</a>.
+Dieses Dokument kann im Vergleich zur englischen Dokumentation veraltet sein. Aktuelle Updates finden Sie in der <a href='{{ page.url | replace: "de/", "en/" }}'>englischen Dokumentation</a>.
```
bjohansebas: I think this would be a bit safer

```suggestion
Dieses Dokument kann im Vergleich zur englischen Dokumentation veraltet sein. Aktuelle Updates finden Sie in der <a href='{{ page.url | replace: "/de/", "/en/" }}'>englischen Dokumentation</a>.
```

bjohansebas (approved):
Great

PR #1705, opened 2024-12-03
docs: add v5 migration note for app.listen

dpopp07:
Adds a note to the `migrating-5` guide alerting users that error events received by a server will now cause the callback argument in `app.listen` to be invoked. In Express 4, this was not the case and errors would be thrown.

Resolves https://github.com/expressjs/express/issues/6191

Any feedback, rewording, etc. is welcome!

PR #1712, opened 2024-12-10
add examples for migration v5

bjohansebas:
Examples are added for better understanding of the changes that were made; this was an idea that came from #1705.

### DIFF
--- a/en/guide/migrating-5.md
+++ b/en/guide/migrating-5.md
@@ -14,8 +14,8 @@ Express 5 is not very different from Express 4; although it maintains the same b
 
 To install this version, you need to have a Node.js version 18 or higher. Then, execute the following command in your application directory:
 
-```console
-$ npm install "express@^{{ site.data.express.next_version }}"
+```sh
+npm install "express@^{{ site.data.express.next_version }}"
 ```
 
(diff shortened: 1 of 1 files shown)


en/guide/migrating-5.md
```diff
@@ -95,44 +127,186 @@ This should not affect your code if you follow the Express 4 documentation of [a
...
+})
+
+// v5
+app.get('/*splat', async (req, res) => {
+  res.send('ok')
+})
+```
+
+{% capture note_wildcard %}
+If your defined route is `/*splat`, it will not include `/`, so you will need to make the route optional.
```
bjohansebas: @blakeembrey 
Is this wording correct? It’s a recurring problem i’ve seen, which is why I wanted to add that note, but I’m not sure if this is the best phrasing.

wesleytodd: > so you will need to make the route optional.

I am not sure I follow what you are saying here. Is it not enough to just end that sentence after the first part?

wesleytodd (approved):
Other than the one comment I made (which is more of asking for clarification than a blocker) this looks good!

en/guide/migrating-5.md
bjohansebas: @wesleytodd I changed the phrasing a bit, I think it’s clearer now.
> `*splat` matches any path without the root path. If you need to match the root path as well `/`, you can use `/{*splat}`, wrapping the wildcard in braces.

UlisesGascon (approved):

PR #1713, opened 2024-12-11
improve changelog and add 4.21.2

bjohansebas:
This change improves the way versions are signaled in the changelog, as well as adding the recent security patch.


bjohansebas:
If we want to have a single source of truth, we might prefer to redirect to https://github.com/expressjs/express/releases

UlisesGascon (approved):

Issue #1716, opened 2024-12-16
The links of the "Additional Examples" section in the docs are broken


nicolasgandrade:
### Description

Currently, all the 4 links under the "Additional Examples" section in the docs are not working. All those links redirect to legacy folders that do not exist anymore in the Prisma project.

URL: https://expressjs.com/en/starter/examples.html#additional-examples

### Expectations

The links should redirect the user to the existing folder in the Prisma project.

nicolasgandrade:
I can replace the old links with the new ones. My only concerns are about the examples using JavaScript and GraphQL, I couldn't find them in the current version of the Prisma repository.

Should we remove that item from the docs?

bjohansebas:
Hi @nicolasgandrade, PRs are always welcome

> Should we remove that item from the docs?

Yes, if there is no replacement, it should be removed.

nicolasgandrade:
Right, I'll get the updated ones and open a PR fixing that. Thanks for the response

bjohansebas:
fixed by #1717

PR #1717, opened 2024-12-17
Fix broken urls in the examples page and translate the page to pt-BR

nicolasgandrade:
This PR replaces 2 links in the examples page that were broken with new ones, removes other 2 that used to redirect to examples that **are not** in the Prisma repository anymore, and also translates the `examples.md` page to Portuguese.

I could have changed the links of the examples that are not anymore in the Prisma repo to one pointing an old commit, however I decided not to do so because since they were permanently removed, I assumed that those examples are not supported anymore. Also, there's no new alternative in the Prisma repo.

Related issue: https://github.com/expressjs/expressjs.com/issues/1716


bjohansebas (approved):
LGTM

carlosstenzel:
@bjohansebas 

If we were to create a new page 'example ' for pt-br, we would have to create it for all other languages, and adjust the menu as well.

bjohansebas:
Sorry for the late response, I had taken a break. It's true that the pages should be added to the other languages in English, but since the author had only focused on those links and adding them to their language, I didn't want to ask them to add it to the other languages. @nicolasgandrade, would you like to add the 'examples' page to the other languages?

carlosstenzel:
@bjohansebas 

I am proposing an improvement in another PR, adding only the menu item. https://github.com/expressjs/expressjs.com/pull/1724

(Considering that this PR has no modifications, you could close it and just add the reference )

nicolasgandrade:
Hey guys, my bad for the delay, I was in a hurry with a delivery in another project.

The #1724 PR makes sense. Actually, I didn't know we could add the item to the menu referencing another language, but yes, I agree that's better than having the item only in the EN and PT pages.

Since @carlosstenzel has already done those changes in his PR, I think I don't have anything else to add, but if there's something else that I'm missing, ping me here.

carlosstenzel:
@nicolasgandrade 

Poderia fazer nesse PR 
- remover a alteração no menu
- atualizar a branch
- Atualizar os links das paginas existentes de exemplo em en, pt-br e outras que tenham o mesmo.
- alterar o nome do PR 

Acho que essas alterações fazem mais sentido para seguirmos com o PR.

Obrigado pela sua contribuição ;)

bjohansebas:
Here it is ready, @carlosstenzel, can you confirm that the Portuguese translation is correct so we can merge it?

nicolasgandrade:
@carlosstenzel Right, I removed the change in the menu, updated the branch and also updated the links in the Indonesian page (this one was the only one translated, the other ones are all pointing to the english version, which was updated in this PR).

Just a question: since the ideia (as far as I understood) is to have the menu link pointing to the english page in all languages for now, should I leave the portuguese translation in this PR as well even not being used now or should I just remove the file from this PR?

bjohansebas:
Please, keep the Portuguese translation

carlosstenzel (approved):

PR #1730, opened 2025-01-13
📄 update license to `CC BY 4.0`

ctcpip:
switches to the [OpenJS Foundation preferred license](https://github.com/openjs-foundation/cross-project-council/blob/main/governance/IP_POLICY_GUIDANCE.md#1-licensing) which is more permissive than the currently used `CC BY-SA 3.0`

edit: we could instead use the MIT license.  I don't have strong feelings either way and only chose CC because the existing license is CC

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,123 +1,397 @@
-# License
-
 Copyright The Express Contributors.
 
-# Creative Commons Attribution-ShareAlike 3.0
+Attribution 4.0 International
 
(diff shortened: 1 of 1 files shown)


bjohansebas:
I would prefer to keep it as MIT, and have consistency across all the repositories in the organization

carlosstenzel (approved):

wesleytodd:
IANAL so have no strong opinion either way, but I agree that consistency across the project is a benefit of going MIT. Is there no more process or requirements for changing the licensing of the docs?

ctcpip:
> Is there no more process or requirements for changing the licensing of the docs?

nope, as long as we are using a license approved by the foundation, which this one and the MIT license both are

bjohansebas (changes requested):
To prevent this from being merged until we decide which license to have.

UlisesGascon (approved):
LGTM! I am also +1 for MIT if we decided to use that one

wesleytodd (commented):
I would be fine with this *or* MIT. MIT is consistent with all the other code project's licenses, and from my very limited quick research they are similar enough that it shouldn't have a meaningful effect on anything we are doing here.

jonchurch (approved):

ctcpip:
discussed at TC meeting today.  we agreed to proceed with the CC license, but left the possibility open for future licensing under MIT, or dual-licensing

bjohansebas (approved):

bjohansebas:
@ctcpip whenever you want, you can merge it. I'm not sure if there's anything else to do.

PR #1733, opened 2025-01-14
Added the minimum supported node.js version to the support page

Phillip9587:
Added the minimum supported node.js version to the support page. Please correct me if I'm wrong about the minimum versions for the 3.x, 2.x, 1.x and 0.14.x release lines. 

Preview: https://deploy-preview-1733--expressjscom-preview.netlify.app/en/support/

### DIFF
--- a/de/support/index.md
+++ b/de/support/index.md
@@ -11,14 +11,14 @@ Only the latest version of any given major release line is supported.
 
 Versions that are EOL (end-of-life) _may_ receive updates for critical security vulnerabilities, but the Express team offers no guarantee and does not plan to address or release fixes for any issues found.
 
-| Major Version | Support Start Date | Support End Date |
-| -- | -- | -- |
-| [**v5.x**{: .no-release }](/{{page.lang}}/5x/api.html) | **not yet released**{: .no-release } | **not yet released**{: .no-release } |
-| [**v4.x**{: .supported }](/{{page.lang}}/4x/api.html) | April 2014 | **ongoing**{: .supported } |
-| [**v3.x**{: .eol }](/{{page.lang}}/3x/api.html) | October 2012 | July 2015 |
-| [**v2.x**{: .eol }](/{{page.lang}}/2x/) | March 2011 | July 2012 |
-| **v1.x**{: .eol } | December 2010 | March 2011 |
-| **v0.14.x**{: .eol } | December 2010 | December 2010 |
+| Major Version | Minimum Node.js Version | Support Start Date | Support End Date |
+| -- | -- | -- | -- |
+| [**v5.x**{: .supported }](/{{page.lang}}/5x/api.html) | 18.0.0 | September 2024 | **ongoing**{: .supported } |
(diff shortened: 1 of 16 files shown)


de/support/index.md
```diff
@@ -11,14 +11,14 @@ Only the latest version of any given major release line is supported.
...
-| -- | -- | -- |
-| [**v5.x**{: .no-release }](/{{page.lang}}/5x/api.html) | **not yet released**{: .no-release } | **not yet released**{: .no-release } |
-| [**v4.x**{: .supported }](/{{page.lang}}/4x/api.html) | April 2014 | **ongoing**{: .supported } |
-| [**v3.x**{: .eol }](/{{page.lang}}/3x/api.html) | October 2012 | July 2015 |
-| [**v2.x**{: .eol }](/{{page.lang}}/2x/) | March 2011 | July 2012 |
-| **v1.x**{: .eol } | December 2010 | March 2011 |
-| **v0.14.x**{: .eol } | December 2010 | December 2010 |
+| Major Version | Minimum Node.js Version | Support Start Date | Support End Date |
+| -- | -- | -- | -- |
+| [**v5.x**{: .supported }](/{{page.lang}}/5x/api.html) | 18.0.0 | September 2024 | **ongoing**{: .supported } |
```
ctcpip: let's just say `18` and omit the `.0.0`

ctcpip (approved):
good idea, I support this addition.  I have not confirmed the actual version numbers for EOL, but LGTM

UlisesGascon (approved):

de/support/index.md
wesleytodd: Is it funny to anyone else that we called this translated but it is not? Why even have this copy instead of just falling back to the english version if it is just english anyway?

wesleytodd (approved):

de/support/index.md
bjohansebas: This was done this way because it's how translations work on the page. If we redirected to the English page, we would make the user click again to return to the language they were in before. There is a lot of improvement in the translation process.

wesleytodd: Makes sense, and I agree this would be awesome to improve upon.

bjohansebas: The easiest way is to migrate to a more modern technology like starlight or docusaurus, which have better support for i18n and we wouldn't have to be copying the content like we're doing now, or a slightly more complicated approach would be to have the translations done by an AI.

bjohansebas: fixed

bjohansebas (approved):

de/support/index.md
wesleytodd: docusaurus gets my vote.

Phillip9587: There are some very promising tools out there to translate the content with AI. If you are interested take a look at for example: https://github.com/midday-ai/languine

bjohansebas: @Phillip9587, there is an [issue](https://github.com/expressjs/expressjs.com/issues/1676) about that, in case you want to participate, 

and for the website migration, there is this [issue](https://github.com/expressjs/discussions/issues/211).

bjohansebas:
@Phillip9587 thanks

Issue #1736, opened 2025-01-21
Improving the tooling/framework used for the doc (Jekyll currently)


bjohansebas:
I’ve been talking with @crandmck about migrating the website to a newer technology that can save us from unnecessary work (for example, copying and pasting content to make it available in other languages, improving accessibility, JavaScript logic that doesn’t add much value to the documentation content, styles like text line spacing, etc.). This is work that can very likely be handled by the framework we choose to use.  

While the focus so far has been on the content rather than the technical aspects, I think it’s time to make this change so we can finally focus solely on the content.  

The important thing here is to choose a tool that is strong in several aspects, such as:  

- i18n  
- Accessibility  
- Ease of creating/modifying content  
- Preferably good Markdown support  

Possible technologies that could be used for this change:  

- [Docusaurus](https://docusaurus.io/)  
- [Astro Starlight](https://starlight.astro.build/)  
- [Hugo](https://gohugo.io/documentation/)  
- [Eleventy](https://www.11ty.dev/)
- [VitePress](https://vitepress.dev/)

With all this in mind, I’d like to invite @expressjs/express-tc, @expressjs/docs-wg, and all members of this community to help us choose a good tool for the documentation.

UlisesGascon:
This is a bold move, and I think it's a good moment to do it 👏 👏.

I have only a small amount of experience with Hugo and Eleventy. I’ve never used Astro, Starlight, or VitePress... but I am a great fan of Docusaurus, mainly because it is widely adopted by other projects in the foundation, such as Jest. We can learn a lot from their base code (https://github.com/jestjs/jest/tree/main/website), as they also support multiple versions of their product. This is something we should consider, like express@3, express@4, and express@5... as a requirement when doing the decision.

That said, I’m not very opinionated on the framework used.

carlosstenzel:
I like the idea of using Docusaurus too

chrisdel101:
I honestly don’t have any experience with these documentation generators (if that’s the word for them) but they all look good. I think Jekyll I works okay, but is quite inefficient and i also dislike the copy paste ideology and would love to get rid of it!
I guess I wonder how we even start this? Would we work on it along side the prod site until it’s ready?
It’s a good and exciting discussion to have

joeyguerra:
[...]

PR #1738, opened 2025-01-23
Feat/poc translate by json - POC - NOT MERGE

carlosstenzel:
Translation test using JSON, instead of copying files to translate.

## modifeld files
_includes/head.html
_includes/header/header.html
_layouts/page.html
assets/js/translations.js
i18n/header.json
i18n/home.json
index.md

## Test pages: 
 header - menu - item (home) 
https://deploy-preview-1738--expressjscom-preview.netlify.app/en/starter/installing
https://deploy-preview-1738--expressjscom-preview.netlify.app/pt-br/starter/installing

description only home in EN


carlosstenzel:
@bjohansebas @chrisdel101 
I am testing a less aggressive approach to changes for translations. It needs many adjustments, but it's a start. What do you think?

ref: https://github.com/expressjs/expressjs.com/issues/1736

bjohansebas:
BTW, the change of the tool is not only for i18n, it is also to avoid having to focus too much on design/accessibility.

Thanks for creating this POC, the way it's working would create layout shifts, which is a bad user experience and Google could/may penalize us for that.

PR #1740, opened 2025-01-27
docs: remove express v2

### LINKED ISSUE #1556: Remove express v2 docs

bjohansebas:
closes #1556


wesleytodd (approved):
Seems good, but as we discussed in the issue, it would be nice to have some docs on how to access them in case someone is looking.

UlisesGascon (approved):

chrisdel101:
Do you still require another reviewer? Or are you good to go?

bjohansebas:
@chrisdel101 I am making some changes to set up the redirects

bjohansebas:
I have applied the redirects on the pages that were from Express v2 and modified the 2x branch so that only version 2 is there. Additionally, in the README, I included some instructions on how to preview the content.

By the way, @expressjs/express-tc can you add protection to the 2x branch so that it can only be read and not modified?

chrisdel101 (approved):
LGTM. 
The only thing is whether we want the links to 2x branch left in on version support page, since it might be hard to find the 2x branch, but I assume you removed those to discourage people from even going there.

jonchurch:
For a big PR like this, prefer to push regular commits and then squash at merge. That way during review it's simpler to know what has changed since last review. Based on your comment @bjohansebas the latest force push update was the redirects, but it's hard to tell with this many changed files

jonchurch (approved):
We can't actually get a true 301 out of Jekyll and Github pages here (or any static site builder) for what it's worth. Will be a 200 to load the page from CDN then a meta redir using JS (or the meta-http tag) on the loaded page itself.

Im okay with that for now, since we are also removing the Nav link from the site itself. 

To fix that we'd need a page rule at Cloudflare, which requires additional config that Im not really convinced we want to maintain there.

jonchurch:
I've enabled branch protections on the 2x branch (via rulesets, for anyone in the future looking for this in github settings)
image

bjohansebas:
Yep, it's not the redirection I would like to have, but it's the only one allowed by GitHub Pages and Jekyll.

Issue #1628, opened 2024-09-21
Set the design system color


bjohansebas:
It would be great to start using CSS variables to maintain consistency in colors. I've been trying to do this in previous PRs, but I think it's time to define it completely. This way, we could simplify dark mode and solve several issues with !important.

Inspired by how shadcn/ui handles colors, I'm going to leave a list of colors we could use:

## Light mode:

--bg: #ffffff (background)
--fg: #383838 (general texts)
--card-bg: #f0f1f3 (header, code blocks, announcement, etc)
--card-fg: #383838 
--hover-bg: #484848 (hover background)
--hover-fg: #fafafa (hover foreground)
--border: #d9e1e4 (borders)
--accent: #0e78ce (Links)

![image](https://github.com/user-attachments/assets/914939a5-f317-4a44-a086-232d11f158b3)

## Dark mode:

--bg: #0c0c0c
--fg: #fafafa (general texts)
--card: #181818 (header, code blocks, announcement, etc)
--card-foreground: #71717A
--hover-bg: #484848 (hover background)
--hover-fg: #fafafa (hover foreground)
--border: #d9e1e4
--accent: #2b8fe0 (Links, hover background)

![image](https://github.com/user-attachments/assets/2a34ebde-6842-4fc1-b4a8-5ffb1799ca9f)

cc/: @expressjs/docs-wg

chrisdel101:
Are we changing the colors too? For dark mode I pulled almost all the colors from github's dark mode. Are we changing darkmode colors for changes sake only? 
Or because we're not satisfied with the ones we have? If it's the first reason, I suggest we leave them. I could be convinced otherwise though if there is a reason :)

I agree with the rest of the stuff. Maybe we can integrate scss here too? If it's not too much work that is. Makes systems like this easier to maintain. Plain CSS is also fine if not.

If no one is working on this I can start.

bjohansebas:
It is not necessary to change all the colors of the dark mode, just to improve a bit the contrast between the colors. They look fine on a computer, but on mobile, they are almost the same (it also depends on the screen). However, if you don’t want to, that’s okay. What I do want is to put all the colors in css variables.

![image](https://github.com/user-attachments/assets/913c51e7-eaf4-401f-978f-fe1818860d8c)


> Maybe we can integrate scss here too? If it's not too much work that is. Makes systems like this easier to maintain. Plain css is also fine if not.

[...]

Issue #1726, opened 2025-01-12
Submenu navbar anchor active class does not get applied


chrisdel101:
The class `#navbar a.active` applies styles but the active class is not being applied correctly. In this example I believe the "Guide" should be set to active, not the "Getting Started"
![Screenshot 2025-01-12 at 10 16 20 AM](https://github.com/user-attachments/assets/699e9f1a-67cb-4e56-b913-7a1a44b32f5f)

chrisdel101:
I think is this actually operating as expected, as per the logic in the templates like:
`<a href="/{{ page.lang }}/starter/installing.html"{% if page.menu == 'starter' %} class=" "{% endif %}>Getting started</a>`

I think this menu would be better if the top level anchors became active when opened. The way it is now is not intuitive like the full size menu is.

PR #1742, opened 2025-02-03
Design system

chrisdel101:
Reallizes #1628.
- moving colors to variables. Attempting to make colors/names constant across themes.
- move language css to folder
- move themes to folder
- remove unused code them applicable
- add light mode theme

Left some comments in the codes to note why changes were made. These need deletion before merge. 
More commentary on reasons for changes given below.

preview: https://deploy-preview-1742--expressjscom-preview.netlify.app/


_includes/head.html
```diff
@@ -5,12 +5,13 @@
...
     <script data-cfasync="false" src="/js/theme.js"></script>
-    <link rel="stylesheet" href="/css/dark-theme.css?_={{ site.time | date: '%s' }}">
+    <link rel="stylesheet" href="/css/themes/dark-theme.css?_={{ site.time | date: '%s' }}">
+    <link rel="stylesheet" href="/css/themes/light-theme.css?_={{ site.time | date: '%s' }}">
     <link rel="stylesheet" href="/css/style.css?_={{ site.time | date: '%s' }}">
     <link rel="stylesheet" href="/css/prism.css">
     <link rel="stylesheet" href="/css/font-awesome.min.css">
     <link rel="stylesheet" href="//fonts.googleapis.com/css?family=Open+Sans:300,400,600,700&amp;amp;subset=latin,latin-ext">
-    <link rel="stylesheet" href="/css/{{ page.lang }}.css">
+    <link rel="stylesheet" href="/css/langs/{{ page.lang }}.css">
```
chrisdel101, line 16: Move these files into folders.

_layouts/home.html
```diff
@@ -3,7 +3,7 @@
  Copyright (c) 2016 StrongLoop, IBM, and Express Contributors
  License: MIT
 -->
-<html lang="{{ page.lang }}">
+<html class="light-mode" lang="{{ page.lang }}">
```
chrisdel101, line 2: Add light-mode by default. Some JS was added to add remove the class, along with dark-mode.

css/themes/dark-theme.css
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  #navmenu>li:hover {
+    & ul {
+      background-color: var(--card-bg);
+    }
+  }
+
+  /* overlay on dropdown menu  */
+  div#overlay {
```
chrisdel101: Changed the overlay to be more pronounced
1_dark_mode_overlay_color

css/themes/light-theme.css
```diff
@@ -0,0 +1,15 @@
+@import "../values.css";
+
+html.light-mode {
+    #top {
+        background: var(--box-fg);
+    }
+
+    #overlay {
```
chrisdel101: Add more pronounced overlay.
1_light_mode_overlay_color

css/style.css
```diff
@@ -320,72 +302,102 @@ header {
...
-  color: #555;
+  color:var(--card-fg);
+}
+/* styles sapply on non-en static-files.*/html [class*='language-'] { 
+  text-shadow: 0 1px var(--bg);
+}
+/* all pre tags instances have code children - get rid of importants - override prism white bg */
+pre,
+/* api docs code blocks */
+pre[class*="language-"]:has(code)[class*="language-"],
```
chrisdel101: - moved plain-text related items here - uz static-files is example
- below was before the fix
3_uz_pre_color_dark
3_uz_pre_color_light

bjohansebas:
Wow, it looks great, I'll review it very soon

_includes/head.html
bjohansebas: If we end up not migrating, i would like to delete all those files

bjohansebas:
@chrisdel101 i have fixed the conflicts

chrisdel101:
@bjohansebas It's still a draft as I'm working fixing a few of the last details, but you can start a review as the remaining things are minor. Overall it's complete, but left as draft due to that. 

- I tried not to make any actual design changes, other than colors, and move things to variables - but there were a few changes that had to happen.
  - redid the overlay as shown above.
  - redid the `#top` arrow icon (in bottom left) that scrolls to the top of the page - its more pronounce on scroll and on hover.
  - added dark mode 'announcements' icon on index.
  - remove rules that did not exist - used Beautiful soup to search code base for things, and if not found I removed them. I tried to note this where possible with comments.  
  - Changed the way the desktop dropdown menu works, and the coloring. Now the top level of the drawer is off when drawer is open and hovering over a child item, and a bottom border was added for clarity on the top level e.g. Guide. I also plan to submit a PR for #1726 as I think in this img Guide should be bolded, not Getting Stared (which is bold since it's the page were currently on)
 
image

- I'll delete all these "help" type comments that exist, which I left in to help with reviewing i.e. explaining why a change was made. 
- I tried to keep the number of colors to a minimum and so got rid of most of the colors that only occurred once or twice, and replaced with one of the system colors. It should be simple to make any changes to unliked colors, or add in new ones, using this same system. So, things do look different in places.
- In the values.css I noted which colors got replaced with which - this was mostly just helpful in dev for me, and I will delete these comments before merge.
I'll add anything else to this block as I complete my PR.

css/style.css
```diff
@@ -755,22 +753,25 @@ footer {
...
+}
 #blog-side-menu {
   list-style: none;
   padding: 0;
 }
 
-#menu em,
-#tags-side-menu em,
-#blog-side-menu em{
+#menu ul a.active {
```
chrisdel101, line 700: Can we remove `#tags-side-menu` and `tags-menu.md`? I'd love that!

bjohansebas: Yep, do it for now

css/themes/dark-theme.css
```diff
@@ -0,0 +1,80 @@
+@import "../values.css";
```
bjohansebas: Why would we include the import? Just adding it like other files should be enough, right?

_layouts/home.html
bjohansebas: Why not do it globally? I don't see the need to add it to just one class.

css/themes/dark-theme.css
bjohansebas: great

css/themes/light-theme.css
bjohansebas: great

css/style.css
bjohansebas: Probably need to finish #1746 first, right?

css/values.css
```diff
@@ -0,0 +1,70 @@
+.light-mode:root {
+  /* [#7e7e7e, dadada] */
+  --grey-fg: #aeaeae; /* %68 */
```
bjohansebas: I see that we only use it for the description, we probably want to remove it and make all the text the same color, right?

bjohansebas:
We should still show when they are an inline code
image
image

_layouts/home.html
chrisdel101: The css is triggered by light-mode or dark-mode now, so it must have a theme class on the html. This just the default - open to change if it's possible. What did you have in mind?

css/themes/dark-theme.css
chrisdel101: REMOVED: Thought it was required to import to access vars, but seems it is not.

css/style.css
chrisdel101: Yeah, unless we want to update this after. Essentially all this would be removed then?

_layouts/home.html
bjohansebas:
```suggestion
<html lang="{{ page.lang }}">
```

Not adding default classes, but well, I see it's used elsewhere, so we would need to debug further. I'll be back soon, I'll keep removing prims from the page.

css/style.css
bjohansebas: Part of that will be removed

_layouts/home.html
chrisdel101: We need someway to trigger the themes. It's pretty well based around this setup now, and is what allows the vars to work mostly in one file - style.css. Don't know it can be unwound - unless you see a way. 
Anyway, it's just a single class so the overhead is very low. And the JS used is very simple, add remove class stuff

bjohansebas: Yep, I can live with that.

css/style.css
bjohansebas: I've finished the [PR](https://github.com/expressjs/expressjs.com/pull/1746), maybe you want to take a look so we can move forward with it.

_layouts/home.html
chrisdel101: I'm testing it now that you say this to make sure. It might be added by the JS already so this class is _not_ needed. If this is the case then yes I'll remove it. Just need to test to make sure.
I removed the class and it still works, so...

css/style.css
chrisdel101: Yes will review it. Will require some significant changes over here after that, which is fine.

_layouts/home.html
chrisdel101: Removed

css/values.css
chrisdel101: removed - changed to menu-grey.

css/style.css
chrisdel101: Done

css/style.css
```diff
@@ -200,35 +193,35 @@ li code {
...
+  background-color: opacity(var(--box-fg), 0.145);
   width: 16px;
 }
 
 *::-webkit-scrollbar-thumb {
   border-radius: 8px;
   border: 4px solid transparent;
   background-clip: content-box;
-  background-color: hsl(0, 0%, 67%);
+  background-color: var(--grey-fg);
```
chrisdel101: Just make sure to check scroll bar and hover.

chrisdel101:
> We should still show when they are an inline code image image

I looked compared the blog page to prod and I don't see any visible difference besides some small line height spacing. Is this still and issue?
If yes, can you be more specific?

chrisdel101:
~~I noticed a problem with the yaml rendering here. Wasn't sure the best way to fix.~~ I changed it to `markdown` and it looks good now.
image
And there spaces at the start of lines. Not sure why. It's like this in prod too. Was it like this with Prism? 
space

carlosstenzel:
It is in every block of code, including the home page.
Captura de Tela 2025-02-06 às 13 09 56

we can see this padding, it seems to resolve

Captura de Tela 2025-02-06 às 13 10 52

carlosstenzel:
@chrisdel101 
 there is space in the code tag

chrisdel101:
@carlosstenzel Where is the space? I can't see it.

carlosstenzel:
image

bjohansebas:
It wasn't intentional, although I already know how to fix it because I had already solved it before due to the same error. I recommend creating an issue instead of discussing it here, since this is not related to this PR.

chrisdel101:
Okay, isn't this the same thing I noted above and you told me to fix using padding?
I thought U meant something else.

carlosstenzel:
> It wasn't intentional, although I already know how to fix it because I had already solved it before due to the same error. I recommend creating an issue instead of discussing it here, since this is not related to this PR.

Sorry guys, we left this adjustment for another PR

css/style.css
```diff
@@ -1,27 +1,21 @@
...
-  --box-fg: rgb(24, 24, 24);
-  --info-bg: rgb(207, 212, 252);
-  --info-accent: rgb(15, 28, 138);
-  --notice-bg: rgb(252, 234, 197);
-  --notice-accent: rgb(115, 82, 13);
-  --warn-bg: rgb(250, 209, 223);
-  --warn-accent: rgb(138, 15, 58);
-  --code-bg: #f0f1f3;
-}
+@import "./values.css";
```
bjohansebas: Let's do this, another link tag in the head. I'm sure we shouldn't use import in CSS, it's considered bad practice if I remember correctly, unless we use a preprocessor like Sass.

css/style.css
```diff
@@ -1,27 +1,21 @@
-/* general */
+/*
+ Copyright (c) 2016 Express Contributors
+ License: MIT
+*/
```
bjohansebas: please remove this

css/style.css
```diff
@@ -1321,6 +1289,9 @@ blockquote {
   #navbar .submenu.open {
     flex-direction: column;
     align-items: initial;
+    > a {
```
bjohansebas, line 1334:
```suggestion
    & a {

```

css/themes/light-theme.css
bjohansebas, line 1: Is there any reason why we can't delete this file and put those styles directly in `styles.css`, applying them directly to their respective tags?

```css
#top {
opacity: .2;
  line-height: 0;
  border-radius: 2px;
  position: fixed;
  bottom: 15px;
  // .....
  background: var(--box-fg);
  
}
```

css/values.css
```diff
@@ -0,0 +1,51 @@
+.light-mode:root {
```
bjohansebas: I meant setting those variables directly in `:root` to make them even more global.
```suggestion
:root {
```

chrisdel101: Those are scoped. Removing those classes means the styles are not applied to the right themes.
Unless you are just saying?

css/themes/light-theme.css
chrisdel101: Because they are specific to light mode. Not every single change could be binary, but i tried my best. There are a still a few that are totally seperate, like these. And they cannot be put in style since the entire style.css file is scoped to light or dark mode.

css/style.css
chrisdel101: But `>` is direct descendant and `&` is just descendant, no?
Seems to still work b/c there is only one descendant but > is what we really want I think.

css/themes/light-theme.css
chrisdel101: Sorry, I think the answer is actually yes we can do it for this one actually.

EDIT: Wait, no, only light mode has a dark background where at the dark-mode does not (it looks bad just flipping)  - so after all it cannot be moved. 
Screenshot 2025-02-06 at 8 17 00 PM
Screenshot 2025-02-06 at 8 16 44 PM

css/style.css
chrisdel101: It's broken or something atm. Will need to check this now. Not sure what happened.

css/style.css
bjohansebas: Ah, okay, I thought it was a typo.

css/style.css
bjohansebas: I think this is already resolved since it was completely fixed in another PR

css/style.css
chrisdel101: There was some sort of overlay on scroll. Must have been caused by some other PR as it appears to be fixed.

css/values.css
chrisdel101: Just to reiterate - root sets variables globally, and this just scopes it to have a class. I rearranged the order for clarity.

chrisdel101:
Hey @bjohansebas, I made the changes and left comments re: what I did not change. I don't think anything else can be moved rules-wise between files now.

I think I've moved over everything I could now into style.css

css/values.css
bjohansebas: So we would have to add the `light-mode` class to the `<html>` tag, as you initially suggested, since there may be users who disable JavaScript, and the page would look like this:
image

css/style.css
```diff
@@ -360,21 +334,26 @@ pre code {
 /* top button */
 
 .scroll #top {
-  opacity: .2;
+  border: 1px solid var(--border); 
+  opacity: .5;
+}
+
+#top:hover {
+  border: 2px solid var(--border);
```
bjohansebas:
```suggestion
  border: 1px solid var(--border); 
```

I would suggest not increasing the border size. The effect it creates when hovering and expanding doesn't look natural.

css/style.css
```diff
@@ -360,21 +334,26 @@ pre code {
...
+  opacity: .5;
+}
+
+#top:hover {
+  border: 2px solid var(--border); 
+  opacity: 1;
 }
 
 #top {
+  opacity: .2;
```
bjohansebas: If it's at the top of the page, we shouldn't show that element, which is why its opacity is set to 0.

bjohansebas:
Great job @chrisdel101! I've left the last two comments, and I can give my approval. btw, I updated the color of the code examples.

css/style.css
chrisdel101: Removed it.
I can't remember why I would have added this. Maybe it was something I forgot to remove?

css/style.css
chrisdel101: Undid the changes back to 1px

css/values.css
chrisdel101: Put the light-mode tag back in.

chrisdel101:
@bjohansebas Hey, I made the newest set of requested changes. I added a few comments to the theme JS file here for clarity, but did not change any JS but I changed the names of a few functions (only the names).
Let me know if you find anything else. Thanks for all the help with this review!

bjohansebas (approved):
LGTM, I'm going to wait a day so more people can review it

chrisdel101:
@bjohansebas Just a heads up, I'm going to be away for awhile for medical reasons - maybe like a month. If this branch gets out of date you might need to take over the updates :)

carlosstenzel (approved):

bjohansebas:
@chrisdel101 Great job and take care!

chrisdel101:
@bjohansebas Once again thanks for all your help on this, and for the thorough review.

PR #1774, opened 2025-02-07
headers and footer in crowdin

bjohansebas:
The texts for the menus and the footer also need to be uploaded.

cc: @ctcpip

### DIFF
--- a/crowdin.yml
+++ b/crowdin.yml
@@ -1,5 +1,7 @@
 files:
   - source: /en/**/*.*
     translation: /%two_letters_code%/**/%original_file_name%
+    ignore:
+      - /en/blog/**/*.*
   - source: /_data/en/**/*.*
     translation: /_data/%two_letters_code%/**/%original_file_name%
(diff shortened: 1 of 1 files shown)


bjohansebas:
We need to ignore the blog for now, there are things we need to do on our side for the blogs to be translated.

ctcpip:
@bjohansebas thanks!  this file is actually managed by the GH integration.  (I am not manually editing this file and pushing it up).  but I have made the changes in crowdin, so they should get pushed via the integration now

ctcpip:
https://github.com/expressjs/expressjs.com/commit/b56a8e61ee402c7981be15248983103fe4ca28f4

bjohansebas:
I would have preferred to handle the entire Crowdin integration with GitHub Actions, as it allows the documentation team to have greater control. I can't edit the configuration in Crowdin, and I don't want to have to message someone from TC every time I need to modify that file for some reason.

Issue #1614, opened 2024-09-13
Sidebar menu space issue


bjohansebas:
The sidebar menu used in the API takes up more space, making it impossible to use the link to the terms of use and privacy policies.

![image](https://github.com/user-attachments/assets/c833c4bb-e6da-47cc-b974-5666ca453a11)

devgar:
There's a jQuery bind managing this menu size, as it is intention to remove all jQuery #1569, to fix this issue Instead of VanillaJS I propose to use CSS. I've experimented a bit with calculated height and scroll animation

https://github.com/user-attachments/assets/cf6dfca7-7b81-4891-93cd-f286bf27f0da

bjohansebas:
@devgar oh great, can you open a PR? So we can test it.

ShubhamOulkar:
Hi @bjohansebas this issue is resolved. So close it.

PR #1767, opened 2025-02-05
ci: deploy website with github actions

bjohansebas:
Deploy with GitHub Actions to better manage our deployment environment, plus now having all three environments with the same Ruby version. This is part of the work for https://github.com/expressjs/expressjs.com/issues/1747

PR #1775, opened 2025-02-07
Feat : Add Table of Contents Component on pages

### LINKED ISSUE #1760: add table of contents
We should add a table of contents on each page to make it easier to navigate the content.

I would suggest placing the table on the right side, as several site generators do, like Docusaurus or Astro Starlight, and probably adding #1759 at the top or bottom.

ShubhamOulkar:
## task 
- [X] design table of content (TOC) component
- [X] responsive TOC btn and sub menu list
- [X] TOC dark theme
- [X] solved stacking context layout issue #1614
- [X] accessibility wcag 2.1 guidelines
- [X] reconstruct content layout design 

## toc added on following pages 
- [X] 4x api 
- [X] 5x api
- [X] 3x api
- [X] [Blogs](https://deploy-preview-1775--expressjscom-preview.netlify.app/2025/01/09/rewind-2024-triumphs-and-2025-vision)  
- [X] [resources/middleware](https://deploy-preview-1775--expressjscom-preview.netlify.app/en/resources/middleware)
- [X] [changelog](https://deploy-preview-1775--expressjscom-preview.netlify.app/en/changelog/)

 | mobile light mode | mobile  dark mode | desktop | 
|------------|-------------|------------|
|Screenshot 2025-02-21 033329|Screenshot 2025-02-21 033439|Screenshot 2025-02-21 033556|

Resolves #1760, #1614, #1807

#### Edits 

To stop flashing on tap, `-webkit-tap-highlight-color : transparent ` is applied. Header mobile dropdown menus might need :focus, :active class selectors. This chrome specific issue.

|before|after|
|-------|------|
|XRecorder_20250221_01_001|XRecorder_20250221_02_001-ezgif com-video-to-gif-converter|


bjohansebas (changes requested):
Thanks for opening this PR.  

- Please don't increase the font size in this PR, that's a separate issue that requires a dedicated PR, which distracts from the main focus of this one.  
- Also, keep the original style for desktop for now. The design you propose looks great for mobile, though some styles need to be adjusted.  
- Please don't format the code in areas that are outside the scope of this PR.  

Most importantly, we should find a way for Jekyll to generate that table. Moving it to another file might not be the best approach since I'm working on making the API translatable in Crowdin, and managing translations in a separate file would be more complicated. 

I'd even be open to using JavaScript to generate the table.  

These are my main comments for now. I'll try to review it more thoroughly tomorrow or the day after, but I'm really tired today.

chrisdel101:
Hi @ShubhamOulkar, Thanks for the PR. I was just looking at this and have some concerns. 
- The menu file should not be removed like this. It seems to me that this is already essentially a kind of TOC for the API pages, but for larger screens only. 
- So you'd really be adding this to screens where the menu is not visible, and mobile. I like the design.
- So the TOC menu should operate and display the dropdowns _in the same way as the current menu does_.  So when you click it opens the sublinks. 
- The current JS can likely be reused for that. It might be possible to just use the same menu or copy and rejig it - up to you. Seems unnecessary to rewrite this, unless that was part of the original issue.
- why only on api V4x? If it's a simple includes I see no reason not to add everywhere it's going to be used.
-  As Sebastian says please remove all changes that are only formatting.

ShubhamOulkar:
@bjohansebas, Thank for your kind suggestions. Now it is ready for review. 

1. accessibility improvements removed: font size and line spacing removed.
2. removed 4x api menu data .yml file.

bjohansebas:
This button should be visible on smaller screens, around 800px, or when the layout actually breaks, not on 1200px screens.
Image

js/menu.js
```diff
@@ -109,10 +113,101 @@ overlay?.addEventListener("click", () => {
...
+  let options = {
+    root: null, // Observe relative to viewport
+    rootMargin: "-57px 0px 0px 0px", // Slight offset to ensure intersection triggers correctly
+    threshold: 0, // Trigger when the element is fully out of view (i.e. behind header)
+  };
+
+  observer = new IntersectionObserver(handleIntersect, options);
+  // observe intersection of TOC btn with header bar
+  observer.observe(firstHeader);
+}
```
bjohansebas: Is it necessary to have an observer? I don't see why we should use it—the less JavaScript, the better. If we can use the menu logic instead, that would be better since this behaves like a menu

We could even use the `<details>` element for this case. Of course, I'll leave it up to you, but the less JavaScript, the better.

bjohansebas:
It shouldn't create that element
image

js/menu.js
chrisdel101: I agree. Adapt or reusing the menu logic should theoretically get you pretty far, and save you time writing new JS.

ShubhamOulkar: > Is it necessary to have an observer?

No, it's not strictly necessary. My main thought was that an asynchronous scroll event is more efficient when many long API pages are open. Since developers usually have multiple pages open in their browsers, this helps with performance. Also, the pages already use jQuery scroll, so I expected a suggestion to shift to an asynchronous scroll approach.

> I don't see why we should use it—the less JavaScript, the better.

The page already uses jQuery for network calls to fetch the JS API. Instead of focusing on writing less JavaScript, we should focus on writing efficient JavaScript. That’s my perspective.

>  If we can use the menu logic instead, that would be better since this behaves like a menu.

The open/close logic for #navmenu is already implemented. The only issue is when the TOC is open, and the user clicks on the nav menu or language picker. These cases are handled within the same logic. Let me know if anything is missing. 

> We could even use the `<details>` element for this case. Of course, I'll leave it up to you, but the less JavaScript, the better.

I’d prefer not to rewrite the existing HTML (since @bjohansebas is working on generating tables in another PR). Also, using <details> would require new CSS for open/close states, and we'd still need JavaScript to handle conflicts with header buttons.

ShubhamOulkar:
Thank you @bjohansebas @chrisdel101
@bjohansebas removed observer api as you suggested. 
@chrisdel101 I adapted to code logic in my recent commits please review it.
Let me know if we can do it more better.

bjohansebas:
I don't think we should display that button yet when scrolling, or we should integrate it better. At first glance, it's not clear what the button does after scrolling.

image  

By the way, fix the conflicts.

bjohansebas:
An example of how it could be better integrated into the page
Close:
image

Open:
image

ShubhamOulkar:
> An example of how it could be better integrated into the page


I think you didn't check max-width=540px  break point. small screen width integrated btn appears below header. This floating btn appears in between 540 to 800 wide screens on scroll. Actually I don't like that arrow, I am searching proper icon for this. Right now only btn title appears on the screen.

bjohansebas:
I would prefer to display the menu as it is on small screens instead of showing that floating menu.

What I mean is to always display this menu until the desktop menu is available

image

bjohansebas:
Lol, these menus have been outdated for a long time. I'm going to work with you to fix them if that sounds good to you.
image

ShubhamOulkar:
> Lol, these menus have been outdated for a long time. I'm going to work with you to fix them if that sounds good to you. 

yup, I know this #1614. This is problem with layout. class content grows as page grows with 153px 3% 7% margins. And header has height 57px and footer 110px height. And this sidebar is positioned relative browser window. Position property take out element from normal document flow and creates new stacking context. That why it gets harder to position this sidebar when we reach bottom of the page. 

I tried some solutions to fix this like using flexbox on class content but due to position property it does not participate in flexbox layout. Then come to conclusion that rewrite layout css or change position when reach bottom on scroll or put spacer class when we reach bottom of page like I put in following image(last two options need JS).

image

I will choose to rewrite layout css because it is easier to maintain in future also we can remove unnecessary css. @bjohansebas, What do you think? 

[check sidebar from typescript doc page](https://www.typescriptlang.org/docs/handbook/utility-types.html#noinfertype). Sidebar is not responsive but we can refer big screen layout of content and sidebar. 

After layout changes problem is fixed 👇🏻

sidebar-fix
`Note :- Update jQuery animation of active links.`

bjohansebas:
I have found this Ruby plugin to use with Jekyll, which would generate the table automatically.
(https://rubygems.org/gems/jekyll-toc/)

Edit: I have found something better, which will make it so we don't need another plugin. (https://kramdown.gettalong.org/converter/html.html#toc)
Edit: It works! I still need to test a few things to make it global, but this is great. 
image


> I will choose to rewrite layout css because it is easier to maintain in future also we can remove unnecessary css. @bjohansebas, What do you think?

That's why I want to do this #1786, to better manage our styles since, at the moment, many things are merged with things that shouldn't be merged. Also, the things we are not using are not from this PR. It's better to make PRs focused on a single thing to avoid surprises in the future.

ShubhamOulkar:
> I would prefer to display the menu as it is on small screens instead of showing that floating menu.
> 

Ok, all documentation website use non floating menus. I put this because page is long and user need scroll to the top to access toc. I will do it after layout is fixed.

ShubhamOulkar:
> I have found this Ruby plugin to use with Jekyll, which would generate the table automatically.
> (https://rubygems.org/gems/jekyll-toc/)
> 
> Edit: I have found something better, which will make it so we don't need another plugin. (https://kramdown.gettalong.org/converter/html.html#toc)
 

Can you put testing pr for this? So that we can test and discuss there.

ShubhamOulkar:
> > Edit: I have found something better, which will make it so we don't need another plugin. (https://kramdown.gettalong.org/converter/html.html#toc)
> 

Can you put testing pr for this? So that we can test and discuss there.

bjohansebas:
> Can you put testing pr for this? So that we can test and discuss there.

I was thinking of testing it here to fix the layout, which was the main idea of the issue—to add it to most pages of the documentation (for now only in English, i18n will be fixed later).

Edit: I'm not going to do it in this PR. Let's fix the structure then.

bjohansebas:
Okay, we can't use new Jekyll plugins until we have control over our production environment (see #1767). I do have an idea on how to do it without a plugin for now, so I'll work on that.

bjohansebas:
I've made some changes that bring us closer to having those menus not completely broken. I think we're on the right track. The changes aren't complete, but you can review them and see what's missing. We also need to fix the colors

ShubhamOulkar:
now we can reuse same toc on all pages 🚀, checkout [page](https://deploy-preview-1775--expressjscom-preview.netlify.app/en/advanced/best-practice-performance).

bjohansebas:
The other pages are broken, and we really don't want to create the TOC manually. Do you want to create another PR to make it global across the other pages? Instead of here, I would like to launch this.

image

ShubhamOulkar:
> The other pages are broken

The reason is that I haven't updated the layout for these pages like I did for the advanced pages. All that needed is to add id=page-doc with class=no-sidebar. We should update the layout in this PR. I added a sidebar on just one page for testing purposes. I'm OK to remove it, but it is important that we must adjust our HTML to reflect the updated layout to prevent any issues when adding sidebars to other pages.

> I would like to launch this

Yet it is not ready. We need to fix accessibility issues. Minimum wcag 2.1 and 2.4 guidelines for  tabindex, focus, focus-visible, hover, etc. Also I am adding same page layout on resources/middleware and changelog pages. So that we can reduce some css classes.

chrisdel101:
I hadn't looked at this in awhile and it's looking quite a bit better! I noticed a few things.

- Did we need to move the menus to the other side of the page? Can't we do what we're doing but keep things on the same side they are on now for design consistency? Seems like a bigger redesign

- On the blog page this spacing looks asymmetrical to me. The `margin-right` is too little and looks weird. How prod is currently spaced looks better
- The menu should be centred in the gap. Try resizing the screen and making sure it looks good all the way through.

Screenshot 2025-02-20 at 1 12 47 AM

- On Mobile, did something happen to the page dimensions? Or is this just me? I don't see the issue in chrome so could be my page zoom, but side by side w prod the issue persists. Suggest viewing in FF or Safari for diffs. 
There should def not be any scroll happening here. 
Screenshot 2025-02-20 at 1 20 55 AM

- In the CSS we _really_ don't_ want to be adding new colors (thanks for using variables though) unless it's strictly necessary. We don't want one-off colors not part of the color scheme. I recommend seeing if you can try to make it work using colors that are already there. We just stripped away tons of colors and it looks good with only like 10 colors site wide. If we do add a color it would need to have a place in the overall color scheme outside of this single use case.

There are alot of file changes here. Are so many file changes necessary to add a TOC? 

Anyway, the functionality is looking nice.

ShubhamOulkar:
> * Did we need to move the menus to the other side of the page? 

 Do you mean right side? Then See reported issue by @bjohansebas #1760. 


> * On the blog page this spacing looks asymmetrical to me. The `margin-right` is too little and looks weird. How prod is currently spaced looks better

This is fixed in current accessibility improvements commit.


> * On Mobile, did something happen to the page dimensions? Or is this just me? I don't see the issue in chrome so could be my page zoom, but side by side w prod the issue persists. Suggest viewing in FF or Safari for diffs.
>   There should def not be any scroll happening here.

Yup, I know this issue. Horizontal scroll is added on mobile devices. I am doing this in accessibility improvements. 

> * In the CSS we _really_ don't_ want to be adding new colors (thanks for using variables though) unless it's strictly necessary.

This color variable is removed in my current accessibility improvements. 

> There are alot of file changes here. Are so many file changes necessary to add a TOC?

 We changed content layout to solve #1614.

ShubhamOulkar:
Hi @bjohansebas, now toc is in good shape. I think we can launch it very soon. Now TOC is accessible with keyboard, mouse also it is distinguishable by visuals. I have made changes to integrate TOC on the page as we discussed above.

bjohansebas:
Great, it's looking good! 

Ideally, the menu for the blog should be on the left, and on the rest of the pages, I think it should be on the right. What do you think?

chrisdel101:
Thanks for making the prev changes. Things look quite a bit nicer, esp the mobile view.

I still don't see that moving the menu to the other side of the page solves any problem. It seems like an unnecessary redesign since it's been on the left for a long time. Also it seems the convention for navigation panes to be on the left.
Was there a reason to move it? 

A bit of margin seems missing
Screenshot 2025-02-22 at 8 02 50 AM
Screenshot 2025-02-22 at 8 04 33 AM

bjohansebas:
@chrisdel101, the idea of placing it on the right works for #1780, so that the layout adapts. That's why I also say that the blog should stay on the left.

ShubhamOulkar:
@chrisdel101 , layout redesign is not for placing side bar on right or left. We could do with old positioning used in css  code just by changing position right value. Layout redesign happen because of the issue #1614 . Previously all components placed manually used positioning. So it gets hard to solve issue #1614 by using js. So I suggested @bjohansebas to redesign layout css using flexbox and removed positioning, unnecessary css classes. Now we can reuse all side bar html on all pages. This is main reason to redesign layout css not placing right or left. Placing right or left is teams choice. I prefer to think which side is more accessible? Who is user? How most of the user read? It is confusing and complex to think. 

About our discussion, I am user of this site for years. I navigated on api pages by using ctrl+f🤣. I really got annoyed. I have tested api pages and are really handy with right sidebar menus. About blogs pages right sidebar are annoyingly hard to read because menus are long with line breaks and words spacing. On blogs pages we can put it on left just because they are navigating to another page. And it is standard position on most websites. 

We must talk about what design we want? Then only we are able to integrate modern design. Now I did with minimalistic approach with existing styles and improved accessibility.

chrisdel101:
I prefer the left since I find if easier to read, think navigation pane in Word, and since it seems your functionality would work on either side, but I'm not requesting you change it. Seem Sebastian is okay with right side too and you had your reasons to put it on the right. 

Yeah I used ctrl-f too. Hopefully this helps us not use that as much ha.

ShubhamOulkar:
👋 @bjohansebas and @chrisdel101, Is it necessary to add underlines in sidebar menus? I prefer to add underlines only in docs because we need to distinguish links from the text content. And adding underlines or outlines on user interaction for sidebar menus. 

Note: some documentation websites use underlines only on links(sidebar links blogs, middleware's) that are navigate to another page. And links that are navigating on same page don't use underlines (api pages, changelogs').

Now we are using standard design for sidebar menus, same page navigation menu's on right and another page navigation menus on left. I want know ur thoughts. Right now I keep underline and outline for testing. Which one should be removed?

bjohansebas:
Good question, I think we could remove the underline in the TOC.
> some documentation websites use underlines only on links(sidebar links blogs, middleware's) that are navigate to another page. And links that are navigating on same page don't use underlines (api pages, changelogs').

That's a good idea, although I think we don't have any links pointing to other websites, at least in the menus.

ShubhamOulkar:
@bjohansebas, TOC is ready to launch 🚀. 

note : fixed #1807 because it is relevant to this pr.

bjohansebas:
Great, I will review it properly very soon.

bjohansebas (approved):
LGTM, There's still work to be done to fix the layout, but it will be in another PR. Thanks for the hard work!

PR #1783, opened 2025-02-10
docs: update documentation for res.type()

dpopp07:
Corrects information about the internal behavior that is no longer valid and re-adds documentation for the alias `res.contentType()`.

Resolves https://github.com/expressjs/express/issues/6330

### DIFF
--- a/_includes/api/en/5x/res-type.md
+++ b/_includes/api/en/5x/res-type.md
@@ -1,6 +1,6 @@
 <h3 id='res.type'>res.type(type)</h3>
 
-Sets the `Content-Type` HTTP header to the MIME type as determined by the specified `type`. If `type` contains the "/" character, then it sets the `Content-Type` to the exact value of `type`, otherwise it is assumed to be a file extension and the MIME type is looked up in a mapping using the `express.static.mime.lookup()` method.
+Sets the `Content-Type` HTTP header to the MIME type as determined by the specified `type`. If `type` contains the "/" character, then it sets the `Content-Type` to the exact value of `type`, otherwise it is assumed to be a file extension and the MIME type is looked up using the `contentType()` method of the `mime-types` package.
 
 ```js
 res.type('.html') // => 'text/html'
@@ -9,3 +9,5 @@ res.type('json') // => 'application/json'
 res.type('application/json') // => 'application/json'
 res.type('png') // => image/png:
 ```
+
+Aliased as `res.contentType(type)`.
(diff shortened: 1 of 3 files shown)


_includes/api/en/5x/res-type.md
```diff
@@ -9,3 +9,5 @@ res.type('json') // => 'application/json'
 res.type('application/json') // => 'application/json'
 res.type('png') // => image/png:
 ```
+
+Aliased as `res.contentType(type)`.
```
dpopp07, line 13: This may have been removed on purpose, in which case I'm happy to take it out of this PR.

It [was previously documented in v3](https://github.com/expressjs/expressjs.com/blob/gh-pages/_includes/api/en/3x/res-type.md?plain=1#L15) and [is still accurate](https://github.com/expressjs/express/blob/master/lib/response.js#L494), which is why I added it, but I understand if we'd prefer not to document it.

wesleytodd: Hm, I thought we had made an effort to deprecate the aliased things. Maybe this one got missed, but [it is for sure still present](https://github.com/expressjs/express/blob/master/lib/response.js#L494) so I don't see any reason not to document it.

wesleytodd (approved):

_includes/api/en/5x/res-type.md
bjohansebas: So this change is applicable to version 4, @dpopp07 do you want to include it?

dpopp07: Sure thing 👍 

I'll add that, as well as an update to the migration guide.

bjohansebas (approved):

_includes/api/en/5x/res-type.md
bjohansebas: thanks

en/guide/migrating-5.md
```diff
@@ -262,6 +263,20 @@ app.get('/user', (req, res) => {
...
+Use the [`mime-types` package](https://github.com/jshttp/mime-types) to work with MIME type values.
+
+```js
+// v4
+const mimeType = express.static.mime.lookup('json')
+
+// v5
+const mime = require('mime-types')
+const mimeType = mime.lookup('json')
+```
```
bjohansebas: Okay, the linter is failing here, please fix it.

dpopp07: Good catch, will do

Edit: fixed

bjohansebas: great

carlosstenzel (approved):

bjohansebas:
thanks @dpopp07

PR #1785, opened 2025-02-11
change menu icons to svg

bjohansebas:
I still need to fix some styles and change another icon. This is to remove Font Awesome.


bjohansebas:
close in favor of #1962

Issue #1780, opened 2025-02-09
Reduce the number of links in the menu.


bjohansebas:
What do you think about reducing the number of links in the menu? That is, keeping the following links:  

- **Docs** (for documentation)  
- **Support** (for the support page)  
- **Blog**  

![Image](https://github.com/user-attachments/assets/e88af0fe-3f4d-41ea-923e-74dd4f697807)

With this, other elements would also be adjusted to make it more compact, but it's just an initial idea.  

And throughout the documentation, we could display the following menu, similar to how other pages do (with our own information, of course):

![Image](https://github.com/user-attachments/assets/932e3f23-7501-43ee-8a8a-b645919fa937)

This would resolve #1704

chrisdel101:
I'm okay with this change and it gets rid of #1704 (my arch nemesis).
As a user, I prefer links over dropdowns.  Maybe we can keep a few more than three links? Small details aside, I'm okay and  like this idea

chrisdel101:
Hey @bjohansebas 
What do you think of this instead? It's a way easier solution than trying to combine the links or adding anothe dropdown. 

This is just a rough example version.

This is right before it breaks for mobile. Larger it could go back to single liners.

![Image](https://github.com/user-attachments/assets/3a7b2e30-b307-4118-856b-bb0d1b65dc29)

it's better if we remove the Home, which we don't really need since we have the logo

![Image](https://github.com/user-attachments/assets/db26398b-1899-4d13-bc9a-29a209dc36be)

PS - the language picker needs some identifier like a an arrow, or at least some border. I'm making an issue for that - which can be closed if others disagree.

bjohansebas:
Having the home link is already a convention on many websites, so it would feel strange to remove it.

> PS - the language picker needs some identifier like a an arrow, or at least some border. I'm making an issue for that - which can be closed if others disagree.

I would really prefer to always have the icon instead of the text, but if we don't make that change, it definitely needs an arrow.

chrisdel101:
It might still work with the Home if we don't want to delete that. I dont want to shrink the text as it's already quite small. What about a Home icon, like a house, with an alt tag? Or the svg w title?
If you're okay with the stacked text in general I can figure out the rest, even if I have to combine one link into two, rather than combine 3 or 4. 
If yes then we can close #1704.

[...]

PR #1806, opened 2025-02-22
fix: update middleware documentation

bjohansebas:
References to Express 3 are removed, and several community middleware that are not widely used are removed.

### DIFF
--- a/en/resources/middleware.md
+++ b/en/resources/middleware.md
@@ -13,25 +13,25 @@ module: mw-home
 The Express middleware modules listed here are maintained by the
 [Expressjs team](https://github.com/orgs/expressjs/people).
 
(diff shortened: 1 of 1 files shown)


chrisdel101:
The content looks good but the page looks a bit broken. On mobile it requires vertical scroll - not just here but in prod too. If that's another PR then I can approve this.
Screenshot 2025-02-22 at 5 43 02 PM

bjohansebas:
Yep, that's already a layout issue that I hope to resolve with https://github.com/expressjs/expressjs.com/pull/1775 and https://github.com/expressjs/expressjs.com/issues/1780

chrisdel101 (approved):

PR #1810, opened 2025-02-23
feat: normalize admoniations

bjohansebas:
Now we can translate the title of the admonitions and the text of the community caveat


_includes/admonitions/warning.html
```diff
@@ -1,4 +1,3 @@
...
+<div class="doc-box doc-warn" markdown="1"><p class="doc-title">{% include icons/warning.svg %} {{ site.data[page.lang].general.warning }}</p>
```
bjohansebas, line 1: What do you think about leaving the admonitions without a title, would the color be enough information?

carlosstenzel: I prefer to keep it to differentiate each alert box. Not everyone will know just by the color.

bjohansebas: agree

carlosstenzel (approved):

PR #1822, opened 2025-02-27
Feat:- Add GitHub edit btn

### LINKED ISSUE #1759: Show a button to edit the content
We should have a button or link to edit the content and redirect to GitHub. This would encourage people to edit and improve the content.

This was removed in #1548 when I redesigned the footer, but it actually should have never been removed.

ShubhamOulkar:
### Button is added on following pages
- [X] pages layout 
- [X] api layouts

## Design
| light mode | dark mode |
|-------------|----------|
|image|image|


_config.yml
```diff
@@ -36,4 +36,5 @@ exclude:
   - node_modules/
 
 
-github: [metadata]
```
bjohansebas, line 39: Do you know what this was for?

ShubhamOulkar: This line is removed for local development testing of btn. Do `site.github` object provide metadata only on Github pages? Any other way to see local developments as similar as production github pages, let me know. Or just remove github url once design is finalized.

carlosstenzel:
I like the idea, but I think the button should have a highlight and not only on :hover.

ShubhamOulkar:
@carlosstenzel , I am thinking about adding edit btn icon on api page for each section and sub section headers. What do you think?

bjohansebas:
> I am thinking about adding edit btn icon on api page for each section and sub section headers. What do you think? 

I don't think it's a good idea, we would have too many buttons for the API. I think it would make more sense for the link to lead to a folder where the API files are in GitHub.

css/style.css
```diff
@@ -549,6 +549,21 @@ html[xmlns] .clearfix {
...
 }
 
+a.edit-github-btn{
+  display: flex;
+  gap: 0.5rem;
+  width: fit-content;
+  padding: 0.5rem;
+  color: var(--info-accent) ;
+  border-radius: 0.3rem;
+  background-color: var(--info-bg);
```
bjohansebas: Please keep the colors that the page has, that color is only for the admonitions, it breaks the aesthetic of the page a bit.

ShubhamOulkar: Yeh, I don't like to use btn with underline and background color. Carlos suggested to add highlight not only on hover. What is the best approach? 

I prefer link with underline(most of documentation website use this approach). And if it is shown as btn then remove underline, but we need to investigate new color for buttons.

_config.yml
```diff
@@ -19,6 +19,7 @@ announcement: false # Enable or disable the announcements.
 
 plugins:
   - jekyll-redirect-from
+  - jekyll-github-metadata
```
bjohansebas, line 24: Although I understand the plugin, I can't run it locally with Docker.
image

It can be fixed by adding:
```yml
# _config.yml
repository: expressjs/expressjs.com
``` 
It would be the same as defining it:
```yml
# _config.yml
github:
  repository_url: "https://github.com/expressjs/expressjs.com"
``` 
So I’m going to remove the plugin, as it is unnecessary.

bjohansebas: Correction, we need part of the plugin, so I will keep it.

bjohansebas (approved):
LGTM, thanks for working on this!

PR #1824, opened 2025-03-03
docs: remove repeated section from performance guide

dpopp07:
I was reading through these docs and noticed a duplicated section at the end. This commit removes the duplicate.


bjohansebas (approved):
LGTM

PR #1833, opened 2025-03-12
Fix a11y in header button 

### LINKED ISSUE #1830: Change theme and language picker buttons are not accessible by keyboard
Image

ShubhamOulkar:
### Task 

- [X] Fixed keyboard accessibility for the toggle theme button.
- [X] Fixed keyboard accessibility for the language picker button.
- [X] Improved the design of the language picker.
- [X] removed duplicate id attributes 
- [X] rename [toggleSystemTheme to setTheme](https://github.com/expressjs/expressjs.com/pull/1833/commits/af2bd89b08c7ed80988856e7d795ee6e0b58162b)
- [X] fix [font url for thai and korean translations](https://github.com/expressjs/expressjs.com/pull/1833/commits/3bc366678fa4483dc237e7d98f76c8c5063b6437)
- [X] [fix bug: highlighting current language](https://github.com/expressjs/expressjs.com/pull/1833/commits/177a849ac20261616bf5a3bff56ecf91df6ba5b3)

### Closes #1830 

Tabbing navigation shown on landing page
theme-lang-fix

### New lang picker design
image

#### Note:- This PR specifically addresses keyboard navigation for the language picker and toggle theme button. Other navigation menus and submenus should also be fully accessible via keyboard but are not covered in this PR. However, the language picker on mobile devices currently has duplicate id attributes, which need to be removed, and CSS design should be update accordingly.


ShubhamOulkar:
Should we consider adding country flags to the language picker?

bjohansebas:
> Should we consider adding country flags to the language picker?

Hmm, I don’t know, if we add the flags as emojis, they might look different depending on the system. If we add them as SVGs, we’d have quite a few images. Isn’t just the name enough?

css/style.css
```diff
@@ -616,11 +616,83 @@ footer {
...
+    content: "";
+    display: block;
+    width: 0.8em;
+    height: 0.5em;
+    background-color: var(--card-fg);
+    clip-path: polygon(100% 0%, 0 0%, 50% 100%);
+    cursor: pointer;
+    pointer-events: none;
+    transition: transform 0.2s ease-in-out;
+  }
```
bjohansebas, line 652: And what if we add this as an SVG?

ShubhamOulkar: We can add an SVG using background-image: url() or an inline <svg> in HTML, but it adds extra rendering overhead. Since this triangle is purely decorative and doesn't need advanced styling like scaling, a CSS-only approach is faster and more efficient. If clip-path feels hard to maintain long-term, we can simply use content: "▼"; for clarity.

ShubhamOulkar:
> Isn’t just the name enough?

It is more than enough.

Should we add a title attribute in English to each language link? This is just a random thought that came to my mind while testing. All language names are translated into their respective scripts, and now I can't tell which one is Chinese, Korean, or Japanese. Also it improves accessibility for all users.

We don't need language-picker-mobile.html; use the same picker on mobile devices with some additional styling. Also, we can remove JS and CSS code.

_includes/header.html
```diff
@@ -242,10 +242,10 @@
     </ul>
   </div>
   <div class="header-right">
-    <div id="theme-icon-container" class="theme-toggle default-theme" title="toggle darkmode">
+    <button type="button" class="theme-btn" title="change theme">
       <i class="fa fa-moon-o fa-2x hidden-dark"></i>
       <span class="sun-icon hidden-light"></span>
```
ShubhamOulkar, line 253: Sun-icon needs to be svg

bjohansebas: yep, see #1785

js/theme.js
```diff
@@ -35,7 +35,7 @@ const themeWatcher = watchColorSchemeChange((colorScheme) => {
     document.addEventListener('DOMContentLoaded', () => {
 
       document
-        .querySelector('.theme-toggle')
+        .querySelector('.theme-btn')
```
bjohansebas: There should only be one theme toggle button per page, so in my opinion, the ID should be kept.

bjohansebas:
It's a good idea to move the styles to classes. The ID should never have been used for styling, but it's necessary for the JavaScript, so it should be kept.  

Make that last change, and we can merge this.

bjohansebas (approved):

PR #1835, opened 2025-03-15
docs: add deprecation note for 'back' redirect in res-redirect.md and res-location.md v4

### LINKED ISSUE #1834: Remove reference to `res.redirect('back')`.
### Description

In the documentation for the Express 5 API, we have a reference to `res.redirect('back')`, which was deprecated in version 4 and no longer exists in version 5.

Image

### Expectations

Maybe replace it with the recommended approach from the changelog or remove that part.

bjohansebas:
closes https://github.com/expressjs/expressjs.com/issues/1834

### DIFF
--- a/_includes/api/en/5x/res-location.md
+++ b/_includes/api/en/5x/res-location.md
@@ -5,13 +5,7 @@ Sets the response `Location` HTTP header to the specified `path` parameter.
 ```js
 res.location('/foo/bar')
 res.location('http://example.com')
-res.location('back')
 ```
-
-A `path` value of "back" has a special meaning, it refers to the URL specified in the `Referer` header of the request. If the `Referer` header was not specified, it refers to "/".
-
-See also [Security best practices: Prevent open redirect 
-vulnerabilities](http://expressjs.com/en/advanced/best-practice-security.html#prevent-open-redirects). 
 
 <div class='doc-box doc-warn' markdown="1">
 After encoding the URL, if not encoded already, Express passes the specified URL to the browser in the `Location` header,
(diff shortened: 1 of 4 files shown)


_includes/api/en/5x/res-redirect.md
```diff
@@ -45,12 +45,5 @@ Path-relative redirects are also possible. If you were on
 res.redirect('..')
 ```
 
-A `back` redirection redirects the request back to the [referer](http://en.wikipedia.org/wiki/HTTP_referer),
-defaulting to `/` when the referer is missing.
-
-```js
-res.redirect('back')
-```
-
```
bjohansebas, line 54: I'm not sure about removing this, maybe the recommended option should be mentioned `req.get('Referrer') || '/'`

ShubhamOulkar: Similar to 5x migration guide

carlosstenzel (approved):

_includes/api/en/5x/res-location.md
```diff
@@ -5,14 +5,8 @@ Sets the response `Location` HTTP header to the specified `path` parameter.
...
 res.location('/foo/bar')
 res.location('http://example.com')
-res.location('back')
 ```
 
-A `path` value of "back" has a special meaning, it refers to the URL specified in the `Referer` header of the request. If the `Referer` header was not specified, it refers to "/".
-
-See also [Security best practices: Prevent open redirect 
-vulnerabilities](http://expressjs.com/en/advanced/best-practice-security.html#prevent-open-redirects). 
-
```
krzysdz, line 15: Is there a reason why this note was removed here?

bjohansebas: we can keep it—there's no reason to remove it. Also, we still need to add the reference to the new recommended option (see https://github.com/expressjs/expressjs.com/pull/1835/files#r1997311521).

PR #1836, opened 2025-03-15
build: use stylelint

bjohansebas:
I would like to propose using Stylelint to have a linter for CSS. This would be part of #1596.  

What do you think, @expressjs/docs-wg, @ShubhamOulkar, @ipreencekmr?


package.json
```diff
@@ -3,15 +3,20 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
-    "test": "eslint --ignore-path .gitignore --ignore-pattern _includes/readmes/ \"**/*.md\""
+    "lint:md": "eslint --ignore-path .gitignore --ignore-pattern _includes/readmes/ \"**/*.md\"",
+    "lint:css": "stylelint \"**/*.css\" --allow-empty-input --cache --cache-strategy=content --cache-location=.stylelintcache",
+    "lint": "npm run lint:md && npm run lint:css"
   },
   "devDependencies": {
     "eslint": "^8.54.0",
```
ShubhamOulkar, line 11: 8.x end of life

package.json
```diff
@@ -3,15 +3,20 @@
...
     "eslint-plugin-import": "^2.29.0",
     "eslint-plugin-markdown": "^2.2.1",
     "eslint-plugin-n": "^15.0.0",
-    "eslint-plugin-promise": "^6.0.0"
+    "eslint-plugin-promise": "^6.0.0",
+    "stylelint": "16.15.0",
+    "stylelint-config-standard": "37.0.0",
+    "stylelint-order": "6.0.4"
   },
   "eslintConfig": {
```
ShubhamOulkar, line 21: we should put in separate configuration file

.stylelintrc.mjs
```diff
@@ -0,0 +1,12 @@
...
+export default {
+  extends: ['stylelint-config-standard'],
+  plugins: ['stylelint-order'],
+  rules: {
+    // Enforces the order of the CSS properties to be in alphabetical order
+    'order/properties-alphabetical-order': true,
+    'no-descending-specificity': null,
+    // Disables the Level-4 Media Queries; Since they're more exotic and less known
+    'media-feature-range-notation': 'prefix',
+  },
```
ShubhamOulkar, line 11: consider adding following plugins
https://www.npmjs.com/package/stylelint-high-performance-animation
https://www.npmjs.com/package/stylelint-declaration-block-no-ignored-properties

Also plugin "plugin/use-defensive-css": [true, { "severity": "warning" , "accidental-hover": true, "background-repeat": true  }] is useful in our code.

package.json
bjohansebas: Yep, although that's for JavaScript, so it doesn't belong in this PR. However, I would also like to update it in the future.

package.json
bjohansebas: Yep, in another PR

.stylelintrc.mjs
bjohansebas: Cool, that will help us maintain better CSS in this repository

ipreencekmr: looks good to me, can we also implement a [commit lint](https://commitlint.js.org/) along with style lint to maintain a consistent commit message format.

carlosstenzel (approved):

.stylelintrc.mjs
bjohansebas: The commit lint not yet, there's no consensus on that since there are limitations in the configuration (see https://github.com/expressjs/discussions/issues/293#issuecomment-2440074415, which mentions part of that, and I'm sure there are more issues and PRs that have tried to implement it). 

Step by step.

bjohansebas:
Next week, I will look into integrating the suggestions and merging this.

bjohansebas:
I prefer to do this when we start the redesign, for now, let’s continue like this without a linter.

PR #1838, opened 2025-03-19
feat: generate sitemap.xml for SEO indexing

### LINKED ISSUE #1762: We should generate a sitemap.xml
We should generate a sitemap.xml so that search engines can discover new pages faster or index those that have not yet been indexed.

ref: https://github.com/tomjoht/documentation-theme-jekyll/blob/gh-pages/sitemap.xml

We might need to separate it by languages, although it should be researched. We can see how other sites with i18n handle it.

ShubhamOulkar:
closes #1762 
## how to test sitemap.xml?

1. Visit [hreflang tool](https://technicalseo.com/tools/hreflang/)
2. Test url 👉🏻 https://deploy-preview-1838--expressjscom-preview.netlify.app/sitemap.xml

## test results from tool 
1. Warning ⚠ Missing region independent link for Portuguese(pt), Chinese(zh)
2. total 561 url's added in sitemap.xml with hreflang tags (17 languages).

## Issue
The website contains inconsistent file formats in its changelog directory, resulting in incorrect alternate URLs generated in the sitemap.xml. Specifically, the following formats are observed: `
ko/changelog/4x.md, 
tr/changelog/4x.md, 
uk/CHANGELOG.md
`
## verify issue 
When selecting Korean (ko) as the language on the changelog page, the website returns a "Page Not Found" error due to the presence of 4x.md, which seems to be an incorrect file format in this context.

> Note: Now I removed all changelog pages from sitemap and manually added only `en/changelog/index.md` to ensure SEO.


sitemap.xml
```diff
@@ -0,0 +1,81 @@
...
+    <lastmod>{{ site.time | date_to_xmlschema }}</lastmod>
+  </url>
+  <url>
+    <loc>https://terms-of-use.openjsf.org/</loc>
+    <lastmod>{{ site.time | date_to_xmlschema }}</lastmod>
+  </url>
+  <url>
+    <loc>https://privacy-policy.openjsf.org/</loc>
+    <lastmod>{{ site.time | date_to_xmlschema }}</lastmod>
+  </url>
```
bjohansebas: Is this necessary? Wasn't the sitemap only for the pages we have on the website?

tr/changelog/4x.md
```diff
@@ -3,13 +3,17 @@ layout: page
 title: Express 4.x Sürüm Notları
 menu: changelog
 lang: tr
-description: Stay updated with the release change log for Express.js 4.x, detailing
+description:
+  Stay updated with the release change log for Express.js 4.x, detailing
   new features, bug fixes, and important changes across versions.
+sitemap: false
```
bjohansebas, line 8: I think the changelogs should be in the sitemap.

uk/CHANGELOG.md
```diff
@@ -1,7 +1,11 @@
...
+---
+
 18.03.2016 Додано сторінки для наступних лінків:
-* http://expressjs.com/uk/starter/static-files.html (у головному меню)
-* http://expressjs.com/uk/guide/writing-middleware.html (у головному меню)
-* http://expressjs.com/uk/guide/migrating-5.html (у головному меню)
-* http://expressjs.com/uk/advanced/best-practice-security.html (у головному меню)
-* http://expressjs.com/uk/advanced/best-practice-performance.html (у головному меню)
 
+- http://expressjs.com/uk/starter/static-files.html (у головному меню)
```
bjohansebas: Please don't make these style changes, we will do this with Crowdin.

bjohansebas: I thought you were bringing that formatting from the original in English. Don't make those kinds of changes in this PR.

ShubhamOulkar: I didn't notice that, forgot to disable text editor formatting. I will remove that. 😑

tr/changelog/4x.md
ShubhamOulkar: Yes, I removed because change log generate two many redirects in sitemap. Also it is not translated except for Korean. Some times it is confusing for robots. I will add single url for English page.

sitemap.xml
ShubhamOulkar: It is common practice to add footer links. It improves seo and user page navigation. Sometimes robots miss footer links on the page. This happens on big websites with lot of pages.

ShubhamOulkar:
Should we consider adding urls for YouTube video and images used on website?

sitemap.xml
```diff
@@ -0,0 +1,85 @@
...
+<?xml version="1.0" encoding="UTF-8"?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:xhtml="http://www.w3.org/1999/xhtml">
+{% for lang in site.data.languages %}
+  <url>
+    {% if lang.code == 'en' %}
+    <loc>{{site.url}}</loc>
+    {% else %}
+    <loc>{{site.url}}/{{ lang.code }}/</loc>
+    {% endif %}
+    <lastmod>{{ site.time | date_to_xmlschema }}</lastmod>
```
bjohansebas: Is this a good practice? We still don't have the option to specify when the file was last modified, something we should add in the future. But as you're proposing, it takes the date the website was built, which I'm not sure is a good practice for Google.

sitemap.xml
```diff
@@ -0,0 +1,85 @@
...
+  </url>
+{% endfor %}
+  <url>
+    <loc>https://nodejs.org/en/</loc>
+    <lastmod>{{ site.time | date_to_xmlschema }}</lastmod>
+  </url>
+  <url>
+    <loc>https://www.netlify.com/</loc>
+    <lastmod>{{ site.time | date_to_xmlschema }}</lastmod>
+  </url>
```
bjohansebas: I don't think we should add Netlify. I understand the others, as they are part of the foundation, but Netlify is not. 

And having bots detect Netlify, I don't think that's our priority.

bjohansebas:
> Should we consider adding urls for YouTube video and images used on website?

Yep, images are a good idea if the tags can be generated automatically like with the pages. If we have to maintain that list of images manually, it's preferably better not to add them.

YouTube videos, no, it would be something manual to maintain.

ShubhamOulkar:
> Yep, images are a good idea if the tags can be generated automatically like with the pages. If we have to maintain that list of images manually, it's preferably better not to add them.
> 
> YouTube videos, no, it would be something manual to maintain.

I am okay with adding them in future updates as the website progresses.

sitemap.xml
ShubhamOulkar: yup, may be help them in page ranking. I will remove it. 

should be add robot.txt to reference sitemap.xml? Jekyll itself doesn’t require a file, but it can be useful for controlling how search engines crawl site.

sitemap.xml
ShubhamOulkar: Agreed on it should be last modified time. And it is good practice recommend on [google developer docs](https://developers.google.com/search/docs/crawling-indexing/sitemaps/build-sitemap#additional-notes-about-xml-sitemaps). I think this page has full explanation. 
This is something I’m interested in working on. I want to use a custom plugin from https://github.com/michaelx/jekyll-last-modified

sitemap.xml
ShubhamOulkar: [link](https://developers.google.com/search/docs/crawling-indexing/sitemaps/overview) explains about external link in `You might need a sitemap` section.

bjohansebas: Yep, adding a robots.txt would be fine.  

But removing Netlify is also necessary.

sitemap.xml
bjohansebas: That plugin is a bit old, so it probably no longer works in the newer versions of Jekyll.

ShubhamOulkar: yup

bjohansebas:
Why are you blocking it? That label is to prevent merging it because it depends on another change elsewhere (for example, in another repository).  

But I don't understand it here, why are you adding it?  

We cannot merge until there is an approval from someone with commit access, so it is not necessary to block it.

ShubhamOulkar:
> Why are you blocking it? That label is to prevent merging it because it depends on another change elsewhere (for example, in another repository).

Sorry I didn't put message there. I think it is best to do this after translation and fixing layout issues. Right now I am prioritizing layout issues.

bjohansebas (approved):
LGTM

bjohansebas:
@ShubhamOulkar the description says not to merge, do you prefer to wait a bit before removing some languages, or go ahead anyway?

ShubhamOulkar:
Lets move on 👍

PR #1841, opened 2025-03-23
feat: add copy code btn

### LINKED ISSUE #1814: Feature Request: Add Copy Button for Code Blocks
**Problem Statement**
Currently, there is no copy button next to code snippets, which makes users to copy the code manually everytime.

Image

**Proposed Solution**
Adding a copy button to code blocks, allowing users to quickly copy code with a single click. This will improve usability.

**Additional Context**
Enhancement Request

ShubhamOulkar:
- [X] added new feature: copy code btn
- [X] a11y: wcag 2.1   
- [X] added copied text popup
- [x] reponsive design

### design 
image

### copy failed 
copy-failed

### test failed state by over writing `writeText()` in console as follows

> navigator.clipboard.writeText = ()=>{throw Error("copy code btn error")}


carlosstenzel (approved):

carlosstenzel:
Good job, thanks for the help ;)

css/style.css
```diff
@@ -352,6 +352,44 @@ pre code {
...
+
+  &:focus-within button {
+    display: block;
+  }
+}
+
+pre:has(code) button{
+  position: absolute;
+  top: 5px;
+  right: 5px;
```
bjohansebas, line 376:
```css
cursor: pointer;
```

bjohansebas (changes requested):
Should we show a small pop-up indicating that it was copied successfully? Below the button—I think I remember that this already exists in HTML, so we wouldn't need JavaScript. I'll look for it.

I don't think the color change is enough.

bjohansebas:
image

Something like this could be good to indicate that it was copied successfully or that something went wrong. This is from the ESLint page, by the way.

ShubhamOulkar:
> image

yup 👍🏻 

Edit :- I added on left side because it overflows over border of `<pre>` code box

css/style.css
```diff
@@ -346,12 +346,101 @@ pre {
...
+
+  &::after { 
+    background-color: var(--supported-fg);
+  }
+  
+  &::before {
+    font-size: 0.85rem;
+    position: absolute;
+    left: -58px;
+    content: "copied !";
```
```suggestion
    content: "copied!";

```

css/style.css
```diff
@@ -346,12 +346,102 @@ pre {
...
+  &:focus-within button {
+    display: flex;
+  }
+}
+
+pre:has(code) button {
+  position: absolute;
+  top: 5px;
+  right: 5px;
+  border: none;
```
ShubhamOulkar, line 377:
```suggestion
  border: none;
  z-index: 100;
```
 Add z index

css/style.css
```diff
@@ -346,12 +346,103 @@ pre {
...
+    &::after {
+      mask-size: 1rem;
+      width: 1rem;
+      height: 1rem;
+    }
+  }
+}
+
+pre:has(code) button.copied {
+  outline-color: var(--supported-fg);
```
bjohansebas, line 412: Why use an outline instead of a regular border?

css/style.css
```diff
@@ -346,12 +346,103 @@ pre {
...
+
+  &::after { 
+    background-color: var(--supported-fg);
+  }
+  
+  &::before {
+    font-size: 0.85rem;
+    position: absolute;
+    left: -58px;
+    content: "copied!";
```
bjohansebas, line 422: There could be cases where an error might occur, right? It would be good to show it.

js/copycode.js
```diff
@@ -0,0 +1,38 @@
...
+    await copyCode(block, button);
+  });
+
+});
+
+async function copyCode(block, button) {
+    const code = block.querySelector("code");
+    const text = code.innerText;
+  
+    await navigator.clipboard.writeText(text);
```
bjohansebas, line 37: Handle the error by showing that the copy operation failed

ShubhamOulkar: Yes, we should use try...catch since writeText() returns promise.

css/style.css
ShubhamOulkar: Great thanks for suggesting

css/style.css
ShubhamOulkar: Nice question!
The component follows user interaction steps. On hovering or focusing the code block, only the copy SVG icon is shown. When the user hovers over the icon itself, it displays a hover/focus outline. This ensures the user is intentionally clicking the copy button. After clicking, a "Copied!" text appears.

Why not use borders?
First, I think the design looks cleaner without borders on the initial hover over the code block.
Second, using an outline on hover/focus for the icon is a good accessibility practice. Since outlines aren’t part of the box model, they don’t affect the component's width or height — everything stays aligned to the layout.
Finally, we want to visually communicate the interaction steps to the user. The outline does this effectively, so borders aren’t necessary in this case.

> Initially, during local development, I used both borders and outlines. But when pushing to remote, I realized the borders were redundant and removed 3 lines of code and set border: none. ☺️

js/copycode.js
```diff
@@ -0,0 +1,57 @@
...
+      button.setAttribute("aria-label","failed!"); 
+      button.classList.add("failed");
+
+      // remove previous timer on multiple clicks (data-timer-id)
+      if (button.dataset.timerId) clearTimeout(button.dataset.timerId);
+      
+      const timer = setTimeout(() => {
+        button.classList.remove("failed");
+        button.setAttribute("aria-label","click to copy code");
+      }, 100000);
```
ShubhamOulkar:
```suggestion
      }, 1000);

```

bjohansebas (approved):

css/style.css
```diff
@@ -346,12 +346,134 @@ pre {
...
+  mask-image: url("../images/copy-btn.svg");
+  mask-size: 1.5rem;
+  mask-repeat: no-repeat;
+  width: 1.5rem;
+  height: 1.5rem;
+  }
+
+  &:is(:hover, :focus) {
+    background-color: var(--hover-bg);
+    outline: 2px solid var(--hover-border);
```
bjohansebas, line 397: I don't think we should use outline here.

bjohansebas (changes requested):
Just a quick comment, thanks for working on this, and sorry for the delay.

css/style.css
ShubhamOulkar: 🤔 I did this for keyboard :focus state. It shows outline on :focus-within pre. 

Can you please test it? My system is broken.

ShubhamOulkar: It is perfectly fine. Can you tell me, why you don't want outline? It shows interaction with actual btn.

bjohansebas: If I click, I don't think we should show an outline. also i'm not really sure we should show that outline when it's selected via keyboard either — it feels like odd behavior. It makes sense for a button, but for a code example in general, I don't think it does.

imagen

Sorry for the delay

ShubhamOulkar: Right, outline should be only on btn on btn interaction not on the code block.

ShubhamOulkar: @bjohansebas, outline style on code block applied on line 351. I did this for accessibility improvements on :focus.

css/style.css
```diff
@@ -346,12 +346,134 @@ pre {
   border-radius: 3px;
   border: 1px solid #ddd;
   background-color: var(--code-bg);
+/* keyboard focus offset improve visibility */
+/* focus on desktop */
+  @media all and (min-width: 1010px) { 
+    &:focus {
+      outline-offset: 2px;
```
ShubhamOulkar: How about removing outline and putting border with 2px on :focus?

css/style.css
```diff
@@ -344,14 +344,133 @@ code {
 pre {
   padding: 16px;
   border-radius: 3px;
-  border: 1px solid #ddd;
+  border: 1px solid var(--border);
   background-color: var(--code-bg);
+/* keyboard focus offset improve visibility */
+    &:focus {
+      border-width: 2px;
```
bjohansebas: Just changing the color should be enough, what do you think? The border still keeps growing when I click on the code block
```suggestion
```

ShubhamOulkar: Right

bjohansebas (approved):
LGTM

bjohansebas:
I'll leave it for another day to see if anyone has something to say

js/copycode.js
```diff
@@ -0,0 +1,58 @@
...
+}
+
+function updateButtonState(button, statusClass, ariaLabel) {
+  button.setAttribute("aria-live", "polite");
+  button.setAttribute("aria-label", ariaLabel);
+  button.classList.add(statusClass);
+
+  // Clear any existing timer
+  if (button.dataset.timerId) clearTimeout(button.dataset.timerId);
+
```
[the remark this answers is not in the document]
ShubhamOulkar: Nice catch

bjohansebas:
thanks @ShubhamOulkar!

PR #1786, opened 2025-02-11
Refactor style.css

bjohansebas:
To make it more maintainable and to better edit our styles, I am separating the styles into multiple files depending on their purpose.

close: #1751

PR #1847, opened 2025-03-26
padding size updated

lakshay12540:

### DIFF
--- a/css/search.css
+++ b/css/search.css
@@ -2,7 +2,7 @@
   display: none;
   height: 2.5em;
   min-width: 100%;
-  padding: 5px;
+  padding: 10px;
 }
 
 .algolia-autocomplete {


bjohansebas (changes requested):
I don't see any changes in the UI (even though you changed the padding). Can you send screenshots of your change so we can better understand your modifications?

ShubhamOulkar:
> I don't see any changes in the UI (even though you changed the padding). Can you send screenshots of your change so we can better understand your modifications?

file `css/search.css` is redundant. should be removed in system design or in #1786. Styles are applied from style.css file. Also it is not linked in head.html

ShubhamOulkar:
@bjohansebas, I think we should close this PR. It is irrelevant as explained above.

PR #1859, opened 2025-04-01
fix(blog): added some clarification around dates in blog post

wesleytodd:
I didn't get to these after our meeting yesterday, but we agreed to soften the language around the dates here and then work to get consensus in the LTS ADR. We can circle back and update this as necessary, but this at least addresses the fact I incorrectly had concrete dates listed.

### DIFF
--- a/_posts/2025-03-31-v5-1-latest-release.md
+++ b/_posts/2025-03-31-v5-1-latest-release.md
@@ -114,7 +114,7 @@ only major version for most of the history of Node.js itself. Because of this, w
 provide a bit longer support. We want to do what is best for the ecosystem, so consider these goals not commitments.
 
 *: v4 is a special case, and we may extend MAINTENENCE support
-**: v5 MAINTENENCE and EOL dates are determined by when v6 is released, these dates reflect the earliest datest if we
+**: v5 MAINTENENCE and EOL dates are determined by when v6 is released, these dates reflect the earliest dates if we
 were to ship v6 on 2025-10-01
 ***: v6 work has not been officially started yet, this is simply the earliest date we can ship based on our proposed policy
 
--- a/_posts/2025-03-31-v5-1-latest-release.md
+++ b/_posts/2025-03-31-v5-1-latest-release.md
@@ -104,7 +104,7 @@ For the existing release lines, we will set the following phase dates:
 | ----- | ------- | ------ | ----------- | --- |
 | 4.x   |         |        | 2025-04-01 | *no sooner than 2026-10-01 |
 | 5.x   | 2024-09-11 | 2025-03-31 | **no sooner than 2026-04-01 | **no sooner than 2027-04-01 |
-| 6.x   | ***no sooner than 2025-10-01 | | | |
+| 6.x   | ***no sooner than 2026-01-01 | | | |
 
 </div>
 
--- a/_posts/2025-03-31-v5-1-latest-release.md
+++ b/_posts/2025-03-31-v5-1-latest-release.md
@@ -102,15 +102,21 @@ For the existing release lines, we will set the following phase dates:
 
 | Major | CURRENT | ACTIVE | MAINTENANCE | EOL |
 | ----- | ------- | ------ | ----------- | --- |
-| 4.x   |         |        | 2025-04-01 | 2026-10-01 or later |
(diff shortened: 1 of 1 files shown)


bjohansebas (approved):

_posts/2025-03-31-v5-1-latest-release.md
```diff
@@ -102,16 +102,22 @@ For the existing release lines, we will set the following phase dates:
...
 
 </div>
+
 As you can see, this means that v5.1.0 being tagged `latest` indicates that we moved from `CURRENT` to `ACTIVE` which
 starts the clock on EOL for v4 by moving it to `MAINTENANCE`. We recognize that v4 is a special case having been the
 only major version for most of the history of Node.js itself. Because of this, we want to remain flexible and also
 provide a bit longer support. We want to do what is best for the ecosystem, so consider these goals not commitments.
 
+*: v4 is a special case, and we may extend MAINTENENCE support
+**: v5 MAINTENENCE and EOL dates are determined by when v6 is released, these dates reflect the earliest datest if we
```
ctcpip:
```suggestion
**: v5 MAINTENENCE and EOL dates are determined by when v6 is released, these dates reflect the earliest dates if we
```

_posts/2025-03-31-v5-1-latest-release.md
```diff
@@ -102,16 +102,22 @@ For the existing release lines, we will set the following phase dates:
...
+were to ship v6 on 2025-10-01
+***: v6 work has not been officially started yet, this is simply the earliest date we can ship based on our proposed policy
```
ctcpip:
```suggestion
***: v6 work has not officially started yet, this is simply the earliest date we can ship based on our proposed policy
```

_posts/2025-03-31-v5-1-latest-release.md
```diff
@@ -102,16 +102,22 @@ For the existing release lines, we will set the following phase dates:
 
 | Major | CURRENT | ACTIVE | MAINTENANCE | EOL |
 | ----- | ------- | ------ | ----------- | --- |
-| 4.x   |         |        | 2025-04-01 | 2026-10-01 or later |
-| 5.x   | 2024-09-11 | 2025-03-31 | 2026-04-01 | 2027-04-01 |
-| 6.x   | TBD after 2025-10-01 | | | |
+| 4.x   |         |        | 2025-04-01 | *no sooner than 2026-10-01 |
+| 5.x   | 2024-09-11 | 2025-03-31 | **no sooner than 2026-04-01 | **no sooner than 2027-04-01 |
+| 6.x   | ***no sooner than 2025-10-01 | | | |
```
ctcpip:
```suggestion
| 6.x   | ***no sooner than 2026-01-01 | | | |
```

this would be more in alignment with the minimum dates, because it reflects a 3-month `current` period before the earliest possible `active` start for v6.  also seems like releasing `6.0.0` six months from today seems ambitious but maybe not?

I don't have strong feelings on this change; if we really think there's a good chance we could ship that early (2025-10-01), then it's fine as is, but would also mean v6 is forced to stay in `current` for 6 months

wesleytodd: The date math I did was based off this line from the ADR:

> During the `ACTIVE` period, a new major version may be released but *not* until at least 6 months have passed
  since the `CURRENT` version became `ACTIVE`.

v5 starts `ACTIVE` today, April 1st 2025, so I added 6 months to that. I believe that is correct according to the language in the ADR, but I do agree there is a good chance we *wont* ship earlier than 2026. So if we want to make this chance so folks don't come looking for something sooner I am alright with that, but it would not technically be the minimum defined by the proposed policy. But again, is is much more likely to be accurate if we push the date further out lol, no argument there.

ctcpip: I don't know how or where that text got written.  We agreed previously (at the 2025-09-25 meeting) to the following:

> A new major version will be released a minimum of 12 months after the last major version release.

wesleytodd: Ah, ok I must have pulled it from an outdated one. I will update the ADR and then we can just merge suggested change I think?

ctcpip (approved):

rxmarbles (approved):

Phillip9587 (approved):

PR #1867, opened 2025-04-11
feat: add feeds for the blog and vulnerabilities

### LINKED ISSUE #1763: We should generate a feed.xml
As the title says, we should generate a feed.xml file for the blog. We will probably need to split it into two files, one for vulnerabilities (@expressjs/security-wg) and another for the blog in general

rowanmanning:
This adds two new feeds to the site: `feed.xml` includes all blog posts, and `vulnerabilities.xml` includes all posts tagged with `vulnerabilities`.

You can validate the feeds via the [W3C Feed Validation Service](https://validator.w3.org/feed/). I also tried subscribing via Feedbin and all the posts appear as expected.

Resolves #1763

### DIFF
--- a/_includes/feed-entry.xml
+++ b/_includes/feed-entry.xml
@@ -11,7 +11,7 @@
   <published>{{ entry.date | date_to_xmlschema }}</published>
   <updated>{{ entry.last_modified_at | default: post.date | date_to_xmlschema }}</updated>
   {%- for tag in post.tags %}
-  <category term="{{ tag | replace: ',', '' | xml_escape }}" />
+  <category term="{{ tag | xml_escape }}" />
   {%- endfor %}
   <content type="html"><![CDATA[ <a href="{{ site.url }}{{ entry.url }}">Read {{ entry.title }}</a> ]]></content>
 </entry>
--- a/_includes/feed-entry.xml
+++ b/_includes/feed-entry.xml
@@ -1,0 +1,17 @@
+{%- assign entry = include.entry -%}
+<entry>
+  <id>{{ site.url }}{{ entry.url }}</id>
+  <link href="{{ site.url }}{{ entry.url }}" rel="alternate" type="text/html" />
+  <title type="text">{{ entry.title }}</title>
+  {%- if entry.author %}
+  <author>
+    <name>{{ entry.author }}</name>
+  </author>
+  {%- endif %}
+  <published>{{ entry.date | date_to_xmlschema }}</published>
+  <updated>{{ entry.last_modified_at | default: post.date | date_to_xmlschema }}</updated>
+  {%- for tag in post.tags %}
+  <category term="{{ tag | replace: ',', '' | xml_escape }}" />
+  {%- endfor %}
+  <content type="html"><![CDATA[ <a href="{{ site.url }}{{ entry.url }}">Read {{ entry.title }}</a> ]]></content>
+</entry>
--- a/_layouts/feed.xml
+++ b/_layouts/feed.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<feed xmlns="http://www.w3.org/2005/Atom" xml:lang="en">
+  <title type="text">{{ page.title }}</title>
+  <link href="{{ site.url }}{{ page.html_url }}" rel="alternate" type="text/html" />
+  <link href="{{ site.url }}{{ page.url }}" rel="self" type="application/atom+xml" />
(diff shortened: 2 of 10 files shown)


_includes/feed-entry.xml
```diff
@@ -0,0 +1,17 @@
+{%- assign entry = include.entry -%}
+<entry>
+  <id>{{ site.url }}{{ entry.url }}</id>
```
rowanmanning, line 3: In Atom, the `id` is [expected to be an IRI](https://www.ietf.org/rfc/rfc4287.html#section-4.2.6) however using the page URL seems to be relatively common practice. An alternative would be to mimic [GitHub's releases feed](https://github.com/expressjs/express/releases.atom) and use something like

```
tag:expressjs.com,<YEAR>:https://expressjs.com/<POST_URL>
```

however this doesn't get around the fact that the ID is based on the page URL.

_includes/feed-entry.xml
```diff
@@ -0,0 +1,17 @@
...
+  <title type="text">{{ entry.title }}</title>
+  {%- if entry.author %}
+  <author>
+    <name>{{ entry.author }}</name>
+  </author>
+  {%- endif %}
+  <published>{{ entry.date | date_to_xmlschema }}</published>
+  <updated>{{ entry.last_modified_at | default: post.date | date_to_xmlschema }}</updated>
+  {%- for tag in post.tags %}
+  <category term="{{ tag | replace: ',', '' | xml_escape }}" />
```
rowanmanning: This is overcomplicated to fix an issue with the current posts. [Jekyll post tags](https://jekyllrb.com/docs/posts/#tags-and-categories) are supposed to be space-separated. The tags in the current posts are comma-separated which results in trailing commas appearing in the tags.

I wasn't sure whether to add this little hack or to edit all of the posts to correct the tags. If the latter than I'll be setting the last modified date of each post; happy to do this if you want me to.

_layouts/feed.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<feed xmlns="http://www.w3.org/2005/Atom" xml:lang="en">
+  <title type="text">{{ page.title }}</title>
+  <link href="{{ site.url }}{{ page.html_url }}" rel="alternate" type="text/html" />
+  <link href="{{ site.url }}{{ page.url }}" rel="self" type="application/atom+xml" />
+  <id>{{ site.url }}{{ page.url }}</id>
+  <rights>Copyright © {{ site.time | date: '%Y' }}, The Express Contributors</rights>
```
rowanmanning: Is this copyright OK? I made assumptions based on the license. It's not a required element so I can ditch it or make changes.

bjohansebas: I don't think we need to include the date — we don't have it in the license of the repo.

Ping @expressjs/express-tc in case they have an opinion on this.

_includes/feed-entry.xml
bjohansebas: It's really better to fix it at the source than to make small hacks

_includes/feed-entry.xml
bjohansebas: I like either of those two options, and i think both are fine.

bjohansebas (commented):
Overall, everything looks good, @rowanmanning  thanks for working on this.

_includes/feed-entry.xml
rowanmanning: OK let's just go with the URL 👍

_includes/feed-entry.xml
rowanmanning: Great 👍 fixed in a separate commit

_layouts/feed.xml
rowanmanning: I removed the date. I'm unable to @ the team though, are you able to give them a heads up?

wesleytodd: Hm, I don't know much about how to represent this in feeds, but this is the correct guidance to follow if we need to have this at all: https://github.com/openjs-foundation/cross-project-council/blob/main/governance/IP_POLICY_GUIDANCE.md#2-copyright-notices

wesleytodd (approved):
I think this is great! Thanks!

_layouts/feed.xml
rowanmanning: Ah great, that's all the guidance I need, thanks. I've copied the format exactly from the linked doc 👍

bjohansebas (approved):
LGTM!

bjohansebas:
@rowanmanning Thanks!

PR #1875, opened 2025-04-15
Fix formatting of *s in v5-1-latest-release

gabriel-bezerra:
Problem: instead of showing asterisks on each line, the website has the text formatted bold and no line breaks.

Solution: add `  ` at the end of the line to break the lines.
Solution: escape the `*` as `\*` to avoid that they are interpreted as formatting commands.

Before
image

After
image


bjohansebas (approved):
Good Catch, thanks!

PR #1876, opened 2025-04-15
update api to 5x

bjohansebas:
This is necessary for the API to be imported correctly, since it’s not going to be translated yet.

Also, it updates it to display version 5 :)


bjohansebas:
cc: @expressjs/docs-collaborators

ShubhamOulkar (commented):
You need to change title and layout in front-matter.

ctcpip (approved):

PR #1877, opened 2025-04-17
Update routing.md

suryanarendrareddy:

### DIFF
--- a/en/guide/routing.md
+++ b/en/guide/routing.md
@@ -361,7 +361,7 @@ module.exports = router
 Then, load the router module in the app:
 
 ```js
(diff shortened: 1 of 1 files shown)


ShubhamOulkar (commented):
Thanks for the PR! The .js extension is optional when using require() in Common JS, and it's common practice to omit it in documentation examples for clarity and readability. We'll stick with the existing version to keep things consistent. This PR would become redundant.

bjohansebas:
@Narendra-017 Thanks for the PR!
Yep, agree with @ShubhamOulkar

PR #1878, opened 2025-04-18
Use 5.x path syntax in API examples

krzysdz:
@infuzz in https://github.com/expressjs/express/issues/6468#issuecomment-2816045469 pointed that the documentation for `app.all()` and `router.all()` uses old path syntax (from beta versions) that is no longer working.

After some searching I found 5 more examples that would not work in Express 5, because of old path syntax.

### DIFF
--- a/_includes/api/en/5x/req-route.md
+++ b/_includes/api/en/5x/req-route.md
@@ -3,7 +3,7 @@
 Contains the currently-matched route, a string. For example:
 
 ```js
-app.get('/user/{:id}', function userIdHandler (req, res) {
+app.get('/user/{:id}', (req, res) => {
   console.dir(req.route, { depth: null })
   res.send('GET')
 })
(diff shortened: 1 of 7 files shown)


_includes/api/en/5x/req-route.md
```diff
@@ -3,24 +3,29 @@
...
-       regexp: /^\/?$/i,
-       method: 'get' } ],
-  methods: { get: true }
+Route {
+  path: '/user/{:id}',
+  stack: [
+    Layer {
+      handle: [Function: userIdHandler],
+      keys: [],
+      name: 'userIdHandler',
```
krzysdz: Arrow function used as callback will result in `handle: [Function (anonymous)]`  and `name: 'anonymous'`.
Should I change this to make ESLint happy?

bjohansebas: Yep, please change it.

bjohansebas (approved):
LGTM

bjohansebas:
thanks @krzysdz!

PR #1884, opened 2025-04-24
Remove router.param(fn) from 5.x docs

krzysdz:
It has been deprecated since 4.11.0 and in 5.x `router.param()` requires name as the first argument.

The description in migration guide was copied from `app.param(fn)` that was mentioned there.


bjohansebas (approved):

bjohansebas:
thanks @krzysdz!

PR #1890, opened 2025-05-01
Use ESLint on JavaScript

bdkopen:
- Updates the `test` script to run on JS files.
- Whitelists `browser` and `jquery` as environments so their variables are seen as defined.
- Resolves all the ESLint errors.


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
-    "test": "eslint --ignore-path .gitignore --ignore-pattern _includes/readmes/ \"**/*.md\""
+    "test": "eslint . --ignore-path .gitignore --ignore-pattern _includes/readmes/ \"**/*.md\""
```
bdkopen: `.md` files are currently part of the `--ignore-pattern`. Should this exclusion be removed from the test script?

When the ignore-pattern is removed there are another 261 errors that could be fixed. Right now the test doesn't enforce the `.md` lint rules.

bdkopen: Without the `.`, ESLint doesn't actually run on any files.

ShubhamOulkar: Let's see if other team members have additional suggestions or perspectives on this approach

package.json
```diff
@@ -3,7 +3,7 @@
...
   },
   "devDependencies": {
     "eslint": "^8.54.0",
```
ShubhamOulkar, line 9:
```suggestion
    "eslint": "^9.26.0",
```
[8.x EOL](https://eslint.org/version-support/)

package.json
ShubhamOulkar:
```
"eslint-plugin-markdown": "^2.2.1"
```
If `.md` files are completely ignored, then the above linter is redundant. But `eslintConfig` explicitly apply linting rules on markdown js code blocks.

package.json
```diff
@@ -21,6 +21,10 @@
     "plugins": [
       "markdown"
```
ShubhamOulkar:
```suggestion
      "markdown",
      "promise"
```
Apply promise linting plugin. Also need to add promise linting rules in eslintConfig

package.json
ShubhamOulkar: Also update other plugins to the newest version

js/app.js
```diff
@@ -1,122 +1,115 @@
-$(function(){
-  var doc = $(document);
+$(function () {
+  const doc = $(document)
```
ShubhamOulkar: Keeping  `;` would be great

ShubhamOulkar (commented):
Hi @bdkopen, thank you for taking up this issue. Overall, the changes looks good! However, I would suggest moving the `eslintConfig` section from package.json into a separate file.

package.json
bdkopen: @ShubhamOulkar would it make more sense to update the ESLint version and other plugins in a separate PR? Migrating the ESLint config to the new format is sometimes quite involved.

package.json
bdkopen: The promise linting plugin is already enabled through the standard [config](https://github.com/standard/eslint-config-standard/blob/master/src/index.ts). Were there specific rules that you wanted enabled?

For now, I removed all these extra ESLint packages because they're already being provided through `eslint-config-standard`.

The `eslint-config-standard` README notes that you can install these three ESLint configs if you want to do a custom config. This repo never implemented a custom config, so there's no need to add them since `eslint-standard-config` includes them already.

js/app.js
bdkopen: @ShubhamOulkar since the repo extends the `eslint-config-standard` configuration, we receive the [no semicolon rule](https://standardjs.com/rules#semicolons). Would you like to break away from the recommended `eslint-config-standard` ruleset by adding an exception?

We could also switch to using ESLint's default recommended configuration over `eslint-config-standard`.

package.json
bdkopen: `eslint-config-standard` hasn't received updates in 2 years and doesn't support ESLint v9. It seems like maintenance has [stalled out](https://github.com/standard/standard/issues/1948) and some work is happening on the [`neostandard` fork](https://github.com/neostandard/neostandard).

We'll either have to switch to the fork or migrate to something new. I mentioned elsewhere on this PR already, but I like ESLint's recommended config over standard.

I would still suggest updating this separately. My goal with this PR was to get ESLint running correctly so the repository remains linted. If you know what configuration you want to use for ESLint v9, I may be able to help with that update too.

bdkopen:
> Hi @bdkopen, thank you for taking up this issue. Overall, the changes looks good! However, I would suggest moving the `eslintConfig` section from package.json into a separate file.

I moved the ESLint config into a new file. I responded to your other comments as well. Thanks for reviewing!

js/app.js
ShubhamOulkar: EsLint default configurations in good approach. 👍

ShubhamOulkar (commented):
Great, thank you for suggesting those options. Switching to ESLint's default configuration instead of `eslint-config-standard` makes sense, and updating ESLint to 9.x in a separate PR is a good approach.
 Let's see if other team members have additional suggestions on this approach

package.json
bjohansebas: MD files should not be excluded, and they weren't excluded before. I think when you added the dot, it now excludes the MD files from the linter.

package.json
bjohansebas: yes, this is something I'd like to discuss as well. I'm happy to use Neostandard, though I'm also open to using tools like Biome, which include both linting and formatting.

package.json
bdkopen: @bjohansebas ESLint v8 requires the `.` to be specified. ESLint v9 with a flat config adds it as a default ([docs](https://eslint.org/docs/latest/use/command-line-interface)).

You can verify this in `main` because `npm test` returns no ESLint errors until you add the `.`. The MD files are excluded because an ignore pattern is specified.

bjohansebas: The main only analyzes MD files, it’s only analyzing the md files. Add a semicolon to one of the code examples in those files, and it will throw errors

bdkopen: Looking at this again, you're right that `\"**/*.md\"` is listed here. I misread the script and thought it was part of the ignore pattern.

Would you like me to update this PR so both the code and .md files are linted or just close the PR?

bjohansebas: Please, just update this PR

bdkopen: @bjohansebas I updated the PR to lint both JS files and the un-ignored markdown files.

bdkopen:
I fixed the `test` command so both Markdown and JS files are linted. Please let me know if there's anything else you'd like done in this PR.

ctcpip (changes requested):
at the very least, we still should have semicolons in code

bdkopen:
> at the very least, we still should have semicolons in code

Sure, I added a rule that overrides the `standard` default [no semicolon rule](https://standardjs.com/rules#semicolons).

ctcpip:
oh...   I didn't realize there would be so many missing semicolons.  to unblock this, what we would ask is to change the configured rules such that the minimal amount of code change via `--fix` would apply.  we don't want to blow up the git history right away.  (maybe in the future though)

in other words, let's focus this PR so that linting is executing, but let's not change rules/files as part of it.  or if that is not possible, which is understandable, this will need to wait till we resolve our future direction on linting

bdkopen:
@ctcpip I dropped the commit enforcing semicolons. Without modifying the lint rules, the linting will remove semicolons in a few files since that is what the `standard` config defines.

This PR can be focused on requiring linting execution as long you're comfortable with enforcing the existing rules. If we don't want to enforce the existing rules, it seems like this PR will be blocked until the direction of linting is determined.

bjohansebas:
I think you could make it so that only JavaScript files have semicolons, and markdown files don’t have semicolons.

bdkopen:
> I think you could make it so that only JavaScript files have semicolons, and markdown files don’t have semicolons.

@bjohansebas that seems like a good intermediate solution. Updated!

ShubhamOulkar (approved):

bjohansebas (approved):

ctcpip (changes requested):
this PR should not modify the linting rules

bdkopen:
It seems like there's not a clear direction with linting in the express project right now (https://github.com/expressjs/discussions/issues/327). I'll close this PR for now. If a linting standard is chosen, I'd be happy to help integrate it within the repos.

PR #1895, opened 2025-05-08
Fix z-index bug in style.css

raaid3:
Fix z-index bug where "copy code" button and header have the same index leading to "copy code" button appearing on top of the header.


ShubhamOulkar (commented):
LGTM

bjohansebas (approved):

bjohansebas:
thanks @raaid3

PR #1901, opened 2025-05-12
Add Azerbaijani language support

Ali-GreenHeart:


bjohansebas:
Thank you for your contribution. However, we can't accept it at the moment because we're now starting to use Crowdin. We're beginning with a small list of languages to maintain translation quality. We may include more languages later, but that will depend on how many people are interested in maintaining each language.

PR #1908, opened 2025-05-19
Update req.md

0xadri:
Fix link to Node’s own `request` object docs


ShubhamOulkar (commented):
Hi @0xadri, the current link to `http.IncomingMessage` is correct and should remain unchanged. Visit express [req](https://github.com/expressjs/express/blob/master/lib/request.js#L30) object for more clarification.

bjohansebas:
>  the current link to http.IncomingMessage is correct and should remain unchanged

Yes, the link is correct.

Issue #1787, opened 2025-02-12
Redesign of the Express Documentation


UlisesGascon:
Over the time we had commented that we want to "refresh" the website, in the past Node.js requested support from the foundation to hire a designer to refresh the website completely. Is this something that we want to explore? Do we have the resources to support this process? I assume that we need to do a great work of content curation, etc... prior to start the re-design process.

Anyway, I am creating this issue to have a common point to discuss it and track the evolution (cc: @expressjs/docs-wg @expressjs/express-tc @bjohansebas @micheleriva @AugustinMauroy)

AugustinMauroy:
IMO before asking for a developer we need to know which website structure do we want.

micheleriva:
@UlisesGascon, @AugustinMauroy, one of my two co-founders (@nglngl) is a designer. We can offer our help for redesigning and most importantly for facilitating any discussion around the redesign, if you find it useful

bjohansebas:
Honestly, help like this would be very useful for the website. Having something well-defined to improve the design would be great. For example, we have several issues like #1780, #1702, #1758, and #1760 that propose changes to the current style.  

Recently, we also changed the dark mode colors, and there are many things I would still like to change. However, the problem is that I am not a designer, and I don't know if the other members are either. I’m not sure how things should be structured to provide the best experience and give the website a fresh look.  

Sometimes, we have had to revert certain changes, like #1668, which was reverted in #1732. Initially, it seemed like a good idea, but due to accessibility concerns, it wasn’t. Situations like this might keep happening because there isn’t a clear design guideline.  

It’s also true that we’ve taken a lot of inspiration from other documentation sites, such as Node.js, Vite.js, and Electron, among others, to try and achieve a fresh style.

nglngl:
Hi there! I really appreciate @AugustinMauroy's comment:

> IMO before asking for a developer we need to know which website structure we want.

This is something where I'd like to help too. I can structure a small workshop where in 45min we can try to understand what kind of structure, style, and features you want on the website. That's part of my daily duties at Orama and I've been doing this for the past 15 years, and I'd love to contribute.

This would also give me a good direction for proposing some designs if you agree. We can sponsor the activity as Orama (we have a strong design team and of course, engineering too) so we can help both on code, design, and overall website structure.

I'm on the OpenJS Slack channel too (my full name is Angela Angelini there) and I'm happy to continue the conversation either on GitHub or Slack :)

bjohansebas:
[...]

PR #1915, opened 2025-05-27
Display author avatars in the blog

### LINKED ISSUE #1758: Show the author's image in the blog
As the title says, it would be great to display the author's GitHub profile picture.

rowanmanning:
This resolves #1758 by adding author images to blog posts, both on the post page and in the list view. It's not overly designed - I went for a simple solution for now that can be expanded on later if the site is refreshed (as suggested in #1787).

This required some foundational work to ensure that:

  * We support multiple authors on a post with different metadata. I also updated all existing posts to ensure that the metadata is in place
  * Moving the author list to an include so that we don't have to repeat a load of Liquid logic in two places on the site

I outlined this work in more detail in [this comment on the original issue](https://github.com/expressjs/expressjs.com/issues/1758#issuecomment-2820523545).

## Screenshots

Overview page with single author:

Screenshot 2025-05-27 at 15 49 14

Overview page with one author with an avatar and one without:

Screenshot 2025-05-27 at 15 49 14

Overview page with two authors:

Screenshot 2025-05-27 at 15 49 23

Light mode:

Screenshot 2025-05-27 at 15 49 33

Blog post page:

Screenshot 2025-05-27 at 15 49 47

### DIFF
--- a/_includes/blog/authors.html
+++ b/_includes/blog/authors.html
@@ -3,7 +3,17 @@
   <div class="blog-authors">
     By
     {% for author in authors %}
-      {{author.name}}{% unless forloop.last %}, {% endunless %}
+
+      {%- assign label = author.github | prepend: '@' -%}
+      {%- if author.name -%}
+        {%- assign label = author.name -%}
+      {%- endif -%}
+
+      {% if label and author.github %}
+        <a href="https://github.com/{{author.github}}" rel="author">{{label}}</a>{% unless forloop.last %}, {% endunless %}
+      {% elsif label %}
+        {{label}}{% unless forloop.last %}, {% endunless %}
+      {% endif %}
     {% endfor %}
   </div>
 {% endif %}
--- a/_includes/blog/authors.html
+++ b/_includes/blog/authors.html
@@ -1,0 +1,9 @@
+{%- assign authors = include.authors -%}
+{% if authors %}
+  <div class="blog-authors">
+    By
+    {% for author in authors %}
+      {{author.name}}{% unless forloop.last %}, {% endunless %}
+    {% endfor %}
+  </div>
+{% endif %}
--- a/_includes/blog/authors.html
+++ b/_includes/blog/authors.html
@@ -10,7 +10,9 @@
       {%- endif -%}
 
       {% if label and author.github %}
-        <a href="https://github.com/{{author.github}}" rel="author">{{label}}</a>{% unless forloop.last %}, {% endunless %}
+        <a href="https://github.com/{{author.github}}" rel="author" class="blog-author-link">
+          <img src="https://github.com/{{author.github}}.png?size=64" alt="" class="blog-author-avatar" />
+          <span class="blog-author-link-label">{{label}}</span></a>{% unless forloop.last %}, {% endunless %}
       {% elsif label %}
         {{label}}{% unless forloop.last %}, {% endunless %}
       {% endif %}
--- a/_includes/feed-entry.xml
+++ b/_includes/feed-entry.xml
@@ -5,7 +5,8 @@
   <title type="text">{{ entry.title }}</title>
   {%- for author in entry.authors %}
   <author>
-    <name>{{ author.name }}</name>
+    {% if author.name %}<name>{{ author.name }}</name>{% elsif author.github %}<name>@{{ author.github }}</name>{% endif %}
+    {% if author.github %}<uri>https://github.com/{{ author.github }}</uri>{% endif %}
   </author>
   {%- endfor %}
   <published>{{ entry.date | date_to_xmlschema }}</published>
--- a/_includes/feed-entry.xml
+++ b/_includes/feed-entry.xml
@@ -4,10 +4,16 @@
   <link href="{{ site.url }}{{ entry.url }}" rel="alternate" type="text/html" />
   <title type="text">{{ entry.title }}</title>
   {%- for author in entry.authors %}
-  <author>
-    {% if author.name %}<name>{{ author.name }}</name>{% elsif author.github %}<name>@{{ author.github }}</name>{% endif %}
-    {% if author.github %}<uri>https://github.com/{{ author.github }}</uri>{% endif %}
-  </author>
+    <author>
+        {% if author.name %}
+          <name>{{ author.name }}</name>
+        {% elsif author.github %}
+          <name>@{{ author.github }}</name>
+        {% endif %}
+        {% if author.github %}
+          <uri>https://github.com/{{ author.github }}</uri>
+        {% endif %}
+    </author>
   {%- endfor %}
   <published>{{ entry.date | date_to_xmlschema }}</published>
   <updated>{{ entry.last_modified_at | default: post.date | date_to_xmlschema }}</updated>
--- a/_includes/feed-entry.xml
+++ b/_includes/feed-entry.xml
@@ -3,11 +3,11 @@
   <id>{{ site.url }}{{ entry.url }}</id>
   <link href="{{ site.url }}{{ entry.url }}" rel="alternate" type="text/html" />
   <title type="text">{{ entry.title }}</title>
-  {%- if entry.author %}
+  {%- for author in entry.authors %}
   <author>
-    <name>{{ entry.author }}</name>
+    <name>{{ author.name }}</name>
   </author>
-  {%- endif %}
+  {%- endfor %}
   <published>{{ entry.date | date_to_xmlschema }}</published>
   <updated>{{ entry.last_modified_at | default: post.date | date_to_xmlschema }}</updated>
   {%- for tag in post.tags %}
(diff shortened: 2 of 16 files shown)


css/style.css
```diff
@@ -1487,9 +1487,22 @@ h2 a {
...
+  text-decoration: none;
+}
+.blog-author-link-label {
+  text-decoration: underline;
+}
+.blog-author-avatar {
+  border-radius: 50%;
+  display: inline;
+  vertical-align: middle;
+  width: 1.5em;
```
wesleytodd, line 1505: Could we just push the avatar over a tiny bit to match the spacing on both sides? I have no idea if this is the right value (aka it is probably too much) but figured I would leave it up to y'all to get the spacing right.

```suggestion
  width: 1.5em;
  padding: .25em;
```

wesleytodd (approved):

_includes/feed-entry.xml
```diff
@@ -3,11 +3,12 @@
...
   <title type="text">{{ entry.title }}</title>
-  {%- if entry.author %}
+  {%- for author in entry.authors %}
   <author>
-    <name>{{ entry.author }}</name>
+    {% if author.name %}<name>{{ author.name }}</name>{% elsif author.github %}<name>@{{ author.github }}</name>{% endif %}
+    {% if author.github %}<uri>https://github.com/{{ author.github }}</uri>{% endif %}
   </author>
-  {%- endif %}
+  {%- endfor %}
```
bjohansebas: Could you separate the code better? I know we don't have a linter, but it's hard to read the code.

_posts/2024-10-15-v5-release.md
```diff
@@ -1,7 +1,10 @@
 ---
 title: "Introducing Express v5: A New Era for the Node.js Framework"
 tags: releases
-author: Wes Todd and the Express Technical Committee
+authors:
+  - name: Wes Todd
+    github: wesleytodd
+  - name: Express Technical Committee
```
bjohansebas, line 7: We could use the GitHub username for the TC, so it doesn't end up without a photo.

css/style.css
rowanmanning: Done 👍 I added a half character of left margin which events things out better. It now looks like this:

Screenshot 2025-05-28 at 09 32 08

_includes/feed-entry.xml
rowanmanning: Done 👍

_posts/2024-10-15-v5-release.md
rowanmanning: I couldn't find a GitHub user for the technical committee specifically, I added [expressjs](https://github.com/expressjs) and it looks like this:

Screenshot 2025-05-28 at 09 42 06

Is this what you meant? If not I can revert. One doubt I had is that we're kind of spamming the blog page with lots of the same image and link out to the GitHub org. Maybe this is fine?

css/style.css
```diff
@@ -1487,9 +1487,23 @@ h2 a {
...
```
ShubhamOulkar, line 1505:
```suggestion
  width: 2em;
  height: 2em;
```
32px avatar looks good

css/style.css
```diff
@@ -1487,9 +1487,23 @@ h2 a {
...
+  color: inherit;
+  margin-left: 0.5ch;
+  text-decoration: none;
+}
+.blog-author-link-label {
+  text-decoration: underline;
+}
+.blog-author-avatar {
+  border-radius: 50%;
+  display: inline;
```
```suggestion
```
Images are `inline-block` by default and `inline` prevents height or aspect-ratio from working properly. Also set equal width and height for a perfect circle. GitHub avatars are perfect squares, so the current border-radius: 50% works fine now, but this ensures it stays consistency in future design updated.

ShubhamOulkar (approved):

css/style.css
rowanmanning: I went with the smaller size because it matched up more with the existing line height so it didn't shift any elements vertically. With `2em` the image no longer aligns with the text, pushing the height of the post metadata to be larger than a regular paragraph. I think the post view looks a little janky with larger images. Happy to go with it if others disagree but would like more opinions.

Screenshot 2025-05-28 at 14 20 11

css/style.css
rowanmanning: Done 👍

_includes/blog/authors.html
```diff
@@ -0,0 +1,21 @@
...
+    {% for author in authors %}
+
+      {%- assign label = author.github | prepend: '@' -%}
+      {%- if author.name -%}
+        {%- assign label = author.name -%}
+      {%- endif -%}
+
+      {% if label and author.github %}
+        <a href="https://github.com/{{author.github}}" rel="author" class="blog-author-link">
+          <img src="https://github.com/{{author.github}}.png?size=64" alt="" class="blog-author-avatar" />
```
ShubhamOulkar, line 14:
```suggestion
          <img src="https://github.com/{{author.github}}.png?size=64" alt="{{label}} avatar" class="blog-author-avatar" />
```

_includes/blog/authors.html
```diff
@@ -0,0 +1,21 @@
...
+    By
+    {% for author in authors %}
+
+      {%- assign label = author.github | prepend: '@' -%}
+      {%- if author.name -%}
+        {%- assign label = author.name -%}
+      {%- endif -%}
+
+      {% if label and author.github %}
+        <a href="https://github.com/{{author.github}}" rel="author" class="blog-author-link">
```
ShubhamOulkar, line 13:
```suggestion
        <a href="https://github.com/{{author.github}}" rel="author" class="blog-author-link" target="_blank">
```

ShubhamOulkar (changes requested):
Accessibility improvements

css/style.css
ShubhamOulkar: > image no longer aligns with the text

most of the websites use not aligned image with the text. text always align at center with respect to the image. [example](https://developer.mozilla.org/en-US/blog/color-models-humans-devices/).

_includes/blog/authors.html
rowanmanning: I avoided target `_blank` because my understanding is that it's bad accessibility practice – it breaks the [change on request](https://www.w3.org/WAI/WCAG21/Understanding/change-on-request.html) success criterion. From the linked guide:

> This success criterion aims to eliminate potential confusion that may be caused by unexpected changes of context **such as automatic launching of new windows**...

Anecdotally I work with an engineer with peripheral vision loss who has moaned to me about this practice in the past 😄

The [related guidance](https://www.w3.org/TR/WCAG20-TECHS/G201.html) says that, if you're opening in a new window with `_blank` then you should either provide a visual indicator (this will clutter the interface in my opinion) or include text like "opens in a new window". For me the lowest-code solution is just to let users decide where the link gets opened.

ShubhamOulkar: Target `_blank` has specific use cases like visiting author profile. You can verify it from any blog/documentation website.

_includes/blog/authors.html
rowanmanning: I struggled to find some good "official" guidance on this, across other websites:

  * MDN includes a generic "Author avatar"
  * GitHub includes a generic "Owner avatar" in some places and the username in others
  * Mastodon includes an empty `alt` attribute

So things aren't consistent. Then I tried using VoiceOver on my Mac to see what the actual result is. With GitHub in a place like this:

Screenshot 2025-05-28 at 15 41 46

The screen reader announces `link image @rowanmanning` then I tab to the name and it announces `link @rowanmanning`. The extra information is offering no real value to a screen reader user. I tried your suggestion with my local copy of the site and you get a similar experience:

> "link image Ulises Gascón avatar [tab] link Ulises Gascón"

Again, I think the information is redundant. In our case the avatar is purely visual because it has the user's name/link right next to it. We can't easily include an `alt` text that actually describes the content of the image ([as this blog post suggests](https://nicolas-steenhout.com/alt-text-for-avatars/)) so the (admittedly lazy) but next best thing is to get screen readers to ignore it.

ShubhamOulkar: Thank you for correcting me. You are right, I didn't consider image is children of anchor links.

_includes/blog/authors.html
rowanmanning: Yep I agree it's _fairly_ common, but it's not good accessibility practice as your original comment suggested. I'd be more on the side of making this accessible than copying the use of `_blank` from other sites - this doesn't fit into any of the ["good reasons" here](https://css-tricks.com/use-target_blank/).

Here are a few blogs that don't open the author link in a new tab to illustrate it's not _everyone_ doing this (grabbed from recent reading): [The ESLint blog](https://eslint.org/blog/2025/05/eslint-v9.27.0-released/), [The Docker blog](https://www.docker.com/blog/introducing-docker-hardened-images/) - but it does provide its own author pages.

ShubhamOulkar: Ok agreed

ShubhamOulkar (approved):
LGTM 
Please resolve merge conflict

bjohansebas (approved):
LGTM

bjohansebas:
thanks @rowanmanning!

PR #1925, opened 2025-05-28
docs: remove contribution guidelines for expressjs.com in contributing page

bjohansebas:
It's a nice attempt, but there are other ways for people to help with the documentation. In the end, this is a document that should be split into multiple pages, since it's quite long and very few people actually read it because of its length (especially if we add something specific from the documentation repo)

I'll leave the script logic here in case we need it in the future.


chrisdel101:
hey @bjohansebas, i haven’t been around for awhile just saw this and was a bit confused. how do you know nobody reads it? or that it’s too long? 
how can documentation be too long?

It was a lot of work to create this content. I’d hope you have good reasons for just removing it completely :)

bjohansebas:
Hi @chrisdel101,

I'm just removing it from that page because it's quite long. I know you did a great job improving that section, and it’s definitely been helpful (thank you!), but that part will only be read by someone specifically interested in contributing to the website (and if they want to contribute, they’ll find the info here: https://github.com/expressjs/expressjs.com/blob/gh-pages/CONTRIBUTING.md). Also, this page should focus more on contributing to Express and its overall ecosystem rather than the website itself (although, honestly, this page should probably be split into multiple pages anyway) plus, it adds more content to translate via Crowdin.

I'm just trying to make the website content as useful and simple as possible. WDYT?

ShubhamOulkar:
> this page should probably be split into multiple pages anyway) plus, it adds more content to translate via Crowdin.

I am agreed. 

> It was a lot of work to create this content. I’d hope you have good reasons for just removing it completely :)

I don't think we need to remove all content, we need to update content with newer changes and creating new pages.

chrisdel101:
Ahh okay. It sounds like you have some specific ideas and I might just be misunderstanding while thinking the same.

Having it moved to a different URL for the website alone would be great - or not? Is what you were thinking?

Having it in repo alone is fine but we have a website so we should use it, no?
Optimally (in my mind) the website would have it's own contributing page. I mean, I contribute to the website and not express itself (much) and so I'm an example of someone that this would appeal to. 

This approach (website contributing page) was shut down before, hence the need to put in the same file as the express contrib and use that complicated script to key things dry.

If you want to actually cut the CONTRIBUTING.md itself from the repo (which is what I thought you meant) you could revert to the version before that PR. It was much more concise.

Make sure to remove the link at the top of the contributing.html page as it is now dead.

bjohansebas:
> If you want to actually cut the CONTRIBUTING.md itself from the repo (which is what I thought you meant) you could revert to the version before that PR. It was much more concise

The file located [here](https://github.com/expressjs/expressjs.com/blob/gh-pages/CONTRIBUTING.md) i don’t intend to delete it and it shouldn’t be deleted, but I think it’s unnecessary to have it on the website [page](https://expressjs.com/en/resources/contributing.html).

The content of this file is actually fine, it has what’s needed and helps new contributors understand how the website is developed, the previous version was missing some things.

> Optimally (in my mind) the website would have it's own contributing page. I mean, I contribute to the website and not express itself (much) and so I'm an example of someone that this would appeal to.

I really don’t think we should have a dedicated section on the website for contributing to the page. It makes sense to have it in the [repository](https://github.com/expressjs/expressjs.com/blob/gh-pages/CONTRIBUTING.md), but not on the [website](https://expressjs.com/en/resources/contributing.html). If someone wants to contribute, they will go to the repository.

Anyway, if you don’t agree, you can block the PR and reach a consensus.

bjohansebas:
By the way, i'm working on https://github.com/expressjs/.github/pull/24 so that this guide is more general, since the content from there will be brought here.

ShubhamOulkar (commented):
I think we should document the following two links on website page.
https://github.com/expressjs/.github/blob/master/CONTRIBUTING.md
https://github.com/expressjs/expressjs.com/blob/gh-pages/CONTRIBUTING.md (for translation and expressjs.com)

crandmck:
Removing the text is fine, but I agree with @ShubhamOulkar that we should replace it with a link tohttps://github.com/expressjs/expressjs.com/blob/gh-pages/CONTRIBUTING.md.

crandmck (changes requested):
Link to CONTRIBUTING.md.

PR #1926, opened 2025-05-28
codeql: remove ruby from CodeQL language matrix

bjohansebas:
We don't need to run CodeQL for Ruby, it's just wasting resources unnecessarily.


ShubhamOulkar (approved):
I am approving this since the project does not contain any Ruby code

PR #1928, opened 2025-05-28
i18n: upload readmes to crowdin

bjohansebas:
This would allow the translation of the main packages' READMEs.

ref: https://github.com/expressjs/multer/issues/1287

@expressjs/express-tc  for approval, since I'm not sure if this is what we want.


bjohansebas:
Well, it looks like more changes still need to be made, but what do you think of the idea?

PR #1935, opened 2025-05-31
Eliminate layout shift in menu toggle arrow with CSS-only solution

ShubhamOulkar:
### Problem 
Previously, the menu toggle arrow in the Table of Contents was implemented by updating the button’s inner HTML with different Unicode arrow entities (e.g., &#x25BA; for closed and &#x25BC; for open) via JavaScript. This approach caused a layout shift when toggling the menu, because the two Unicode arrow characters have different widths and font rendering. As a result, the button size and surrounding layout would visibly jump when the menu was opened or closed, negatively impacting user experience and visual stability.

### Solution
This PR refactors the arrow indicator to use a CSS-only solution. The arrow is now rendered using a CSS ::after pseudo-element on the toggle button, with its direction controlled by toggling a class and applying a transform: rotate() property. This ensures the arrow always occupies the same space, eliminating layout shifts and providing a smoother, more consistent UI. The button’s HTML remains static, and all visual changes are handled via CSS transitions.

### Mobile screens specific issue
|before|after|
|-------|------|
|Edited_20250531_175842|Edited_20250531_183603|

### DIFF
--- a/_includes/api/en/3x/menu.md
+++ b/_includes/api/en/3x/menu.md
@@ -1,4 +1,4 @@
-<button id="menu-toggle" title="show table of content">On this page <span>&#x25BA;</span></button>
+<button id="menu-toggle" title="show table of content">On this page</button>
 <ul id="menu">
     <li><a href="#express">express()</a></li>
     <li id="app-api"> <a href="#application">Application</a>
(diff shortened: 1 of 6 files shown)


bjohansebas (approved):
LGTM

ShubhamOulkar (commented):
lgtm

PR #1945, opened 2025-06-08
Fix ID on Header on Performance Best Practices Page

bdkopen:
Fixes a minor formatting issue where `{#in-environment}` was appearing in plain text instead of being the ID on the `h2` element.


ShubhamOulkar (approved):
thank you @bdkopen

PR #1947, opened 2025-06-10
doc: Fix ID on Header on Performance Best Practices Page (#1945)

nakumdushyant:
fix: correctly link id to header


bjohansebas (changes requested):
What change caused this? The ID is being applied correctly with the way it's done in the main branch.

PR #1948, opened 2025-06-12
fix: content overflow on small screen 

### LINKED ISSUE #1831: Fix Content Overflow on API Pages for Mobile Devices
Content breaks layout around 470px and header around 360px. This causing horizontal scroll on mobile devices.

Image
 
[SC reflow 1.4.10](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html)

ShubhamOulkar:
> Note: Grid layout is introduced in our design

closes #1831

preview https://deploy-preview-1948--expressjscom-preview.netlify.app/en/5x/api

image


ShubhamOulkar (commented):
LGTM. I suggest fast-tracking this PR, as the overflow issue is affecting a all of users.

_includes/api/en/5x/app-METHOD.md
```diff
@@ -10,39 +10,37 @@ PUT, POST, and so on, in lowercase. Thus, the actual methods are `app.get()`,
...
-* `subscribe`
-* `trace`
-* `unlock`
-* `unsubscribe`
-</td>
-</tr>
-</table>
+<div class="methods-columns">
+  <ul>
+    <li><code>checkout</code></li>
```
bjohansebas, line 15: I think we should keep the Markdown as it is. Jekyll performs several processes behind the scenes, especially for code highlighting, so I don't think we should move this to HTML.

ShubhamOulkar: > I don't think we should move this to HTML.

Tables aren't responsive by default, and the original layout caused horizontal overflow on small screens. I used semantic HTML `<ul>` with Flexbox to solve this. It’s now responsive and easier to maintain for future design updates.

The only downside is having HTML in Markdown, which may slightly affect doc editing for contributors — but I think the improved UX and maintainability are worth it.

> I think we should keep the Markdown as it is

As an alternative, we could also wrap Markdown lists in a styled `<div class="">` to avoid full HTML, but that still requires `some custom CSS.`  

```
<div class="">
  - mkcol
  - move
  - m-search
  - notify
  - options
  - patch
  - post
</div>
```
`We need to remove this unnecessary table. Let me know if you have any concerns with this approach.`

bjohansebas (approved):
LGTM, Sorry for the delay.

ShubhamOulkar:
> Sorry for the delay.

 Do not mention, Understandable 😊

PR #1964, opened 2025-07-06
chore: ♻️ remove jquery

### LINKED ISSUE #1590: refactoring from jQuery to modern JavaScript

ShubhamOulkar:
### Vanilla JS written for following components

- [X] scroll to top of the page
- [X] Display current selected language in language picker component
- [X] Menu highlighting on scroll
- [X] i18n notice 

closes #1590

### ~31 kB reduction in JS payload
Screenshot 2025-07-21 193649


css/style.css
```diff
@@ -481,14 +481,14 @@ pre:has(code) button.failed {
   }
 }
 
-/* top button */
+/* scroll to top button */
 
-.scroll #top {
+body.scroll #top {
```
bjohansebas: why?

ShubhamOulkar: It's not finalised. I added to make sure that class only applied to body, since I am changing pre written JS and without typos 😁. I will remove it.

js/app.js
```diff
@@ -1,122 +1,127 @@
...
-    while (c.charAt(0) === ' ') c = c.substring(1, c.length);
-    if (c.indexOf(nameEQ) === 0) return decodeURIComponent(c.substring(nameEQ.length, c.length));
+  function readCookie(name) {
+    const nameEQ = encodeURIComponent(name) + "=";
+    const ca = document.cookie.split(';');
+    for (var i = 0; i < ca.length; i++) {
+      var c = ca[i];
+      while (c.charAt(0) === ' ') c = c.substring(1, c.length);
+      if (c.indexOf(nameEQ) === 0) return decodeURIComponent(c.substring(nameEQ.length, c.length));
+    }
```
bjohansebas: We can now use new properties, there's no need to use var. Let's use const/let where possible.

_includes/head.html
```diff
@@ -65,7 +65,6 @@
     {% else %}
     <meta property="twitter:image" content="https://expressjs.com/images/og.png">
     {% endif %}
-    <script data-cfasync="false" src="//ajax.googleapis.com/ajax/libs/jquery/2.1.1/jquery.min.js"></script>
     <script data-cfasync="false" src="/js/app.js"></script>
```
bjohansebas:
```suggestion
    <script data-cfasync="false" defer src="/js/app.js"></script>
```

We can use defer here, and we won’t need to use `document.addEventListener("DOMContentLoaded", function () {` anymore (https://github.com/expressjs/expressjs.com/blob/c327d43be63208fbcabf8dc7aca15a2394de1b51/js/app.js#L1)

js/app.js
```diff
@@ -1,122 +1,127 @@
...
-    }  
+  // scroll to top of the page
+  if (scrollToTopBtn) {
+    scrollToTopBtn.addEventListener("click", function (e) {
+      e.preventDefault();
+      window.scrollTo({
+        top: 0,
+        behavior: "smooth"
+      })
+    });
```
bjohansebas: Why remove the link's default behavior and handle it with JavaScript? The less JavaScript we use, the better.

ShubhamOulkar: I rewrite the following animate jQuery. I just don't want to change any default behavior in this PR. Just focused on shifting JS code to browsers API.  
```
 // top link
  $('#top').click(function(e){
    $('html, body').animate({scrollTop : 0}, 500);
    return false;
  });
```
Yes, as you said, we can implement this only with CSS. But we should do it in separate PR.

_includes/head.html
ShubhamOulkar: Yes, we should use `defer`. But doesn't `data-cfasync="false"` do the same ? Optimize page load performance.

bjohansebas: This is worth discussing with more people, because up to now we've always disabled [Rocket Loader](https://developers.cloudflare.com/speed/optimization/content/rocket-loader/) using that property (data-cfasync="false"). In the meantime, it would be better to use defer. I've even considered proposing that we disable[ Rocket Loader](https://developers.cloudflare.com/speed/optimization/content/rocket-loader/) entirely, but I haven't done so yet since I haven't fully analyzed how it works.

js/app.js
bjohansebas: Let’s see which one lands first, #1965 or your PR 😊

_includes/head.html
ShubhamOulkar: Let's do it in separate PR. Since doesn't make sense to use defer on single file. I am not going to change all other files.

js/app.js
ShubhamOulkar: I hope so, this will land first.

bjohansebas:
I don’t understand, we’re already using defer on the other JavaScript files we import, and that’s how it’s being handled. We’ve manually disabled Rocket Loader because it previously caused issues (see https://github.com/expressjs/expressjs.com/issues/1508), so using defer here makes complete sense and should be part of this PR. 

whether or not to use Rocket Loader is a separate discussion, but I’d really opt for keeping it disabled.

_includes/head.html
bjohansebas: I don’t understand, we’re already using defer on the other JavaScript files we import, and that’s how it’s being handled. We’ve manually disabled Rocket Loader because it previously caused issues (see https://github.com/expressjs/expressjs.com/issues/1508), so using defer here makes complete sense and should be part of this PR.

whether or not to use Rocket Loader is a separate discussion, but I’d really opt for keeping it disabled.

ShubhamOulkar: Oh, I noticed only top theme.js file. menu and copy code are deferred. 😂 Funny is copy code written by me, I don't remember. 🤣🤣

ShubhamOulkar: Let's add 'defer' 👍

ShubhamOulkar (commented):
LGTM 👍

@bjohansebas, Is it possible to land this in production?

bjohansebas (approved):
LGTM, the code looks pretty good. I’ve been testing the site quite a bit and haven’t found any bugs. It’s been open for a month hoping more people could review it, but well, it’s time to merge it. Thanks! Incredible work.

Issue #1943, opened 2025-06-04
Improve design for api, docs and blogs pages


ShubhamOulkar:
### Demo: 
[design preview](https://expressjsnewdesign.netlify.app/)

This is a minor redesign improves the Express.js website’s UI and usability across both dark and light modes. The update includes navbar, layout, spacing, color, and typography tweaks to create a more modern and consistent look. 

### Task to do
- [X] https://github.com/expressjs/expressjs.com/issues/1831 || #1948 
- [X] Bring consistency in header and footer design || https://github.com/expressjs/expressjs.com/pull/1962
- [X] #1590 || https://github.com/expressjs/expressjs.com/pull/1964
- [x] #1760 
- [x] https://github.com/expressjs/expressjs.com/issues/1780 
- [x] https://github.com/expressjs/expressjs.com/issues/1755
- [x] https://github.com/expressjs/expressjs.com/issues/1754
- [x] https://github.com/expressjs/expressjs.com/issues/1753
- [x] https://github.com/expressjs/expressjs.com/issues/1731
- [x] https://github.com/expressjs/expressjs.com/issues/1704

Please review and share your feedback! I’m open to making further improvements as needed.

bjohansebas:
Hi @ShubhamOulkar, there are things I like and others I don't, so it's best if you open separate PRs for any changes you want to make.

ShubhamOulkar:
@bjohansebas, Can you list out what you don't like in the [design](https://expressjsnewdesign.netlify.app/)?


> it's best if you open separate PRs for any changes you want to make.

I am thinking of dividing into three PR's. Some of the issues are interlinked like (#1780, #1731, #1704), (#1755, #1754), (#1753). Let me know if I am missing any other issue interlinked to these issues.

PR #1965, opened 2025-07-07
feat(#1932): implement smooth scrolling for anchor navigation across site

### LINKED ISSUE #1932: Feature Request: Implement Smooth Scrolling
### Description
Currently, scrolling between sections or anchor links on the page is abrupt and not user-friendly. This feature request proposes implementing smooth scrolling behavior across the application to improve user experience and visual flow.

### Expected Behavior
- When navigating to anchors (e.g., #section-id), the scroll should transition smoothly.
- Applicable to both in-page navigation and programmatic scrolling.

Mohit5Upadhyay:
## Description
This PR implements smooth scrolling for in-page navigation across the Express.js website as mentioned in Issue : #1932 

This improves the user experience by ensuring that navigation to sections via anchor links transitions smoothly, rather than jumping abruptly.

Closes #1932 

## Video

https://github.com/user-attachments/assets/0efeb664-b772-446d-98d3-b033cf68abb8


js/app.js
```diff
@@ -1,12 +1,56 @@
...
   $('#top').click(function(e){
     $('html, body').animate({scrollTop : 0}, 500);
     return false;
   });
 
+  // Smooth scrolling for all anchor links
+  $(document).on('click', 'a[href^="#"]:not([href="#"]):not([href="#top"])', function(e) {
+    var href = $(this).attr('href');
+    if (smoothScrollToElement(href)) {
+      e.preventDefault();
```
ShubhamOulkar: Can you please remove this line https://github.com/expressjs/expressjs.com/blob/gh-pages/js/menu.js#L186 ? Since default behavior stopped here.

js/app.js
```diff
@@ -1,12 +1,56 @@
...
+  }
+
   // top link
   $('#top').click(function(e){
     $('html, body').animate({scrollTop : 0}, 500);
     return false;
   });
 
+  // Smooth scrolling for all anchor links
+  $(document).on('click', 'a[href^="#"]:not([href="#"]):not([href="#top"])', function(e) {
```
ShubhamOulkar:
```suggestion
  $("#api-doc").on('click', 'a[href^="#"]:not([href="#"]):not([href="#top"])', function(e) {
```
We should minimize the scope of this click event. Currently, $(document).on('click'...) attaches the handler to all anchor links with `#` across every page. However, smooth scrolling is only needed on API pages, since the TOC exists only there. Scoping it to `#api-doc` avoids unnecessary event handling and prevents inconsistent anchor `#` behavior on other pages. Also, checking `"#api-doc"` exist on page is required before applying click event.

ShubhamOulkar (changes requested):

bjohansebas (changes requested):
Thanks for the PR!

As i mentioned in the issue, this can be handled through CSS, javascript is not necessary.

ShubhamOulkar:
> this can be handled through CSS,

I am OK with CSS only solution. But this approach comes with some limitations, like we cannot control scroll behavior and can not add header offsets. But it is fine if headings are not hiding behind the header bar on anchor link click.

Mohit5Upadhyay:
Thanks @bjohansebas for the suggestion about a CSS-only solution!
Unfortunately, I think for this project, CSS alone isn’t sufficient because:
- IDs like #app.METHOD `(http://localhost:4000/en/5x/api.html#app.listen)` contain dots, which CSS can’t reliably scroll to without custom JS.
- We need a precise scroll offset (120px) to avoid headings hiding behind the navbar.

So I’d propose keeping the JS smooth scroll, scoped only to `#api-doc` to avoid global impact as per suggested by @ShubhamOulkar.

Let me know if you’d prefer another approach!
Thanks again for your feedback.

bjohansebas:
>But it is fine if headings are not hiding behind the header bar on anchor link click.

The headings are never hidden by the header bar, or at least I’ve been trying to reproduce it and haven’t been able to.

> IDs like #app.METHOD (http://localhost:4000/en/5x/api.html#app.listen) contain dots, which CSS can’t reliably scroll to without custom JS.

That doesn’t matter in CSS, since the effect will be applied to all elements anyway.
```css
* {
  scroll-behavior: smooth;
}
```

Mohit5Upadhyay:
Hi @bjohansebas and @ShubhamOulkar,

I’ve updated the implementation to follow your suggestion of a CSS-only smooth scrolling solution.
- I removed all the custom JavaScript smooth scroll code I previously added.
- Instead, I’ve applied this in CSS:
```
* {
  scroll-behavior: smooth;
}

h1, h2, h3, h4, h5, h6,
[id] {
  scroll-margin-top: 120px;
}

@media (prefers-reduced-motion: reduce) {
  html {
    scroll-behavior: auto;
  }
}
```

Let me know if you’d like any further changes!
Thanks for your time! 🙏

css/style.css
```diff
@@ -1,6 +1,20 @@
...
   box-sizing: border-box;
   -webkit-tap-highlight-color: transparent;
+  scroll-behavior: smooth;
+}
+
+/* handle header offset with scroll-margin */
+h1, h2, h3, h4, h5, h6,
+[id] {
+  scroll-margin-top: 120px;
+}
```
bjohansebas: I'm not sure if this is actually doing anything, and if it's supposed to have the same effect as in the MDN example https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-margin-top, then I'm -1 on this, I don't think we should do it.

ShubhamOulkar: | without scroll top margin | with scroll top margin |
|------|-------|
|image|image|

We can clearly see that margin spaces added to the header on scroll from the top of the viewport.
This property is used to increase visible area around the target element so that it is accessible to the user. 

In our context, why it is important? 
I expressed my concerns over the issue of hiding headers behind the top navigation bar in a comment above. 
> But it is fine if headings are not hiding behind the header bar on anchor link click.

This happens if top navigation is positioned sticky at the top. See image, `express.raw()` hide behind sticky top navigation.
image

It is not required right now, without scroll top margin, header scroll into viewport at top 0px. But I think adding some small margin 60px to the top looks good for me. Also our header is 57px height. Maybe I will go with sticky top navigation bar in #1943. 
image

[see scroll_padding_and_margin](https://web.dev/articles/css-scroll-snap#scroll_padding_and_margin)

css/style.css
```diff
@@ -1,6 +1,20 @@
 * {
   box-sizing: border-box;
   -webkit-tap-highlight-color: transparent;
+  scroll-behavior: smooth;
+}
+
+/* handle header offset with scroll-margin */
+h1, h2, h3, h4, h5, h6,
+[id] {
+  scroll-margin-top: 120px;
```
```suggestion
  scroll-margin-top: 60px;
```
see comment https://github.com/expressjs/expressjs.com/pull/1965/files#r2206464995

css/style.css
```diff
@@ -264,11 +278,6 @@ a {
   padding-top: 40px;
 }
 
-#api-doc *:target, #page-doc *:target {
```
ShubhamOulkar, line 261: I support removing this
since we are using scroll top margin only on headers.

css/style.css
```diff
@@ -1,6 +1,20 @@
...
+}
+
+/* handle header offset with scroll-margin */
+h1, h2, h3, h4, h5, h6,
+[id] {
+  scroll-margin-top: 120px;
+}
+
+/* Respect user motion preferences for access */
+@media (prefers-reduced-motion: reduce) {
```
ShubhamOulkar, line 14: It is Good practice to consider accessibility
+1

ShubhamOulkar (changes requested):
Please do remove this line https://github.com/expressjs/expressjs.com/blob/gh-pages/js/menu.js#L186. On TOC main section heading click, it should scroll to main sections on the page. Currently, it is disabled. Only subsections get scrolled.

Mohit5Upadhyay:
Hi @bjohansebas and @ShubhamOulkar,

Thanks so much for the thoughtful reviews!

Regarding `scroll-margin-top`:
I agree with @ShubhamOulkar that it’s helpful to prevent headings from hiding behind the sticky nav bar.
> But I think adding some small margin 60px to the top looks good for me. Also our header is 57px height. Maybe I will go with sticky top navigation bar in https://github.com/expressjs/expressjs.com/issues/1943.

I’ll adjust the value to 60px as suggested, since our header height is ~57px.

Remove `event.preventDefault()` in menu.js so main TOC sections scroll as expected.

Let me know if there’s anything else you’d like tweaked!
Thanks again for your guidance.

ShubhamOulkar (commented):
I just tested, but default CSS scroll is too fast. On clicking main menu, the page scrolls really fast to that menu. Should we try putting some animation key frames to add some delays to scrolls?

Mohit5Upadhyay:
hi @ShubhamOulkar,

Thank you for the suggestion regarding adding delays to scroll using animation keyframes.

I explored the possibility of controlling scroll speed using CSS keyframes alone. However, after testing locally, it turns out that CSS keyframes cannot manipulate scroll behavior that we need. 

While `scroll-behavior: smooth` provides a basic smooth scroll effect, it doesn’t allow customization of speed or delay — and keyframes don’t apply to scroll position singly to achieve this. 

To achieve the effect you’re suggesting (slower or delayed scrolling), we would need to implement a small JavaScript solution that programmatically controls the scroll animation.

Please let me know if you'd like me to proceed with a JavaScript-based implementation to match the desired behavior, or if you have any other suggestions — I’d be grateful to explore them!

Thanks!

bjohansebas:
> Please let me know if you'd like me to proceed with a JavaScript-based implementation to match the desired behavior, or if you have any other suggestions — I’d be grateful to explore them!

I’m -1 on implementing this in JavaScript. I want to simplify things as much as possible, and I don’t see it as reasonable to maintain that JavaScript when the platform itself already provides this solution (it might not be perfect, but it’s better than having to maintain JavaScript).

js/menu.js
```diff
@@ -183,7 +183,6 @@ toggleBtn?.addEventListener("click", (e) => {
 // Open/Close sub TOC content on click
 document.querySelectorAll("#menu > li > a").forEach((link) => {
 	link.addEventListener("click", function (event) {
-	  event.preventDefault(); // stop navigation to submenu
```
bjohansebas, line 189: why?

bjohansebas: This needs to stay. I've been testing, and on mobile it doesn’t work correctly, with this line, it works.

ShubhamOulkar: This line stop scrolling to submenu section headings. E.g. Request,  Response,  etc. 
Actually here two scrolling actions happening inside TOC and on the page.

ShubhamOulkar: Yeh on mobile it is very bad experience

ShubhamOulkar: We can stop this on small screens by using line 4 menu.js. Let's focus on desktop. It is important for desktop experience. 
 @Mohit5Upadhyay could you try to control behavior in side menu bar? Slow down submenu opening ans scrolling. I think we can do this in CSS. 😃

bjohansebas (approved):
LGTM

ShubhamOulkar (approved):
Need to remove line no 5 from app.js. Otherwise LGTM 👍

Issue #1973, opened 2025-07-16
Review current HTTP Headers, DNS, etc...


UlisesGascon:
So far seems like we can invest some time to improve several things:

- The HTTP headers in the website and discuss if we want to apply headers like: Content Security Policy, Strict Transport Policy, X-Content-Type-Options, X-Frame-Options, X-XSS-Protection...
- CA Authorization in the TLS layer 
- Enable HSTS
- Add a `Security.Txt`file pointing to the current project security policy?

I used [Web Check](https://web-check.xyz/check/https%3A%2F%2Fexpressjs.com) to do a fast review, so this is not yet an exhaustive list

bjohansebas:
- Enable HSTS:

While transferring the DNS to a Cloudflare account where both the TC and the foundation could have access in https://github.com/expressjs/expressjs.com/issues/955, it wasn't activated. Maybe @wesleytodd or @expressjs/docs-captains know why it wasn't activated? I assume it's because it's not strictly necessary, but it would be a nice plus to have it.

- Security.txt: I think this is a good thing to have. There are no side effects to adding it. I'm going to open a PR to add it (ref: https://securitytxt.org/).

- Robot.txt: Should we add it? Does it make any difference having it or not? Either way, if it's added, the value would be `User-Agent: *`, and we don’t want Google to stop crawling our pages.

- HTTP headers: The following headers are the ones recommended by Cloudflare when I ran the analysis with https://radar.cloudflare.com/scan, none of them have a value at the moment

  - Strict-Transport-Security: This is the HSTS, see the comment above.
  - X-Frame-Options: This should be set to [sameorigin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Frame-Options#sameorigin). It's not like we're going to render our own page within another page of ours, but if that case ever comes up, we would need to change this header anyway. So allowing it from the same origin should be fine. The other option would be [deny](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Frame-Options#deny). Either option is fine with me, though I prefer [sameorigin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Frame-Options#sameorigin).

  - X-Content-Type-Options: Should we set it to [nosniff](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Content-Type-Options#nosniff)? I don't think so, we should leave this value as it is.

[...]

PR #1974, opened 2025-07-16
docs: add security.txt

bjohansebas:
ref: https://github.com/expressjs/expressjs.com/issues/1973

preview: https://deploy-preview-1974--expressjscom-preview.netlify.app/.well-known/security.txt

cc: @expressjs/security-wg


ctcpip (changes requested):
shouldn't we try to comply with [the spec](https://www.rfc-editor.org/rfc/rfc9116#name-example-of-an-unsigned-secu)?

.well-known/security.txt
```diff
@@ -0,0 +1,2 @@
+Contact: express-security@lists.openjsf.org
+Policy: https://github.com/expressjs/express/security/policy
```
ShubhamOulkar:
```suggestion
# Our security policy
Policy: https://github.com/expressjs/express/security/policy
```

security.txt
```diff
@@ -0,0 +1,2 @@
+Contact: express-security@lists.openjsf.org
```
ShubhamOulkar:
```suggestion
# Our security contact
Contact: express-security@lists.openjsf.org
```

ShubhamOulkar (changes requested):

bjohansebas:
- encryption: as far as I know, we don't have a PGP key for emails, so it's not needed unless we want to start using it.
- Acknowledgments: we don't have that page, do we want it?
- Canonical: this is not necessary, this is the source of the security.txt file.
- expire: we don't need it, because why would we set an expiration date on this content? We will always keep it updated, even though the specification says it's required.
- hiring: we don't need it, we are not hiring, but we are always looking for collaborators
- Preferred-Languages: if not specified, it defaults to English. Do we want to add more languages? We are a diverse group in terms of languages.

UlisesGascon:
> encryption: as far as I know, we don't have a PGP key for emails, so it's not needed unless we want to start using it.

If any reporter requests PGP encryption, we can accommodate them using our personal PGP keys. However, we don’t have a shared/team key at this time.

> Acknowledgments: we don't have that page, do we want it?

Personally, I like the idea. It would add an extra step for each report, but many reporters are doing excellent work and I think it’s worth the effort to recognize them publicly. Should we bring this up for discussion in the security working group?

> Preferred-Languages: if not specified, it defaults to English. Do we want to add more languages? We are a diverse group in terms of languages.

I think English is the best option to simplify report digestion

> expire: we don't need it, because why would we set an expiration date on this content? We will always keep it updated, even though the specification says it's required.

I am afraid that this is mandatory in the spec (https://www.rfc-editor.org/rfc/rfc9116#name-expires). We don’t expect this information to become stale, but the specification says the Expires field must always be present and recommends that the value be less than a year into the future to avoid staleness.

To comply with this requirement, we use one of the following approaches:

- Option A: Set a fixed date like the end of each calendar year (for example, December 31, 2025) and use a reminder system to update it annually.
- Option B (preferred): Automatically generate this field with a rolling expiration, such as today plus 180 days, so it always stays within the recommended freshness window.

We avoid using long-term future dates like the year 2099, since that would technically comply but go against the intent of keeping the file current and accurate.

We can do the automation in the future, so we can land this PR soon.

security.txt
```diff
@@ -0,0 +1,4 @@
+# Our security contact
+Contact: express-security@lists.openjsf.org
```
UlisesGascon: The value MUST follow the URI syntax described in RFC 3986. Email addresses must use the `mailto:` URI scheme.
Ref: https://www.rfc-editor.org/rfc/rfc9116#section-2.5.3

```suggestion
Contact: mailto:express-security@lists.openjsf.org
```

ShubhamOulkar:
> * expire: we don't need it, because why would we set an expiration date on this content? We will always keep it updated, even though the specification says it's required.

Instead of setting an expiration date, I'd prefer to define the `scope of the security.txt` file for specific domains.

> *Acknowledgments: we don't have that page, do we want it?

Tracking in discussion is a good idea. My personal opinion, we should not do it in open environment.

bjohansebas:
> Instead of setting an expiration date, I'd prefer to define the scope of the security.txt file for specific domains.

@ShubhamOulkar I don't quite understand this idea.

> Personally, I like the idea. It would add an extra step for each report, but many reporters are doing excellent work and I think it’s worth the effort to recognize them publicly. Should we bring this up for discussion in the security working group?

Yes, please bring the discussion to the security team. This decision would be outside the scope of the documentation team.

> Option B (preferred): Automatically generate this field with a rolling expiration, such as today plus 180 days, so it always stays within the recommended freshness window.

I can work on this new script, I enjoy automating things.

ShubhamOulkar:
We should place the "security.txt" file under the "/.well-known/" path, e.g., https://example.com/.well-known/security.txt as per RFC8615 of a domain name. Ref: https://www.rfc-editor.org/rfc/rfc9116#name-location-of-the-securitytxt

> I don't quite understand this idea.

Its main aim is to define the process of reporting security vulnerabilities.
https://www.rfc-editor.org/rfc/rfc9116#name-scope-of-the-file

bjohansebas:
I created a script to update the expiration date. 

For the Acknowledgments field, i created this issue for discussion https://github.com/expressjs/expressjs.com/issues/2132

.well-known/security.txt
```diff
@@ -0,0 +1,6 @@
+# Our security contact
+Contact: mailto:express-security@lists.openjsf.org
+# Our security policy
+Policy: https://github.com/expressjs/express/security/policy
+# Expires this policy
+Expires: 2025-12-31T00:00:00Z
```
bjohansebas, line 6: I left this date to verify that the automation works and opens the PR correctly.

.github/workflows/update-external-docs.yml
```diff
@@ -10,6 +10,34 @@ permissions:
...
+            - name: Set up Node.js
+              uses: actions/setup-node@v6
+
+            - name: Update Security Expiry Docs
+              working-directory: .github/scripts
+              run: |
+                node ./update-security-expiry.mjs
+
+            - name: Create Pull Request
+              uses: gr2m/create-or-update-pull-request-action@v1
```
efekrskl, line 34: Could you pin the versions of the actions?

bjohansebas: I’ve preferred not to pin actions in this repository, except for the deploy workflow, which is the one that needs the most care. Pinning actions only causes Dependabot to open more PRs updating hashes, and verifying the changes from one hash to another is complicated, something I prefer to avoid to save us extra work.

https://github.com/expressjs/discussions/blob/master/docs/adr/366-action-pinning.md

efekrskl (approved):

bjohansebas:
cc: @expressjs/security-wg

jonchurch (commented):
I’m generally cautious about automations that open PRs, even though we do already use that pattern in this repo. My first instinct here was a much simpler solutio, a non-blocking CI check validates if the expiry is ~30 days away.

### But to review the approach w/ PR creation:
I’d prefer not to add new script infrastructure for something this small. The current approach runs a script which opaquely handles both validation and bumping the date, when really this should be two things: 

* do we need to update the expiry soon? 
* And then (a human) validating the security.txt and bumping the expires

I don't think what this job is doing is obvious from the GHA, you have to read the script to know what this job is really doing.

We can gate the job and perform the bump entirely inline in GitHub Actions without introducing a bespoke JS parser. For example:

```yaml
# Gate: is the expiry coming up?
- name: Check security.txt expiry
  id: expiry
  run: |
    expires=$(grep -i '^Expires:' .well-known/security.txt | sed 's/.*: //')
    if [ "$(date -d "$expires" +%s)" -le "$(date -d '+30 days' +%s)" ]; then
      echo "bump=true" >> "$GITHUB_OUTPUT"
    fi

- name: Bump Expires
  if: steps.expiry.outputs.bump == 'true'
  run: |
    sed -i "s/^Expires:.*/Expires: $(date -u -d '+180 days' +%Y-%m-%dT%H:%M:%SZ)/I" \
      .well-known/security.txt

- name: Create Pull Request
  if: steps.expiry.outputs.bump == 'true'
  uses: gr2m/create-or-update-pull-request-action@v1
```

This keeps the automation surface very small and avoids adding one off scripts in the repo. And if something breaks in this, it breaks loudly, and without us having to add any error handling. Which is desireable IMO.

Code is cheap to produce nowadays, but I still strongly prefer producing, reviewing, and maintaining as little as possible. 

### On intent:
Separately from the mechanics, I want to preserve the spirit of the Expires field. It’s meant to be an explicit revalidation that the contact and disclosure information is still accurate, not just a date that gets mechanically refreshed.

If we do go the PR based route, I’d strongly prefer the PR description explicitly ask reviewers to confirm the contents of security.txt, not just merge to extend the date. This is essential IMO for if none of us are around who understand this file anymore. (requested that change in https://github.com/expressjs/expressjs.com/pull/1974#discussion_r2673781848)

.github/workflows/update-external-docs.yml
```diff
@@ -10,6 +10,34 @@ permissions:
...
+              env:
+                GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+              with: 
+                commit-message: 'docs: update security expiry docs'
+                title: 'docs: update security expiry docs'
+                body: >
+                  This auto-generated PR updates the security expiry docs in the expressjs.com repository. 
+                  
+                  cc: @expressjs/security-wg
```
jonchurch:
```suggestion
                 body: >
                  This PR extends the `Expires` date in `.well-known/security.txt`.

                  Before merging, please explicitly confirm that:
                  - The security contact listed is still correct and actively monitored
                  - The linked security policy is still accurate
                  - The project’s preferred vulnerability disclosure path has not changed

                  Only merge if the above information is still valid; otherwise, update it before extending the expiration.

                  cc: @expressjs/security-wg
```

jonchurch (changes requested):

bjohansebas:
> This keeps the automation surface very small and avoids adding one off scripts in the repo. And if something breaks in this, it breaks loudly, and without us having to add any error handling. Which is desireable IMO.

I have to say I’m not good at Bash, which is why I didn’t do it that way, but your solution seems better and simpler to me.

ShubhamOulkar (approved):

bjohansebas:
thanks @jonchurch !

.github/workflows/update-external-docs.yml
```diff
@@ -10,6 +10,47 @@ permissions:
...
+    update-security:
+        runs-on: ubuntu-latest
+        if: github.repository_owner == 'expressjs'
+        steps:
+            - uses: actions/checkout@v6
+
+            - name: Check security.txt expiry
+              id: expiry
+              run: |
+                set -eo pipefail
```
jonchurch, line 22: Love me some pipefail
https://gist.github.com/mohanpedala/1e2ff5661761d3abd0385e8223e16425

jonchurch (commented):
I pushed an update that swaps this to using bash, sed, and grep to accomplish the expires check and then update expires if needed

jonchurch (approved):

bjohansebas:
Hmm, the file can’t be retrieved. I’m not sure if it’s due to GitHub Pages, Cloudflare, or Jekyll itself.

PR #1934, opened 2025-05-30
correct lang code for es and zh-cn

ShubhamOulkar:
Introduced in https://github.com/expressjs/expressjs.com/pull/1874

PR #1975, opened 2025-07-16
build: apply language based on folder. 

bjohansebas:
closes https://github.com/expressjs/expressjs.com/pull/1934

This PR updates how language metadata is applied to files in the site. Instead of using the `lang:` front matter variable within individual files, the configuration now applies language settings based on file paths using Jekyll’s `_config.yml scope and values` functionality.


bjohansebas:
It seems everything is working fine, @expressjs/docs-wg  please review it.

Makefile
```diff
@@ -16,7 +16,7 @@ help: ## This help
 
 serve: ## Local server
 	@echo "${BLUE}Starting expressjs.com at http://localhost:4000${RESET}"
-	docker run -p 4000:4000 -p 35729:35729 -v $(DIR):/usr/src/app expressjs.com bundle exec jekyll serve --host 0.0.0.0 --incremental
+	docker run -p 4000:4000 -p 35729:35729 -v $(DIR):/usr/src/app expressjs.com bundle exec jekyll serve --host 0.0.0.0
```
bjohansebas, line 19: There is a bug where dynamic texts are not loading correctly, so I'm going to remove this, since that fixes it.

css/style.css
```diff
@@ -9,23 +9,8 @@ body {
...
-  grid-area: header;
-}
-
-.content {
-  grid-area: content;
-}
-
-footer {
-  grid-area: footer;
+  display: flex;
```
ShubhamOulkar: These changes nothing to do with this PR. By the way, I don't find any issue with grid on the Home page. Except for `class=home` on the `main` element. 
It is better to do in separate PR.

ShubhamOulkar (changes requested):

_config.yml
```diff
@@ -14,6 +14,61 @@ defaults:
...
+    values:
+      lang: ko
+  -
+    scope:
+      path: "pt-br"
+    values: 
+      lang: "pt-br"  
+  -
+    scope:
+      path: "zh-ch"
```
ShubhamOulkar:
```suggestion
      path: "zh-cn"
```

_config.yml
```diff
@@ -14,6 +14,61 @@ defaults:
...
+    values: 
+      lang: "zh-ch"
```
ShubhamOulkar:
```suggestion
      lang: "zh-cn"
```

ShubhamOulkar (changes requested):
We need to document this approach.

bjohansebas:
> We need to document this approach.

I'll do it in another PR later, updating more information about Crowdin and all of that.

css/style.css
bjohansebas: imagen

The page breaks without this, it only happens on the landing pages, so this change is necessary. I don't have much time to look for more solutions, and I'm trying to fix all the issues with Crowdin to make an announcement.

bjohansebas: imagen

bjohansebas: I know this shouldn't happen, but it does, so I'm going to add the change again, and any improvements can be addressed in another PR.

ShubhamOulkar: This footer goes up because of class="content" is not present on home page. But we have class="home". Also, documentation outdated box appears on translation page. So this breaks page layout. You can verify it by removing popup box, layout works. This breaks because more than one box appears in grid layout, which is not defined in grid layout. We should fix it another PR. I will open issue for this.

ShubhamOulkar:
> I'll do it in another PR later, updating more information about Crowdin and all of that.

Need to how `page.lang` will work in code base. OR if any Jekyll doc link about this setup.

bjohansebas:
> Need to how page.lang will work in code base. OR if any Jekyll doc link about this setup.

see https://jekyllrb.com/docs/configuration/front-matter-defaults/

ShubhamOulkar:
You can remove CSS changes from this PR. It got fixed in #1978. Since we are changing website build process in current PR. It is better to keep it separate. Also that is CSS issue, nothing to do with build changes.

ShubhamOulkar (approved):
LGTM 👍

I don't know why `404.md` is not listed here? If changes got removed then please add it.

image

bjohansebas:
The 404 files can't be translated, which is why they are not uploaded, since Jekyll only serves the root 404 page.

Edit: Or at least that’s what I remember happening when the 404 page was integrated.

bjohansebas:
@ShubhamOulkar Thank you for the help.

PR #1978, opened 2025-07-18
fix: grid layout and refactor css

### LINKED ISSUE #1977: Grid layout breaks on translation home pages
imagen

This happens because grid is defined for header, content and footer child elements. But our home page do not have content class as child element but has home class. Also grids not defined for translation box that appears on translation home pages. So this breaks our layout, and grid deciding orientation of child elements on its own. 

> Note: Current layout works perfect for api and blogs pages.

_Originally posted by @bjohansebas in https://github.com/expressjs/expressjs.com/pull/1975#discussion_r2214523439_

ShubhamOulkar:
closes #1977 

|before|after|
|----|----|
|image|image|

### DIFF
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -14,7 +14,7 @@
     <div id="overlay"></div>
 
     {% if page.lang != 'en' %}
-         <div id="i18n-notice-box" class="doc-box doc-notice">
+         <div id="i18n-notice-box" class="doc-notice">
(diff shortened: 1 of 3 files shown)


bjohansebas (approved):

bjohansebas:
Fast-track it so we can proceed with the PR that fixes the i18n.

PR #1980, opened 2025-07-20
docker: match GH page ruby version

jonchurch:
## The Problem

We are behind a single patch for ruby version in the dockerfile meant for local dev of the site and the `.ruby-version` (used for netlify builds, or building from the directory), compared to GH pages official version.

## The Solution

Bump both to be 3.3.4 to be in line with GH pages.

Drop the sha pinning, to allow the dockerfile to pull the version for the correct platform.

## Context
In https://github.com/expressjs/expressjs.com/pull/1916 dependabot wanted to bump the ruby version up to beyond what GH pages uses, which made me check the GH pages latest version. The goal of the dockerfile is to ease local dev, and to match the environment that GH pages uses, so I closed the dependabot PR. Dependabot will still try to update it later as patches come out unless we tell it not to. GH updates these like maybe once a year? But the GH pages version will always be behind what dependabot wants.


bjohansebas (approved):
LGTM. I hope that someday we can stop depending on GitHub Pages dependencies. That’s why we now use workflows to deploy to production and avoid relying on those versions that GitHub used. However, I’m not sure it’s worth starting to update that. I say this because, with the redesign, I’d really prefer that we migrate to another technology that we all know, since most of us know JavaScript, and we are a Node.js framework, so we know JavaScript. I have no problem continuing to wrestle with Ruby, but I’m definitely much happier developing with javascript (react, astro, nextjs, etc).

PR #1983, opened 2025-07-22
Fix unclickable submenus on touchscreen tablet

### LINKED ISSUE #1170: Dropdowns missing on touchscreen tablet
None of the dropdowns show up on touchscreen. This is a problem because it prevents me from accessing a lot of pages.

On properly working websites, the dropdowns appear either when I just tap on the menu, or at least when I tap on the menu and drag outside it.

The workaround is that they do show up when I plug in a mouse or touchpad, or when I enable device mode in devtools.

I am using `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.116 Safari/537.36` on a Surface Pro 7.

danielzgtg:
Fixes: #1170

It works by cancelling the onclick URL change for the current (bold) parent menu item. This matches the mobile experience. Touchscreens can now tap a menu to go there (preserving current UX), tap again to open the submenu, then tap the submenu item. Previous behavior with mice is unchanged.

# Before

before

# After

after

### DIFF
--- a/js/menu.js
+++ b/js/menu.js
@@ -20,7 +20,10 @@ let activeDrawer // active dropdown nav link
 
 
 for (const el of itemsMenu) {
-	el.addEventListener("click", () => {
+	el.addEventListener("click", (e) => {
+		if (e.target.parentNode === el && e.pointerType === "touch" && el.querySelector("a.active")) {
+			e.preventDefault(); // Tapping active menu on touchscreen tablet should open menu, not click
+		}
 		if (isSmallScreen || 'ontouchstart' in document.documentElement) {
 		// HANDLE ACTIVE LINKS IN DROPDOWN NAV	
 		// if no activeDrawer set then page was set in md logic


_includes/header.html
```diff
@@ -13,12 +13,12 @@
...
     <ul id="navmenu">
       <li>
-        <a href="/" id="home-menu" {% if page.menu=='home' %} class="active" {% endif %}>
+        <a id="home-menu" {% if page.menu=='home' %} class="active" {% else %} href="/" {% endif %}>
           {{ site.data[page.lang].menu.home }}
         </a>
       </li>
       <li id="getting-started-menu" class="submenu">
-        <a href="/{{ page.lang }}/starter/installing.html" {% if page.menu=='starter' %} class="active" {% endif%}>
+        <a {% if page.menu=='starter' %} class="active" {% else %} href="/{{ page.lang }}/starter/installing.html" {% endif%}>
```
ShubhamOulkar:
```suggestion
        <a {% if page.menu=='starter' %} class="active" {% else %} href="/{{ page.lang }}/starter/installing.html" {% endif%}>
```
I do not support removing `href` if link is active. It is bad practice. This breaks a11y features. Not having an href is semantically incorrect for anchor tags

ShubhamOulkar (changes requested):
See the comment below.
 
I’d like to highlight that, by default, :hover events generally do not fire on touchscreen devices, as there's no concept of "hover" with touch — users either tap or don’t. This behavior is documented on MDN.

However, some hybrid devices (like the Microsoft Surface) may trigger hover events when using a mouse or trackpad — but not via touch.

To ensure consistent interactivity across all devices, we should either:

    - reuse/modify the existing click event from menu.js for both desktop and mobile, or

    - add support for touch events where needed.

> Note: We doing redesign of top navigation bar, this issue will get removed there.

danielzgtg:
> by default, :hover events generally do not fire on touchscreen devices

In addition to the tablet, hover events work on my Android phone using both Chrome and Firefox.

> we should either

I finally found the listener on menu.js:23. Tomorrow, I will:
* Revert the HTML changes
* Split the bold fixes into another PR
* preventDefault for non-mouse pointerType after line 23

js/menu.js
```diff
@@ -20,7 +20,10 @@ let activeDrawer // active dropdown nav link
 
 
 for (const el of itemsMenu) {
-	el.addEventListener("click", () => {
+	el.addEventListener("click", (e) => {
+		if (e.target.parentNode === el && e.pointerType !== "mouse" && el.querySelector("a.active")) {
```
ShubhamOulkar:
```
  e.pointerType !== "mouse"
```
This should solve the problem with pointer type `pen` and `touch`, right?

OR should we just use?
```
  e.pointerType === 'touch'
```
I am asking this because I don't want to affect touch pen users(ipads).

danielzgtg: Fixed.

Yes, you're right that pen users have hover and don't need this, as I tested just now with a Wacom tablet.

js/menu.js
```diff
@@ -20,7 +20,10 @@ let activeDrawer // active dropdown nav link
 
 
 for (const el of itemsMenu) {
-	el.addEventListener("click", () => {
+	el.addEventListener("click", (e) => {
+		if (e.target.parentNode === el && e.pointerType === "touch" && el.querySelector("a.active")) {
```
ShubhamOulkar, line 24: LGTM 👍

Thank you for helping us to fix this after long time 👍.
😃It is not required, Can you update PR description with some live visuals(gif) from your screen? So that other peoples can understand this issue.

ShubhamOulkar (approved):

js/menu.js
danielzgtg: Updated with before/after videos. The usability difference is there though the taps are invisible as touchscreens don't show the cursor.

ShubhamOulkar: description says not supported MIME type.

danielzgtg: @ShubhamOulkar What browser and device? I normally encode to the modern and patent-free AV1 format, which should work unless on an old Mac/iPhone. I've reencoded to gif for your convenience, but that outdated format has less compression and no pause button.

ShubhamOulkar: no problems now it is visible. By the browser was firefox 141.0.

ShubhamOulkar (commented):
LGTM 
🎉 awesome shout-out to @danielzgtg! 🙌

bjohansebas (approved):
I can't reproduce it, but if the error is fixed, it's fine by me. I don't have a tablet, and on the phone the size isn't enough to check it. The code looks fine.

LGTM

PR #1985, opened 2025-07-23
ci: run lighthouse report

ShubhamOulkar:
### Lighthouse Report
image

[Download lighthouse report](https://github.com/expressjs/expressjs.com/actions/runs/16548339611/artifacts/3623171482) 

ref: https://github.com/GoogleChrome/lighthouse?tab=readme-ov-file#using-the-node-cli

perf: https://pagespeed.web.dev/analysis/https-expressjs-com-en-api-html/wld1auit0n?form_factor=mobile

### DIFF
--- a/.github/workflows/lighthouse.yml
+++ b/.github/workflows/lighthouse.yml
@@ -103,13 +103,16 @@ jobs:
           path: |
             lighthouse-report.md
 
-      - name: 💬 Comment on PR
+      - name: 💬 Comment on PR with Lighthouse results
         uses: actions/github-script@v7
         with:
           script: |
+            const fs = require('fs');
+            const report = fs.readFileSync('lighthouse-report.md', 'utf8');
+
             github.rest.issues.createComment({
               issue_number: context.payload.pull_request.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
-              body: `${{ steps.format.outputs.comment }}`
+              body: report
             });
--- a/.github/workflows/lighthouse.yml
+++ b/.github/workflows/lighthouse.yml
@@ -1,7 +1,7 @@
-name: Lighthouse report
+name: Lighthouse audit
 
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, synchronize]
 
 permissions:
@@ -17,7 +17,7 @@ jobs:
       - name: Checkout PR code
         uses: actions/checkout@v4
         with:
-          ref: ${{ github.event.pull_request.head.ref }}
+          ref: ${{ github.event.pull_request.head.sha }}
           repository: ${{ github.event.pull_request.head.repo.full_name }}
 
       - name: Setup Node.js with npm cache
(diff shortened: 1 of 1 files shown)


UlisesGascon:
Great job @ShubhamOulkar!  In the past I’ve used Unlighthouse to generate reports from scratch and upload them to Netlify. We could even set up a secondary project for that. Here's the link I used: https://unlighthouse.dev/integrations/ci#github-actions-netlify-example

Your approach looks really solid as well. Nice work! 👍

bjohansebas:
I have my doubts about how important it really is to have Lighthouse, but I think we could give it a try. What I would like, and this would be a requirement for me to accept it, is that the report can be commented on the issue. As it stands, no one would notice that the report is being generated, and if no one sees it, then what’s the point of generating it? Making it accessible would be important.

> i used Unlighthouse to generate reports from scratch and upload them to Netlify

I’d prefer not to rely too much on Netlify, but if this tool works independently of where the preview is generated, it would be fine with me too.

ShubhamOulkar:
> the report can be commented on the issue. As it stands, no one would notice that the report is being generated, and if no one sees it, then what’s the point of generating it? Making it accessible would be important.
>
I tried commenting report on PR in #1984. But it keep hitting 403 error denial of access 
I don't think it is inaccessible, because we know where results are. 

>https://unlighthouse.dev/integrations/ci#github-actions-netlify-example

If this tool provide ous correct SEO results then I am +1 on this. I will try this tool.

bjohansebas:
> I tried commenting report on PR in https://github.com/expressjs/expressjs.com/pull/1984. But it keep hitting 403 error denial of access

That was a permissions issue with the actions I can help you with that if you’d like

> I don't think it is inaccessible, because we know where results are.

Yes, but not really, keep in mind that not everyone is keeping an eye on the workflows, and a comment would make them more visible. On the other hand, if the comment isn't added in regular PRs, it could easily be missed, which defeats the purpose. Having to dig into the workflow and check the results takes time, so people might just skip it, something that a simple comment would solve. That’s why the comment is almost essential.

ShubhamOulkar:
Should we override `X-Robots-Tag: noindex` in Netlify preview? 

|netlify preview SEO blocker results| actual expressjs.com SEO results|
|---------|---------|
|image|image|

.github/workflows/lighthouse.yml
```diff
@@ -0,0 +1,140 @@
+name: Lighthouse audit
+
+on:
+  pull_request_target:
+    types: [opened, synchronize]
```
bjohansebas, line 5:
```suggestion
```

bjohansebas (commented):
I still don’t understand why the workflow isn’t running, it didn’t run for https://github.com/expressjs/expressjs.com/pull/1988 either when it was using pull_request_target. It only ran when switched to pull_request.

ShubhamOulkar:
> I still don’t understand why the workflow isn’t running, it didn’t run for #1988 either when it was using pull_request_target. It only ran when switched to pull_request.

[see GITHUB_REF for pull_request in the table](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request)

[see GITHUB_REF for pull_request_target](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target). To trigger the action on PR for `pull_request_target` we need `ref` of base branch. So we need to add workflow on base branch then only it will trigger on each new PR. 

You can refer from this [repo](https://github.com/ShubhamOulkar/connectRTC/blob/main/.github/workflows/lighthouse.yml)

ShubhamOulkar:
Hi @UlisesGascon, I tried `unlighthouse` in #1990 but got same SEO results. This does not bypass https://github.com/expressjs/expressjs.com/pull/1985#issuecomment-3111782050. 

Should we do this in Netlify? https://docs.netlify.com/manage/monitoring/lighthouse/#set-up-embedded-lighthouse-scores-and-reports-for-your-site.

bjohansebas (approved):
LGTM, sorry for the delay

PR #1986, opened 2025-07-23
Fix API reference nav missing bold

danielzgtg:
This fixes the "API reference" nav missing bold. Split from #1983.

# Before

API Reference menu item missing bold

# After

API Reference menu item properly bolded


ShubhamOulkar (approved):
Thank you, @danielzgtg, for noticing and fixing this issue. I removed this without noticing this use case.

LGTM 👍

Issue #1553, opened 2024-08-01
"RequestError [HttpError]: Resource not accessible by integration" for `.github/workflows /translation.yml`


IamLizu:
As seen in #1552, the translation.yml gives the following error,

```console
RequestError [HttpError]: Resource not accessible by integration
    at /home/runner/work/_actions/actions/github-script/v5/dist/index.js:4560:21
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async eval (eval at callAsyncFunction (/home/runner/work/_actions/actions/github-script/v5/dist/index.js:4888:16), <anonymous>:17:1)
    at async main (/home/runner/work/_actions/actions/github-script/v5/dist/index.js:4943:20) {
  status: 403,
  response: {
    url: 'https://api.github.com/repos/expressjs/expressjs.com/issues/1552/labels',
    status: 403,
    headers: {
      'access-control-allow-origin': '*',
      'access-control-expose-headers': 'ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Used, X-RateLimit-Resource, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type, X-GitHub-SSO, X-GitHub-Request-Id, Deprecation, Sunset',
      'content-encoding': 'gzip',
      'content-security-policy': "default-src 'none'",
      'content-type': 'application/json; charset=utf-8',
      date: 'Thu, 01 Aug 2024 05:28:32 GMT',
      'referrer-policy': 'origin-when-cross-origin, strict-origin-when-cross-origin',
      server: 'github.com',
      'strict-transport-security': 'max-age=[31](https://github.com/expressjs/expressjs.com/actions/runs/10192586788/job/28195713099?pr=1552#step:3:32)536000; includeSubdomains; preload',
      'transfer-encoding': 'chunked',
      vary: 'Accept-Encoding, Accept, X-Requested-With',
      'x-accepted-github-permissions': 'issues=write; pull_requests=write',
      'x-content-type-options': 'nosniff',
      'x-frame-options': 'deny',
      'x-github-api-version-selected': '2022-11-28',
      'x-github-media-type': 'github.v3',
      'x-github-request-id': '3841:1809F2:10FEF4:1FA623:66AB1D00',
      'x-ratelimit-limit': '5000',
[...]

PR #1988, opened 2025-07-25
ci: add lighthouse 

bjohansebas:

### DIFF
--- a/.github/workflows/lighthouse.yml
+++ b/.github/workflows/lighthouse.yml
@@ -1,4 +1,4 @@
-name: Lighthouse Audit
+name: Lighthouse report
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/lighthouse.yml
```diff
@@ -0,0 +1,119 @@
+name: Lighthouse Audit
+
+on:
+  pull_request_target:
+    types: [opened, synchronize]
```
bjohansebas, line 5:
```suggestion
    branches:
    - gh-pages
```

.github/workflows/lighthouse.yml
```diff
@@ -0,0 +1,119 @@
+name: Lighthouse report
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+
+permissions:
+  contents: read
+  pull-requests: write
```
ShubhamOulkar, line 9: Your branch has write access, mine is not [see](https://github.com/expressjs/expressjs.com/actions/runs/16546954947/job/46796082488?pr=1985). 😃

Maybe this will not work on PR which are created from forked new branch. Should we try it by merging it?

bjohansebas: By default, all PRs created from forks don’t have permissions, which is why pull_request_target is used. We ran into this issue a long time age ,see #1553.

PR #1989, opened 2025-07-26
docs(migrating-5): add section about express.static dotfiles defaulting to ignore (#1987)

### LINKED ISSUE #1987: Migration guide should mention the `dotfiles` default change in `express.static`
### Description

The **Migrating to Express 5** guide does not mention a significant breaking change in the default behavior of the `express.static` middleware.

In Express 5, the `dotfiles` option now defaults to `ignore`. This is a change from Express 4 less strict behavior and can lead to unexpected 404 Not Found errors after migration, as files inside dot-directories (e.g., `.well-known`) are no longer served by default.

This is a common "gotcha" that can be difficult to debug for developers who are not aware of this specific change.

### Steps to reproduce

A common use case affected by this is serving files from the `.well-known` directory, which is essential for standards like Let's Encrypt (ACME), Android App Links (`assetlinks.json`), and Apple's Universal Links.

**Code that worked in Express 4 but fails silently in Express 5:**
```javascript
// In public/.well-known/assetlinks.json exists
app.use(express.static("public"));

// A request to GET /.well-known/assetlinks.json will result in a 404.
```
### Expectations

The migration guide should explicitly warn developers about this change and provide the recommended solution to prevent breaking important functionality.

I suggest adding a section `express.static.dotfiles` to the **Migrating to Express 5** documentation that highlights this change. It could include:

A clear statement that express.static.dotfiles option default is now 'ignore'.

An explanation of the impact (e.g., folders like .well-known will become inaccessible).

The recommended, secure solution for serving specific dot-directories:

```javascript
// Recommended fix for serving .well-known in Express 5
app.use("/.well-known", express.static("public/.well-known", {
  dotfiles: "allow"
}));

// Keep the secure default for all other static assets
app.use(express.static("public"));
```
This would greatly help the community and make the migration process smoother.

Thank you for your great work on Express !

VishalRMahajan:
### Summary
This PR updates the "Migrating to Express 5" guide to document the breaking change where express.static's dotfiles option now defaults to "ignore".

### Fixes Issue
Fixes #1987

### Changes
- Added a new section to migrating-5.md explaining the dotfiles default change.
- Provided example code for migration from Express 4 to Express 5.

image

### DIFF
--- a/en/guide/migrating-5.md
+++ b/en/guide/migrating-5.md
@@ -472,7 +472,7 @@ Example of breaking code:
 
 ```js
 // v4
-app.use(express.static("public"));
+app.use(express.static('public'))
 ```
 
 After migrating to Express 5, a request to `/.well-known/assetlinks.json` will result in a **404 Not Found**.
@@ -481,8 +481,8 @@ To fix this, serve specific dot-directories explicitly using the `dotfiles: "all
 
 ```js
 // v5
-app.use("/.well-known", express.static("public/.well-known", { dotfiles: "allow" }));
-app.use(express.static("public"));
+app.use('/.well-known', express.static('public/.well-known', { dotfiles: 'allow' }))
+app.use(express.static('public'))
 ```
 
 This approach allows you to safely serve only the intended dot-directories while keeping the default secure behavior for other dotfiles, which remain inaccessible.
--- a/en/guide/migrating-5.md
+++ b/en/guide/migrating-5.md
@@ -66,6 +66,7 @@ You can find the list of available codemods [here](https://github.com/expressjs/
   <li><a href="#path-syntax">Path route matching syntax</a></li>
   <li><a href="#rejected-promises">Rejected promises handled from middleware and handlers</a></li>
   <li><a href="#express.urlencoded">express.urlencoded</a></li>
+  <li><a href="#express.static.dotfiles">express.static dotfiles</a></li>
   <li><a href="#app.listen">app.listen</a></li>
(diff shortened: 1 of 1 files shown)


en/guide/migrating-5.md
```diff
@@ -462,6 +464,30 @@ Details of how Express handles errors is covered in the [error handling document
 
 The `express.urlencoded` method makes the `extended` option `false` by default.
 
+<h3 id="express.static.dotfiles">express.static dotfiles</h3>
+
+In Express 5, the `express.static` middleware's `dotfiles` option now defaults to `"ignore"`. This is a change from Express 4, where dotfiles were served by default. As a result, files and directories starting with a dot (`.`), such as `.well-known`, will no longer be accessible and will return a **404 Not Found** error. This can break functionality that depends on serving dot-directories, such as Android App Links, and Apple Universal Links.
```
bjohansebas: In Express 4, the default value allowed files inside a directory that starts with a dot.
https://expressjs.com/en/4x/api.html#dotfiles
```suggestion
In Express 5, the `express.static` middleware's `dotfiles` option now defaults to `"ignore"`. This is a change from Express 4, where dotfiles were served by default. As a result, files inside a directory that starts with a dot (`.`), such as `.well-known`, will no longer be accessible and will return a **404 Not Found** error. This can break functionality that depends on serving dot-directories, such as Android App Links, and Apple Universal Links.
```

bjohansebas (approved):
LGTM, thanks!

PR #1999, opened 2025-08-02
chore(fonts):⚡reduce font payload

ShubhamOulkar:
### page load time reduced by ~300-400ms 

|express.com|preview|
|-----|-----|
|expressjs com tree|a tree|
|expressjs com perf|a|

### ~230-270 kB reduction in font payload

|new font files|old font files|
|------|-----|
|Screenshot 2025-08-02 190411|image|

Font source: https://github.com/fontsource/font-files/tree/main/fonts/variable/open-sans/files


ShubhamOulkar:
> 🚦 Lighthouse Results (Mobile & Desktop)

Need to remove spaces to edit comment correctly. Also need to increase wait time for Netlify preview.

Happy 😃 mobile results are improving.

ShubhamOulkar:
I just realized that for improving performance on mobile needs refactoring in logic and assets file. It should be good to do it in other PR.

bjohansebas (changes requested):
We need to know where you got the fonts before accepting this PR.

.github/workflows/lighthouse.yml
```diff
@@ -20,6 +20,26 @@ jobs:
...
+          if [ -z "$NON_MD_FILES" ]; then
+            echo "Only .md files changed. Skipping job."
+            echo "skip_lighthouse=true" >> "$GITHUB_ENV"
+          else
+            echo "skip_lighthouse=false" >> "$GITHUB_ENV"
+          fi
+      
+      - name: Exit early if skipped
+        if: env.skip_lighthouse == 'true'
+        run: echo "Skipping Lighthouse audit because only .md files changed." && exit 0
```
bjohansebas: I’m not comfortable with this. pull_request_target has full permissions, and I don’t want to introduce vulnerabilities into this repository. I’m even thinking it might be better to handle the Lighthouse integration from Netlify.

Also, I feel that Lighthouse doesn’t provide very reliable data, in every PR I’ve seen it run, the results can be quite inaccurate, even when only a markdown file was changed.

ShubhamOulkar: > Also, I feel that Lighthouse doesn’t provide very reliable data, in every PR I’ve seen it run, the results can be quite inaccurate, even when only a markdown file was changed.

Our assets and website design logic causing this issue. I think we should get same result in netlify integration because netlify use `lighthouse` to generate results. [ref](https://github.com/netlify/netlify-plugin-lighthouse/blob/main/package.json#L32)

> handle the Lighthouse integration from Netlify.

I support this and suggested before in my previous PR as well, we should consider netlify option just because we don't want unnecessary CI maintenance.



> I don’t want to introduce vulnerabilities

That why I prefer `pull_request` action for simple task like lighthouse. But this action can not comment on PR's. By the way, there should not be a problem with security because I am using `sha`. I have seen #2031 this PR.

ShubhamOulkar:
> We need to know where you got the fonts before accepting this PR.

@bjohansebas, I got this woff2 fonts from https://github.com/fontsource/font-files/tree/main/fonts/variable/open-sans/files

.github/workflows/lighthouse.yml
ShubhamOulkar: I'm removing this workflow from this PR. Let's just keep this PR for fonts.

ShubhamOulkar: My latest commit got 100% results for desktop and on the edge for mobile. 😃

Phillip9587:
Fontsource provides npm packages for the fonts: https://fontsource.org/fonts/open-sans/install

ShubhamOulkar:
> Fontsource provides npm packages for the fonts: https://fontsource.org/fonts/open-sans/install

Hi @Phillip9587, we don't use any bundler in this project so that we can not use this. See [Prerequisites section](https://fontsource.org/docs/getting-started/install). Fontsource provide [CDN option](https://fontsource.org/docs/getting-started/cdn), but I prefer self-hosting fonts. Tell me if I am missing something because you're suggesting this second time I guess.

ShubhamOulkar (commented):
LGTM 👍 Please, whoever reviews it, merge it as soon as possible since it improves user experience for slower Internet connections.

fonts/open-sans/fonts.css
```diff
@@ -1,19 +1,21 @@
...
   font-style: italic;
-  font-weight: 100 700;
+  font-weight: 300 800;
   font-display: swap;
   src: local("Open Sans"),
-       url("./woff2/OpenSans-Italic.woff2") format("woff2"),
-       url("./woff/OpenSans-Italic.woff")   format("woff"),
+       url("./woff2/open-sans-latin-wght-italic.woff2") format("woff2-variations"),
+       url("./woff/OpenSans-Italic.woff")   format("woff");
+  unicode-range: U+0000-00FF,U+0131,U+0152-0153,U+02BB-02BC,U+02C6,U+02DA,U+02DC,U+0304,U+0308,U+0329,U+2000-206F,U+20AC,U+2122,U+2191,U+2193,U+2212,U+2215,U+FEFF,U+FFFD;
```
bjohansebas, line 20: Why is that? Why should we make a condition for that font to download?

ShubhamOulkar: It is for font optimization, if page don't use any of this character from this range then font will not download. This is effective when we are importing fonts from third party source like https://github.com/fontsource/font-files/tree/main/fonts/variable/open-sans/files. All fonts won't get download, only characters used by page, that fonts will get download. But in our case, we are self-hosting only two files that are used. So will not see any difference in optimization. But I decided to go with font source [importing style](https://github.com/fontsource/font-files/blob/main/fonts/variable/open-sans/wght.css).

bjohansebas (approved):
LGTM, thanks @ShubhamOulkar!

PR #2030, opened 2025-08-03
ci: fix multiple comments in Lighthouse workflow

bjohansebas:
The problem is that new comments are being created every time a new commit is added, which generates spam, and it should actually update the existing comment, but it's not working.

I really don't know why the includes is not working properly. I want to try this before disabling the workflow until it is truly fixed.

### DIFF
--- a/.github/workflows/lighthouse.yml
+++ b/.github/workflows/lighthouse.yml
@@ -2,7 +2,6 @@ name: Lighthouse audit
 
 on:
   pull_request_target:
-    types: [opened, synchronize]
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/lighthouse.yml
```diff
@@ -108,31 +108,32 @@ jobs:
...
             const { data: comments } = await github.rest.issues.listComments({
               issue_number: context.payload.pull_request.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
             });
-      
+
             const botComment = comments.find(comment =>
-              comment.user.type === 'Bot' &&
-              comment.body.includes('🚦Lighthouse Results')
```
ShubhamOulkar, line 116: This space causing issue

ShubhamOulkar (approved):
Let's test it.

ShubhamOulkar:
Got fixed 👍

PR #2034, opened 2025-08-08
docs: Improve clarity of next('route') usage in Route Handlers section

yelishgiri:
Description:
This PR adds a concrete example demonstrating the use of next('route') in the “Route handlers” section of the Express routing documentation.

Currently, the docs mention next('route') but don’t show a simple, self-contained example, which could make it unclear for new users. This change:

Adds a GET /user/:id example that uses next('route') to skip to the next matching route.
Includes a short bullet-point explanation of the behavior for two different request paths (/user/5 and /user/0).

Before:
The section explains next('route') but doesn’t show a minimal code sample demonstrating it in action.

After:
The section now contains:
```
app.get('/user/:id', (req, res, next) => {
  if (req.params.id === '0') {
    return next('route')
  }
  res.send(`User ${req.params.id}`)
})

app.get('/user/:id', (req, res) => {
  res.send('Special handler for user ID 0')
})

```

In this example:
GET /user/5 → handled by first route → sends "User 5"
GET /user/0 → first route calls next('route'), skipping to the next matching /user/:id route

Why this change improves the docs:

Makes the next('route') behavior obvious without requiring guesswork.
Matches the style and tone of existing Express documentation.
Adds a practical example to complement the textual explanation.


en/guide/routing.md
```diff
@@ -245,6 +245,26 @@ In Express 4.x, <a href="https://github.com/expressjs/express/issues/2495">the `
 
 You can provide multiple callback functions that behave like [middleware](/{{ page.lang }}/guide/using-middleware.html) to handle a request. The only exception is that these callbacks might invoke `next('route')` to bypass the remaining route callbacks. You can use this mechanism to impose pre-conditions on a route, then pass control to subsequent routes if there's no reason to proceed with the current route.
 
+You can use `next('route')` to skip the rest of a route's callbacks and pass control to the next route. For example:
```
bjohansebas: These two sentences repeat content, whereas if the message can be improved and shown through the example, we shouldn’t repeat the content that was already mentioned earlier.

yelishgiri: Thanks for pointing that out. I’ve removed the second sentence so the explanation is given only once, with the example showing the behavior directly. This should make the section more concise and avoid redundancy.

bjohansebas (approved):
LGTM, thanks

PR #2036, opened 2025-08-11
Clarify path matching behavior in Express.js documentation

### LINKED ISSUE #1892: The Path examples documentation is misleading
### Description

In the documentation for [Path examples](https://expressjs.com/en/5x/api.html#path-examples), which is linked to from every other method like `app.get` or `app.use`, the descriptions are misleading. Each of the examples starts with:

> This will match paths starting with...

Which seems to indicate that the handler set up on a route will trigger for every path that _starts_ with that string instead of only exact matches which is how it actually works.

For the first example:

> This will match paths starting with /abcd:
> 
> ```
>  app.use('/abcd', (req, res, next) => {
>    next()
>  })
> ```

As read, this seems to indicate it will trigger for routes like `/abcde` or `/abcd/efg` or `/abcd/efg/123.js` which is not the case.

I do understand there may be some small caveat here when mixing in the actual `Routing` class where it _will_ handle nested or combined paths but that should have it's own notes or mention in the documentation. As it's currently written that is not clear.

### Expectations

The path syntax should be explicit about how it works. By default the assumption after reading is that paths are exact matches unless doing something special with path patterns or wildcards or regular expressions.

kushal4345:
## Summary
Updated the path examples documentation to clarify that route matching in Express.js is exact by default, unless path patterns, wildcards, or regular expressions are used.

## Details
Previously, the docs said “This will match paths starting with…”, which could be interpreted as matching partial strings (e.g., `/abcde`) when using `app.use('/abcd', ...)`. This is misleading because Express matches exact paths unless explicitly configured otherwise.

## Change
- Reworded the examples to clearly explain the exact-match behavior.
- Added note about wildcards and regular expressions for broader matching.
Fixes #1892

### DIFF
--- a/_includes/api/en/4x/app-use.md
+++ b/_includes/api/en/4x/app-use.md
@@ -79,7 +79,8 @@ mounting middleware.
 <tr>
 <td>Path</td>
 <td markdown="1">
-This will match paths starting with `/abcd`:
+This will match the exact path `/abcd` and any paths beginning with `/abcd/` (including subdirectories), but will not match paths like `/abcde` :
+
 
 ```js
 app.use('/abcd', function (req, res, next) {
--- a/_includes/api/en/4x/app-use.md
+++ b/_includes/api/en/4x/app-use.md
@@ -79,7 +79,7 @@ mounting middleware.
 <tr>
 <td>Path</td>
 <td markdown="1">
-This will match the exact path `/abcd` and any paths beginning with `/abcd/` (including subdirectories), but will not match paths like `/abcde` :
+Matches the exact path `/abcd` and any sub-paths starting with `/abcd/` (for example, `/abcd/foo`):
 
 
 ```js
--- a/_includes/api/en/5x/app-use.md
+++ b/_includes/api/en/5x/app-use.md
@@ -80,7 +80,8 @@ mounting middleware.
 <tr>
 <td>Path</td>
 <td markdown="1">
-This will match paths starting with `/abcd`:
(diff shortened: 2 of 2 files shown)


_includes/api/en/4x/app-use.md
```diff
@@ -79,7 +79,8 @@ mounting middleware.
 <tr>
 <td>Path</td>
 <td markdown="1">
-This will match paths starting with `/abcd`:
+This will match the exact path `/abcd` and any paths beginning with `/abcd/` (including subdirectories), but will not match paths like `/abcde` :
```
bjohansebas: By saying that it’s exact, it already indicates that no other routes different from /abcd will be added to the middleware.

```suggestion
Matches the exact path `/abcd` and any sub-paths starting with `/abcd/` (for example, `/abcd/foo`):
```

_includes/api/en/5x/app-use.md
```diff
@@ -80,7 +80,8 @@ mounting middleware.
 <tr>
 <td>Path</td>
 <td markdown="1">
-This will match paths starting with `/abcd`:
+This will match the exact path `/abcd` and any paths beginning with `/abcd/` (including subdirectories), but will not match paths like `/abcde` :
```
bjohansebas:
```suggestion
Matches the exact path `/abcd` and any sub-paths starting with `/abcd/` (for example, `/abcd/foo`):
```

bjohansebas (changes requested):
Thank you for your contribution, just a small comment from my side.

kushal4345:
I’ve pushed the requested changes. Please let me know if anything else is needed

bjohansebas (approved):
LGTM, thanks @kushal4345!

PR #2039, opened 2025-08-14
Update links to 5x api docs on `use middleware` page

ShubhamOulkar:
same as title

preview:- https://deploy-preview-2039--expressjscom-preview.netlify.app/en/guide/using-middleware

### DIFF
--- a/en/guide/using-middleware.md
+++ b/en/guide/using-middleware.md
@@ -10,7 +10,7 @@ redirect_from: "/guide/using-middleware.html"
 
 Express is a routing and middleware web framework that has minimal functionality of its own: An Express application is essentially a series of middleware function calls.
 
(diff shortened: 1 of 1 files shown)


ShubhamOulkar:
I think it is better to handle dynamically update the links that are pointing to latest API docs.
So that we don't need to update on every new release.

bjohansebas (approved):
LGTM, thanks!

bjohansebas:
> I think it is better to handle dynamically update the links that are pointing to latest API docs.
So that we don't need to update on every new release.

We have this page, which contains the content of the latest API, although I’m concerned that in the future some APIs might be removed and those links could break. I’m going to open a discussion to talk about how to improve the content across versions.

https://expressjs.com/en/api.html

PR #1933, opened 2025-05-30
feat: Improve language picker UI with icons

ShubhamOulkar:
### Task 
- [X] Improved the design of the language picker by adding language icons.
- [X] removed duplicate id attributes 
- [X] [fix bug: highlighting current language](https://github.com/expressjs/expressjs.com/pull/1933/commits/c840a1f396d7b2ba8f60f51ebb38ca049f7fbdb7)
- [x] [refactor css and bug fixes](https://github.com/expressjs/expressjs.com/pull/1933/commits/34b40ed333b1023dc0c9f7af0159c554c7f8c23e)

closes #1804 

## Design 
(Following design rejected)
|desktop dark|desktop light|
|--------|-------|
|image|image|

## bug: Mobile view, current language color highlighting
image

# ~~How SVG Icons Were Created?~~
- ~~The SVG flag icons were sourced from [Nucleo SVG Flag Icons](https://nucleoapp.com/svg-flag-icons).~~
- ~~Each SVG file was copied and added to the repository using the VS Code editor.~~
- ~~While there are several ways to edit SVGs in VS Code, including using extensions for live preview or advanced editing, I used the default text editor setup to make the required changes.~~
- ~~The SVGs were customized and optimized to fit the required sizes(20x22) and design specifications for our language picker UI, so that we don't need to write CSS to customize icons.~~

PR #2040, opened 2025-08-18
feat : add new language picker component

### LINKED ISSUE #1804: Language picker: Add identifying feature and remove/fix duplicate nav ids
The language picker needs to be more distinguishable announcing it as a page feature and not just some text. As it is it just looks like text to me. Could be as simple as some kind of border, or fancier with an arrow 

Image

Also it has duplicate ids to navbar. These should be replaced, or changed to classes if they are reusing styles.

- Ids should not be duplicated in a codebase normally
- it's very confusing to see these `navbar` and `navmenu` identifiers inside the picker like this.

ShubhamOulkar:
See https://github.com/expressjs/expressjs.com/pull/1933#issuecomment-3086477541
closes #1804 

image

cc @bjohansebas


.github/workflows/lighthouse.yml
```diff
@@ -29,12 +29,12 @@ jobs:
...
           PREVIEW_URL="https://deploy-preview-${{ github.event.pull_request.number }}--expressjscom-preview.netlify.app"
           echo "PREVIEW_URL=$PREVIEW_URL" >> "$GITHUB_ENV"
-          MAX_RETRIES=12
+          MAX_RETRIES=10
           DELAY=10
 
           echo "Checking Netlify preview: $PREVIEW_URL"
           for i in $(seq 1 $MAX_RETRIES); do
-            if curl -s --head --max-time 5 "$PREVIEW_URL" | grep "200 OK" > /dev/null; then
+            if curl -s -I "$PREVIEW_URL" | grep "HTTP/.* 200" > /dev/null; then
```
ShubhamOulkar, line 37: Correcting header status string. Now lighthouse audit is not running because of wrong status checking.

_includes/language-picker.html
```diff
@@ -1,22 +1,18 @@
 <div class="desktop-lang-switcher">
-{% assign url_parts = page.url | split: '/' %}
-{% assign url_remainder = url_parts | slice: 2, url_parts.size | join: '/' %}
-{% assign current_lang = page.lang %}
```
bjohansebas: I think this should stay, otherwise we won’t be able to know the current URL to keep the page where it was and only change the language.

ShubhamOulkar: Right, thank you for pointing out.

_includes/header.html
```diff
@@ -12,11 +12,6 @@
   <nav id="navbar" aria-label="primary">
     <input id="q" placeholder="🔎 search">
     <ul id="navmenu">
-      <li>
-        <a href="/" id="home-menu" {% if page.menu=='home' %} class="active" {% endif %}>
-          {{ site.data[page.lang].menu.home }}
```
ShubhamOulkar, line 17: Translations for this is removed. This link is redundant because express logo also points to homepage.

bjohansebas (approved):
LGTM, thanks

PR #2042, opened 2025-08-18
docs: document array and multiple-args support in req.is

### LINKED ISSUE #2038: Document that req.is accepts an array
### Description

We also accept that req.is can take either an array or multiple arguments as input, giving this method more flexibility.

See https://github.com/expressjs/express/pull/6703

### Expectations

We should document this on https://expressjs.com/en/5x/api.html#req.is and https://expressjs.com/en/4x/api.html#req.is to make that information more accessible.

GauravKesh:
This PR updates the Express 4.x and 5.x API documentation to clarify that
`req.is()` accepts either an array or multiple arguments, which provides more
flexibility when checking against content types.

Fixes: #2038

### DIFF
--- a/_includes/api/en/4x/req-is.md
+++ b/_includes/api/en/4x/req-is.md
@@ -21,8 +21,26 @@ req.is('application/json')
 req.is('application/*')
 // => 'application/*'
 
+
+// Using arrays
+// When Content-Type is application/json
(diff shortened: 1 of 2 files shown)


bjohansebas (approved):
LGTM!

Issue #1941, opened 2025-06-02
Mention different path matching strategies


ties-s:
### Description

The documentation for the various HTTP method handlers (like [GET](https://expressjs.com/en/api.html#app.get.method)) link to [Path Examples](https://expressjs.com/en/api.html#path-examples) for more information about paths. The Path Examples are located underneath the `use` method and describe prefix matching. The HTTP method handlers however use full path matching, which is not mentioned.

### Expectations

Mention different behaviour either with the link, or in the examples

PR #2043, opened 2025-08-19
docs: clarify path matching behavior for route methods vs app.use() (#1941)

kushal4345:
This PR updates the documentation to address #1941 issue

Adds a note explaining that:

app.use() uses prefix-based path matching (e.g., /api will match /api/users).
HTTP method handlers (app.get(), app.post(), etc.) use exact path matching (e.g., app.get('/api') will only match /api).
Provides a clear code example to highlight the difference.
Ensures users are not confused when switching between app.use() and route methods.

This improves the clarity of the path examples section in the docs.


kushal4345:
The Netlify preview is working fine, but the Lighthouse check is failing with a 404. Since this PR only changes documentation text, the Lighthouse failure seems unrelated. Could you please re-run or skip this check?

PR #2045, opened 2025-08-19
docs: clarify path matching behavior for route methods vs app.use() (#1941)

kushal4345:
This PR updates the documentation to address #1941  issue

Adds a note explaining that:

app.use() uses prefix-based path matching (e.g., /api will match /api/users).
HTTP method handlers (app.get(), app.post(), etc.) use exact path matching (e.g., app.get('/api') will only match /api).
Provides a clear code example to highlight the difference.
Ensures users are not confused when switching between app.use() and route methods.

This improves the clarity of the path examples section in the docs.


kushal4345:
I wanted to kindly follow up on my PR submitted about two weeks ago that updates the documentation to clarify the difference between app.use() prefix-based path matching and route methods’ exact path matching (addresses issue #1941).

Please let me know if any further changes or adjustments are needed from my side. I’d be happy to make updates to help get it merged.

ShubhamOulkar (changes requested):
Similar examples are already present on following page 
image

PR #2047, opened 2025-08-23
 routing.md

KunalChoudhary03:
This will helps you a lot.

### DIFF
--- a/en/guide/routing.md
+++ b/en/guide/routing.md
@@ -22,6 +22,8 @@ In fact, the routing methods can have more than one callback function as argumen
 With multiple callback functions, it is important to provide `next` as an argument to the callback function and then call `next()` within the body of the function to hand off control
 to the next callback.  
 
+You do not need to install any additional npm package (such as `route`) for routing in Express.
+
 The following code is an example of a very basic route.
 
 ```js


KunalChoudhary03:
I see the Lighthouse check is failing with a 404 error. 
However, my change only affects the documentation text and not the URLs. 
Could you please confirm if this is something I need to fix, 
or if it’s unrelated to my changes?

bjohansebas:
Hey, thanks for your contribution. Don’t worry about Lighthouse, and I really don’t think we need to clarify that. Or is there some text that confuses you and makes you think we need to install another package?

KunalChoudhary03:
I agree with you on this point. However, beginners might find it a bit
confusing, and providing some clarification could really help them. For
experienced developers, it won’t be an issue at all, but adding that extra
note would make things easier for those who are just starting out.

bjohansebas (changes requested):
Yep, I understand the context for beginners, but I want to keep the guide focused. Unless there’s some text that’s really confusing and makes people think it needs to be handled separately, then that text should be improved. Otherwise, I won’t be able to merge this PR because it doesn’t align with the guide.

KunalChoudhary03:
I understood it, but I felt a bit confused, so I thought others might feel
the same. That’s why I just wanted to clarify, nothing else.

ShubhamOulkar (changes requested):
Docs clearly mentioned that it is built in middleware. I don't think it is confusing. You can learn it from https://expressjs.com/en/guide/using-middleware.html

KunalChoudhary03:
Ok, thanks.

ShubhamOulkar:
I am considering @KunalChoudhary03 agreed on our suggestions. SO I am closing this PR.

KunalChoudhary03:
Okay

PR #2049, opened 2025-08-28
add express-healthcheck to en/resources/middleware.md

amirparsadd:
express-healthcheck is a really cool library for healthchecks and it would be great if it was added to the docs for developers to see and use.


bjohansebas:
Hey, thanks, but unfortunately your package is new, it’s not an important package yet, and we don’t have a consensus on which third-party packages can be added and which can’t. I hope your package is successful, but I’m going to close this

PR #2055, opened 2025-09-09
feat: generate dynamic docs header links

ShubhamOulkar:
#### New design
|light|dark|
|-----|----|
|Screenshot 2025-09-14 084145|Screenshot 2025-09-14 084157|


css/style.css
```diff
@@ -1116,6 +1116,26 @@ h2 a {
...
 }
 
+/* h2 -> h3 header links */
+.heading-link::after {
+  content: '#';
+  color: var(--card-fg);
+}
+
+.heading-link {
+  margin-left: 0.3em;
```
jonchurch:
```suggestion
  padding-inline: 0.3em;
```

I prefer padding for this, as it expands the clickable area to include the spacing, vs margin which adds spacing that is not clickable

css/style.css
```diff
@@ -1116,6 +1116,26 @@ h2 a {
...
+  font-size: 1rem;
```
jonchurch:
```suggestion
```

Leaving the font-size to inherit from the contained header will let the anchor match the heading size, otherwise you get mismatches like this:

Image

css/style.css
```diff
@@ -1116,6 +1116,26 @@ h2 a {
   font-size: 16px;
 }
 
+/* h2 -> h3 header links */
+.heading-link::after {
+  content: '#';
+  color: var(--card-fg);
```
jonchurch:
```suggestion
```

If taking the nit https://github.com/expressjs/expressjs.com/pull/2055/files#r2370267450 we can drop color from here, or set color to inherit here. Both work, but styling the link is slightly more centralized.

css/style.css
```diff
@@ -1116,6 +1116,26 @@ h2 a {
...
+  margin-left: 0.3em;
+  font-size: 1rem;
+  text-decoration: none;
+  opacity: 0;
+  transition: opacity 0.2s ease-in-out;
+}
+
+h2:hover .heading-link,
+h3:hover .heading-link,
+.heading-link:is(:hover, :focus, :focus-visible) {
```
jonchurch, line 1141:
```suggestion
:target .heading-link,
.heading-link:is(:hover, :focus, :focus-visible) {
```

We could also show the anchor persistently when a user arrives via deep link into the content, or has clicked on a heading. Can sometimes help you orient after landing on a page.

This is purely stylistic feedback

css/style.css
```diff
@@ -1116,6 +1116,26 @@ h2 a {
...
+/* h2 -> h3 header links */
+.heading-link::after {
+  content: '#';
+  color: var(--card-fg);
+}
+
+.heading-link {
+  margin-left: 0.3em;
+  font-size: 1rem;
+  text-decoration: none;
```
jonchurch, line 1132:
```suggestion
  text-decoration: none;
  color: inherit;
```

Nit: We can move styling of the pseudo element (the `::after` with `#`) color up a level. so we style the `a` itself to inherit from the header, and the pseudo element will inherit from there.

(will bork anything where the header is not the one providing a color, but that's useful in its own way)

js/copycode.js
```diff
@@ -1,4 +1,5 @@
 const codeBlocks = document.querySelectorAll("pre:has(code)");
+const headers = document.querySelectorAll('h2[id],h3[id]');
```
jonchurch: This is broad enough that it is also selecting our sidebar which has an h3:

Image


We can scope it down to only select headers which appear inside the `<main>` tag
```suggestion
const headers = document.querySelectorAll('main :is(h2,h3)[id]');
```

js/copycode.js
```diff
@@ -13,6 +14,15 @@ codeBlocks.forEach((block) => {
   });
 });
 
+// add heading links
+for (const heading of headers) {
```
jonchurch, line 18:
```suggestion
for (const heading of headers) {
    if (heading.querySelector('.heading-link')) continue;
```
Nit: A little paranoid but this will guard against the script running twice somehow by accident or hot reloading or something, preventing us from adding the elements twice.

jonchurch (changes requested):
For some reason I thought we already had these, or had them in a PR, but I can't find any evidence for that.

Great idea. 

Do note though, this will add these anchor links to `h2, h3` on **all pages** of the site!

I didn't check all pages to see if it is messed up anywhere, but I caught some bugs on the docs page at least.

Is there a reason why this is living inside the `copycode.js` script file? it could move to the `app.js` file inside the jquery DOMContentLoaded callback at the top of the file

css/style.css
ShubhamOulkar: I think if we do font size inherit, padding will grow even bigger links especially on h2 headers.

css/style.css
ShubhamOulkar: yes I will do that

css/style.css
ShubhamOulkar: This is great, I have completely forgotten this. Thank you 💖

css/style.css
ShubhamOulkar: yes

js/copycode.js
ShubhamOulkar: I want thinking of implementing lazy attach links using  `IntersectionObserver` or `MutationObserver`. WDYT?

> Is there a reason why this is living inside the copycode.js script file?

Copy code and header link share same logic and needs asynchronous implementation.

css/style.css
ShubhamOulkar: This looks good 👍
image

ShubhamOulkar:
I have addressed all reviews. Please take a look.

> it could move to the `app.js` file inside the jquery

Yes, I will move this code to app.js. Just a reminder, jQuery is migrated to web APIs. So no more jquery.

> Do note though, this will add these anchor links to `h2, h3` on **all pages** of the site!

Yes, this will add anchor on all pages. Here are some issues that are affecting this PR https://github.com/expressjs/expressjs.com/pull/2055#discussion_r2370334622 and #2056. First is fixed here itself, #2056 needs to be fixed.

ShubhamOulkar:
bump @jonchurch

js/copycode.js
ShubhamOulkar: The DOM is reloaded completely on each hot reload, so duplicates won’t occur. Still adding this check as a defensive measure in case the script ever re-runs

js/copycode.js
bjohansebas, line 1: Why is this new logic here instead of in app.js?

ShubhamOulkar:
@bjohansebas, I think astro can do it in better way. You can close this PR if not necessary.

ShubhamOulkar:
astro can do it in better way.

PR #2061, opened 2025-09-15
Update best-practice-performance.md

ibandhiya:
update link on optimize app performance


jonchurch:
Thank you but no thank you. 

We don't update these with 3rd party recommendations anymore. I will leave it at that.

PR #2063, opened 2025-09-20
[DRAFT] chore: Maintainance

ShubhamOulkar:
See discussion https://github.com/expressjs/expressjs.com/discussions/2004

### Why this change?

- [X] Helps us manage assets more gracefully going forward.
- [X] Makes future redesigns or migrations to a new platform smoother.
- [X] Improves DX (Developer Experience) and reduces unnecessary clutter at the root level.

> Note:- While the diff size is large, a future migration would result in an even bigger diff and introduce more room for mistakes. Handling this restructuring now will minimize those risks.


jonchurch (changes requested):
This PR is big (59 files), alters files being moved, has diffs in the moved files, and the. There is no explanation for these changes.

The linked Jekyll tutorial does not recommend this restructuring, and there’s no explanation provided for why it’s being done.

jonchurch:
Before I would approve a PR like this, it would need to be limited to the changes required to achieve the goal. So no extraneous diffs.

I would also want a compelling reason for touching so many files in the first place. 

Otherwise it looks like rearranging the furniture for no reason, and Id be inclined to close the PR.

ShubhamOulkar:
@jonchurch Thanks for the review! This will really help us manage website assets more gracefully, and it’ll make things much smoother if we ever redesign the site or migrate to a new platform. It improves DX and avoids unnecessary root clutter. Kind of like furniture—you can only use it if it’s arranged properly. 🫠

ShubhamOulkar:
> we ever redesign the site or migrate to a new platform

> his PR is big (59 files), alters files being moved, has diffs in the moved files

This PR is quite big (59 files), and it touches both moved files and their diffs. But I’d also point out that if we ever do a migration, the diffs will end up being even bigger than this. I’d like to avoid that scenario, since larger diffs usually lead to more mistakes.

ShubhamOulkar:
@bjohansebas 
Will this is helpful while migrating to astro? Otherwise I will close this PR.

bjohansebas:
I think this will no longer be necessary, because technically the page is going to be built from scratch, so these kinds of changes won’t be needed anymore. Let’s focus now on the migration and on fixing the documentation

Thanks for the PR

PR #2075, opened 2025-10-04
fix: ensure footer is pushed to the bottom on large screens

### LINKED ISSUE #2079: Footer not anchored at bottom on large/wide screens
# The homepage layout on large and wide screens breaks:

I already submitted a 3 lines fix for it a few days ago but it awaits merging:  
https://github.com/expressjs/expressjs.com/pull/2075

The footer is not anchored at the bottom of the screen:
Image

Expected layout:
Image

sibaiper:
### Summary

Fixes footer layout issue on very wide screens where the footer does not stick to the bottom of the page.

### Problem

* On large screens, the homepage layout leaves the footer floating above the bottom of the viewport.
* This looks fine on normal screens, but on very wide screens the gap is obvious.

Screenshots:

* **Big screen (before):** original_big_screen

* **Small/normal screen (before):** original_small_screen

### Changes Made

* Set `html` and `body` to take up the full screen height.
* Adjusted footer styling so it always sticks to the bottom.

### Testing

* Verified locally on normal and wide screens.
* Layout works consistently across different viewport sizes.

---

Let me know if further adjustments are needed.

closes #2079

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -17,20 +17,8 @@ h1, h2, h3, h4, h5, h6,
   }
 }
 
-html {
-  /*
-    html & body need to take the whole
-    visible area in order for the footer
-    to be pushed all the way down or else
-    on really big screens or when the user
-    is zoomed out, the footer is not all
-    the way at the bottom of the screen
-  */
-  height: 100%;
-}
-
 body {
-  height: 100%;
+  height: 100vh;
   font: 400 14px/1.6 "Open Sans", sans-serif;
   background: var(--bg);
   margin: 0;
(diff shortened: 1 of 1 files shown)


css/style.css
```diff
@@ -17,7 +17,20 @@ h1, h2, h3, h4, h5, h6,
...
+    to be pushed all the way down or else
+    on really big screens or when the user
+    is zoomed out, the footer is not all
+    the way at the bottom of the screen
+  */
+  height: 100%;
+}
+
 body {
+  height: 100%;
```
```suggestion
  height: 100vh;
```
I think we should use `vh` instead of height inheritance.

ShubhamOulkar (approved):
LGTM

PR #2080, opened 2025-10-10
Remove reference to unimplemented feature

calebegg:
path-to-regexp does not support this syntax at all. There was a PR to implement it, but it seems pretty dead. https://github.com/pillarjs/path-to-regexp/pull/362

### DIFF
--- a/en/guide/routing.md
+++ b/en/guide/routing.md
@@ -216,31 +216,6 @@ In express 5, Regexp characters are not supported in route paths, for more infor
 
 {% include admonitions/caution.html content=warning-regexp %}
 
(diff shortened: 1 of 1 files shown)


bjohansebas (changes requested):
Thank you for your contribution, but unfortunately we maintain the documentation for both Express v4 and v5, so we can't remove this information that easily.

It may be confusing, and perhaps we can improve the way we present this information.

calebegg:
I see. Maintaining one source of documentation for two dramatically different versions of an application seems challenging. I would suggest that y'all take a broader approach to that issue than can reasonably be addressed in one PR.

PR #2092, opened 2025-11-01
docs: document req.params changes in 5.x

### LINKED ISSUE #2090: 5.x - undocumented wildcard behaviour change in `req.params`
### Description

Anthony Moulen [pointed out on Slack](https://openjs-foundation.slack.com/archives/C02QB1731FH/p1761775704128149) a change in `req.params` behaviour when using wildcards. Since [`path-to-regexp@7`](https://github.com/pillarjs/path-to-regexp/releases/tag/v7.0.0) wildcards [split the matched path on `/`](https://github.com/pillarjs/path-to-regexp/pull/269#issuecomment-2227138973).

| route 4.x | route 5.x | request path | `req.params` 4.x | `req.params` 5.x |
| --------- | --------- | ------------ | ---------------- | ---------------- |
| `'/:name'` | `'/:name'` | `/a%2fb` | `{ name: 'a/b' }` | `{ name: 'a/b' }` |
| `'/*'` | `'/*name'` | `/` | `{ '0': '' }` | _not matched_ |
| `'/*'` | `'/{*name}'` | `/` | `{ '0': '' }` | `{}` |
| `'/*'` | `'/*name'` | `/a%2fb/c` | `{ '0': 'a/b/c' }` | `{ name: [ 'a/b', 'c' ] }` |
| `'/*'` | `'/*name'` | `/a%2fb` | `{ '0': 'a/b' }` | `{ name: [ 'a/b' ] }` |

The migration guide describes the changes in path matching - that wildcard parameters now must be named and that `{}` is required to make the path segment optional. What it does not described is that:

1. Wildcard parameters in `req.params` are always represented as arrays of decoded segments, not a single string.
2. (less important) Optional wildcard segment that is not matched does not exist in `req.params`, while in 4.x it was an empty string.
3. (not important) `req.params` object in 5.x has null prototype.

### Expectations

Migration guide should mention different `req.params` behaviour with regard to wildcard parameters.

This probably can be combined with `req.params` documentation improvements requested in #1891.

krzysdz:
Fixes #2090


krzysdz:
Just one thing that I remembered - `req.params` has null prototype, **unless the path is a regular expression**. With regexes `req.params` is a normal object (see [`regexpMatcher`](https://github.com/pillarjs/router/blob/cb2a5fcdf31b9cc92296183ceeff8126b5c57453/lib/layer.js#L62-L84) in `router`)

en/guide/migrating-5.md
```diff
@@ -507,14 +508,46 @@ const server = app.listen(8080, '0.0.0.0', (error) => {
...
+
+// v5
+app.get('/:file{.:ext}', (req, res) => {
+  // GET /image
+  console.dir(req.params)
+  // => [Object: null prototype] { file: 'image' }
+})
+```
+
+The object has null prototype.
```
bjohansebas: The `req.params` object now has a **null prototype**. Additionally, there are two important behavioral changes:

**Wildcard parameters are now arrays:**

Wildcards (e.g., `/*splat`) capture path segments as an array instead of a single string.

```js
app.get('/*splat', (req, res) => {
  // GET /foo/bar
  console.dir(req.params)
  // => [Object: null prototype] { splat: [ 'foo', 'bar' ] }
})
```

**Unmatched parameters are omitted:**

In Express 4, unmatched wildcards were empty strings (`''`) and optional `:` parameters (using `?`) had a key with value `undefined`. In Express 5, unmatched parameters are completely omitted from `req.params`.

```js
// v4: unmatched wildcard is empty string
app.get('/*', (req, res) => {
  // GET /
  console.dir(req.params)
  // => { '0': '' }
})

// v4: unmatched optional param is undefined
app.get('/:file.:ext?', (req, res) => {
  // GET /image
  console.dir(req.params)
  // => { file: 'image', ext: undefined }
})

// v5: unmatched optional param is omitted
app.get('/:file{.:ext}', (req, res) => {
  // GET /image
  console.dir(req.params)
  // => [Object: null prototype] { file: 'image' }
})
```

bjohansebas (commented):
Thanks @krzysdz!


What do you think about changing it to this suggestion, which includes more information and code examples showing what changed, to make the change easier to understand?

krzysdz:
Your suggestion is definitely clearer. I'll admit that I'm not great at writing documentation.

There is just the question about prototype that may or may not be `null`... I did not notice it initially and forgot to update the PR when I tested some more things and realised that regular expressions have `req.params` with a regular `Object` prototype. To be fair this inconsistency is probably something that should be changed in `router`.

bjohansebas:
> There is just the question about prototype that may or may not be null... I did not notice it initially and forgot to update the PR when I tested some more things and realised that regular expressions have req.params with a regular Object prototype. To be fair this inconsistency is probably something that should be changed in router. 

yep, when it's not regexp then the object is [created here](https://github.com/pillarjs/path-to-regexp/blob/05a5a973702a863fb69415294503d13cb9d18b20/src/index.ts#L451), whereas [here](https://github.com/pillarjs/router/blob/cb2a5fcdf31b9cc92296183ceeff8126b5c57453/lib/layer.js#L68) is where the object is created, probably we should change that in the router so both ways are the same

bjohansebas (approved):
LGTM!

PR #2101, opened 2025-11-01
docs: clarify req.params type and usage with regular expressions

### LINKED ISSUE #2100: `req.params` docs claim that this property can be an array
### Description

When 4.x documentation was created the docs were copied from 3.x. One of the breaking changes was [making `req.params` an object](https://github.com/expressjs/express/blob/4453d83ccaed5c80f0c10a23d01216eff612ee56/History.md?plain=1#L1707). This was partially corrected in #134, but the documentation still claims that if a regular expression is used, `req.params` is an array:

https://github.com/expressjs/expressjs.com/blob/738f860585a6f34511ba2b2da40c5bd76ef49579/_includes/api/en/4x/req-params.md?plain=1#L11

This isn't a huge problem, since values captured by unnamed capture groups have keys `'0'`, `'1'`, `'2'` etc. and `req.params[n]` access (where `n` is an integer), due to how JS works, will work. **EDIT:** In practice V8 [optimizes](https://v8.dev/blog/fast-properties#named-properties-vs.-elements) array-indexed property access, so I won't mention this in my PR.

This error is present both in 4.x and 5.x docs (and [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/c030728c3e25d15bcb3fbf48edf2df784b9f08db/types/express-serve-static-core/index.d.ts#L383-L388) - looks like this caused some problems, because they trusted the docs https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37502#issuecomment-522155668).

### Expectations

API docs (4.x and 5.x) should clearly explain that `req.params` always is an object and how keys are assigned.

krzysdz:
Fixes #2100

### DIFF
--- a/_includes/api/en/4x/req-params.md
+++ b/_includes/api/en/4x/req-params.md
@@ -8,13 +8,15 @@ console.dir(req.params.name)
 // => 'tj'
 ```
 
-When you use a regular expression for the route definition, capture groups are provided in the array using `req.params[n]`, where `n` is the n<sup>th</sup> capture group. This rule is applied to unnamed wild card matches with string routes such as `/file/*`:
+When you use a regular expression for the route definition, capture groups are provided as integer keys using `req.params[n]`, where `n` is the n<sup>th</sup> capture group. This rule is applied to unnamed wild card matches with string routes such as `/file/*`:
 
 ```js
 // GET /file/javascripts/jquery.js
 console.dir(req.params[0])
 // => 'javascripts/jquery.js'
 ```
+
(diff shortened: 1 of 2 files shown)


bjohansebas (approved):
Good catch, for referencese https://github.com/expressjs/express/pull/1835/ is where the change was made for express v4/v5

ref https://github.com/pillarjs/router/blob/cb2a5fcdf31b9cc92296183ceeff8126b5c57453/index.js#L68

PR #2102, opened 2025-11-06
Add comment for app.get() method usage

Melkzedk:
Added comment to explain app.get() method.

### DIFF
--- a/en/starter/hello-world.md
+++ b/en/starter/hello-world.md
@@ -17,6 +17,7 @@ const express = require('express')
 const app = express()
 const port = 3000
 
+//app.get() is a method used to define a route that responds to HTTP GET requests.
 app.get('/', (req, res) => {
   res.send('Hello World!')
 })


efekrskl:
Thanks for the PR!

IMO it could be helpful to have some comments in the hello world example, but I'm not sure if this particular comment is helpful. Maybe you could check similar frameworks (e.g fastify) to see how they handle their hello world example

ShubhamOulkar (changes requested):
@Melkzedk, thank you for your efforts. But these changes can not improve documentation.

en/starter/hello-world.md
```diff
@@ -17,6 +17,7 @@ const express = require('express')
 const app = express()
 const port = 3000
 
+//app.get() is a method used to define a route that responds to HTTP GET requests.
```
bjohansebas, line 20: What do you think if we add something like this:

```js
const app = express()
const port = 3000

// Define a route handler for GET requests to the root URL ('/')
// When someone visits http://localhost:3000/, this function runs
app.get('/', (req, res) => {
  res.send('Hello World!')  // Send back the response
})

// Start the server and listen for incoming requests on the specified port
app.listen(port, () => {
  console.log(`Example app listening on port ${port}`)
})
```
cc: @efekrskl @ShubhamOulkar @expressjs/docs-wg

efekrskl: Looking more in depth, [the hello world page](https://expressjs.com/en/starter/hello-world.html) contains some extra details outside the example, which already cover these details. I would say we can close the PR.

bjohansebas: I agree, maybe it made sense to add some commentary directly in the code example, but I’m inclined to close the PR.

PR #2103, opened 2025-11-06
Add Nodemon usage instructions for live server restarting

Melkzedk:
Added instructions for using Nodemon to restart the Node.js server automatically during development.

### DIFF
--- a/en/starter/generator.md
+++ b/en/starter/generator.md
@@ -74,6 +74,20 @@ Then install dependencies:
 ```bash
 $ cd myapp
 $ npm install
+
+## Using Nodemon for Live Server Restarting
+Nodemon automatically restarts your Node.js server whenever you make changes to your files — perfect for development.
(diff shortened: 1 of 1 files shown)


efekrskl:
Thanks for the PR!

Unfortunately, I see several issues with the PR and I don't think it'd be a good contribution. Just to name a few:

1) I don't understand how is this relevant to express-generator docs (at best, it'd be a different section in docs)
2) Now that `node --watch` is stable, this might be favored (although, I might be wrong, not sure if they map 1:1)
3) Your example doesn't even work, as the main entry file is not `server.js` with express-generator (as shown in the docs)

ShubhamOulkar (changes requested):
agreed with @efekrskl

bjohansebas (changes requested):
i agree with the others, i'm going to close

PR #2126, opened 2025-12-02
docs: remove deprecated res.redirect('back') from Express 3.x and 4.x API docs

### LINKED ISSUE #1834: Remove reference to `res.redirect('back')`.

vini4721:
Fixes #1834

## Changes
- Removed deprecated `res.redirect('back')` example from Express 3.x API documentation
- Removed deprecated `res.redirect('back')` example and deprecation notice from Express 4.x API documentation

## Reason
The `res.redirect('back')` method was deprecated in Express 4.x and completely removed in Express 5.x. These examples should no longer appear in the API documentation as they may confuse users.

## Related
- The migration guides correctly explain that `'back'` is no longer supported and recommend using `req.get('Referrer') || '/'` instead
- Express 5.x documentation already does not include this deprecated example

### DIFF
--- a/_includes/api/en/3x/res-redirect.md
+++ b/_includes/api/en/3x/res-redirect.md
@@ -43,3 +43,9 @@ would land you at `http//example.com/admin/post`:
 res.redirect('..')
 ```
 
+The final special-case is a `back` redirect, redirecting back to
+the Referer (or Referrer), defaulting to `/` when missing.
+
+```js
+res.redirect('back')
+```
--- a/_includes/api/en/3x/res-redirect.md
+++ b/_includes/api/en/3x/res-redirect.md
@@ -43,9 +43,3 @@ would land you at `http//example.com/admin/post`:
 res.redirect('..')
 ```
 
-The final special-case is a `back` redirect, redirecting back to
-the Referer (or Referrer), defaulting to `/` when missing.
-
-```js
-res.redirect('back')
-```
--- a/_includes/api/en/4x/res-redirect.md
+++ b/_includes/api/en/4x/res-redirect.md
@@ -45,14 +45,5 @@ Path-relative redirects are also possible. If you were on
 res.redirect('..')
 ```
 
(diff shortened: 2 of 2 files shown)


krzysdz:
Deprecated API should be marked as such (and it is), not removed from documentation if it still exists. It was removed from Express 5 docs, because it does not exist in Express 5. If you see somewhere in old code `res.redirect('back')`, you may want to check in the documentation what does it do. If the documentation is removed, it will be very confusing for users.

Your PR not only removes documentation for a feature that still exists (talking about Express 4), but also the recommendation how to replace it.

vini4721:
Thank you for the feedback @krzysdz! I understand the issue now.

I need clarification: Issue #1834 mentions that `res.redirect('back')` appears in the Express 5 API documentation. However, when I searched the repository, I found it in the 3.x and 4.x files:
- `_includes/api/en/3x/res-redirect.md`
- `_includes/api/en/4x/res-redirect.md`

I did NOT find a separate 5.x API documentation file for `res.redirect()`.

Could you please point me to which specific file contains the Express 5 API reference that needs to be updated? I want to fix this correctly by only removing it from Express 5 docs while keeping it in 3.x/4.x docs where it's properly marked as deprecated.

krzysdz:
The 5.x documentation for `res.redirect()` is under [`_includes/api/en/5x/res-redirect.md`](https://github.com/expressjs/expressjs.com/blob/gh-pages/_includes/api/en/5x/res-redirect.md) and the issue #1834 was fixed by PR #1835 almost 9 months ago. If you're looking for problems to fix, closed issues aren't the best place to start.

bjohansebas (changes requested):
Hi, thanks for your contribution. This PR doesn’t seem to have any changes, unless GitHub is broken and not showing them to me.

If it’s as the title says, we don’t need to remove those references in the other APIs, as @krzysdz already mentioned.

PR #2129, opened 2025-12-10
fix: correct broken German middleware link

surajbhatt27:
Fixed broken link in German middleware documentation by removing duplicate /de segment.


efekrskl:
[The link looks fine to me](https://expressjs.com/de/guide/using-middleware.html), are we sure it's broken?

PR #2144, opened 2025-12-29
docs: add example for req.accepts usage

hamzafrd:
### What this PR does
Adds a practical example demonstrating how `req.accepts()` can be used
to respond with different content types and handle unsupported formats.

### Why
The existing documentation explains the method but lacks a concrete
example, which can be confusing for new users.

### Checklist
- [v] Documentation only
- [v] No behavior changes
- [v] Tests pass


bjohansebas:
I think we'll keep it simple for now. We can evaluate in the future whether to add it or not

PR #2154, opened 2026-01-13
docs: add manual setup guide for Handlebarsdocs

### LINKED ISSUE #2128: Templating engine pages could be improved
### Description

> The [Express application generator](https://expressjs.com/en/starter/generator.html) uses [Pug](https://pugjs.org/api/getting-started.html) as its default, but it also supports [Handlebars](https://www.npmjs.com/package/handlebars), and [EJS](https://www.npmjs.com/package/ejs), among others.
> To render template files, set the following [application setting properties](https://expressjs.com/en/4x/api.html#app.set), in the default app.js created by the generator

For anybody that didn't use the generator, this really isn't a very helpful page. I followed "Installing" which is the first item in the "Getting started" section and now the docs are assuming I've done something else. I am getting the error `Error: Module "handlebars" does not provide a view engine.` Googling is pointing me towards  the express-handlebars package but that seems at odds with what is written here. The docs page here says "it also supports Handlebars" and links to the Handlebars NPM package, not to express-handlebars.

### Expectations

Don't assume everyone has used the generator.

VikramSinghRakwal06:
### Description
This PR addresses a documentation gap in the "Using template engines with Express" guide. Currently, the documentation assumes users are using the `express-generator`. Developers building applications manually often encounter `Error: Module "handlebars" does not provide a view engine` because the core `handlebars` package is not Express-compliant without a wrapper.

### Changes
- Added a **Manual Setup (Without Generator)** section.
- Included a clear code example for manually configuring the `express-handlebars` engine using `app.engine()`.
- Clarified the distinction between the core library and the necessary Express adapter.

### Impact
Improves the developer experience (DX) for users building minimal or custom Express configurations, reducing common setup errors.

### Related Issues
Fixes #2128

### DIFF
--- a/en/guide/using-template-engines.md
+++ b/en/guide/using-template-engines.md
@@ -20,19 +20,20 @@ To render template files, set the following [application setting properties](/{{
 * `views`, the directory where the template files are located. Eg: `app.set('views', './views')`.
 This defaults to the `views` directory in the application root directory.
 * `view engine`, the template engine to use. For example, to use the Pug template engine: `app.set('view engine', 'pug')`.
+
 ### Manual Setup (Without Generator)
-+
-+ If you are not using the Express generator, you must manually configure the engine to ensure Express can interface with it properly. While engines like Pug work out of the box, others (like Handlebars) require a specific wrapper package.
-+
-+ For example, to use Handlebars manually:
-+ 1. Install the `express-handlebars` package: `$ npm install express-handlebars`
-+ 2. Configure it in your `app.js`:
-+
-+ ```js
-+ const { engine } = require('express-handlebars');
-+ app.engine('handlebars', engine());
-+ app.set('view engine', 'handlebars');
-+ ```
+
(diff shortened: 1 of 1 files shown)


ShubhamOulkar (changes requested):
Hi @VikramSingh-vartagram, new added section breaks original flow of the document. Also, I think these changes are irrelevant because line no.22 already shows how to set view engine.

VikramSinghRakwal06:
Thank you for the feedback! While line 22 explains how to toggle the view
engine setting, it assumes the engine is already Express-compliant out of
the box (like Pug).

​Many users building apps manually (without the generator) run into issues
with Handlebars because it requires an explicit app.engine() configuration
to work with Express. My goal was to provide the missing setup logic for
these non-default engines to prevent common 'Module does not provide a view
engine' errors.

​If you feel the separate section breaks the flow, would you prefer I
integrate this example directly into the 'Manual Setup' notes or as a
sub-bullet under the view engine property?

PR #2159, opened 2026-01-19
ci: Fix error in the workflow.

bjohansebas:
There was a small error in the workflow.


jonchurch:
Was there a failure somewhere?

Why does making this multiline change something?

This change shouldn't have any impact

bjohansebas:
see https://github.com/expressjs/expressjs.com/actions/runs/21151216689/workflow

jonchurch (approved):
Ah, yaml parsing issue. Probably the `:`

Multiline does fix parsing issues like that 👍

PR #2162, opened 2026-01-20
ci(deploy): ensure .well-known dir is uploaded to gh-pages

jonchurch:
fixes issue in https://github.com/expressjs/expressjs.com/pull/1974#issuecomment-3770111739 where the `.well-known/security.txt` was not being published to GH pages

That upload-pages-artifact action wraps the upload-artifact action which has this config option: https://github.com/actions/upload-artifact#usage

since `_site` is built fresh each time, we shouldnt need to worry about any credentials getting uploaded or anything

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -33,9 +33,11 @@ jobs:
         run: bundle exec jekyll build --baseurl "${{ steps.pages.outputs.base_path }}"
         env:
           JEKYLL_ENV: production
-      - name: Upload artifact
-        # Automatically uploads an artifact from the './_site' directory by default
+      - name: Upload gh-pages artifact
         uses: actions/upload-pages-artifact@v4
+        with:
+          path: _site
+          include-hidden-files: true # required for /.well-known/security.txt
 
   deploy:
     environment:


bjohansebas (approved):
LGTM, thanks!

bjohansebas:
Hmm, that doesn’t work either https://expressjs.com/.well-known/security.txt

bjohansebas:
see https://github.com/actions/upload-pages-artifact/issues/129 and github.com/withastro/action/pull/88/changes We need to change more things in our infrastructure to do the deployment.

jonchurch:
Dang. Theres a PR open against the action to add an include, which would help us here

https://github.com/actions/upload-pages-artifact/pull/134

In the meantime we most likely need to do something like the astro workflow you linked to

.github/workflows/deploy.yml
```diff
@@ -33,9 +33,11 @@ jobs:
         run: bundle exec jekyll build --baseurl "${{ steps.pages.outputs.base_path }}"
         env:
           JEKYLL_ENV: production
-      - name: Upload artifact
-        # Automatically uploads an artifact from the './_site' directory by default
+      - name: Upload gh-pages artifact
         uses: actions/upload-pages-artifact@v4
+        with:
+          path: _site
+          include-hidden-files: true # required for /.well-known/security.txt
```
ShubhamOulkar, line 40: We should remove this. This parameter is not present on this action. https://github.com/actions/upload-pages-artifact

ShubhamOulkar (commented):
PR description linked to wrong action `upload-artifacts'.

ShubhamOulkar:
I think we have two options to fix this issue: use the "Deploy from branch" source option from Settings, or downgrade to v3.

My opinion is to downgrade to v3. Since .well-known/security.txt contains only public security contact, policy, and date information, and GitHub's official workflow still uses v3, there's no problem with the downgrade.

PR #2166, opened 2026-01-21
ci(deploy): include .well-known folder in GitHub Pages deployment

ShubhamOulkar:
We can not access file at https://expressjs.com/.well-known/security.txt because .well-known directory is ignored by action `upload-pages-artifact@v4`.

- [X] Remove unsupported parameters
- [ ] downgrade to v3 / use deploy from branch from settings.

PR #2162 #1974


.github/workflows/deploy.yml
```diff
@@ -35,9 +35,6 @@ jobs:
           JEKYLL_ENV: production
       - name: Upload gh-pages artifact
         uses: actions/upload-pages-artifact@v4
```
ShubhamOulkar, line 37:
```suggestion
        uses: actions/upload-pages-artifact@v3
```

My opinion is to downgrade to v3. Since .well-known/security.txt contains only public security contact, policy, and date information, and GitHub's official workflow still uses v3, there's no problem with the downgrade.

bjohansebas: I’m a bit concerned that during the time it takes us to migrate to Astro, that version might get deprecated and we end up back at the initial state again. After the migration, we wouldn’t have to worry about that.

ShubhamOulkar: hmm, both of this will take time to happen⌛. I think we should use this cheap fix for right now.

ShubhamOulkar:
@bjohansebas, should we close this PR?

bjohansebas:
Let’s try testing by going back to the previous version.

jonchurch (changes requested):
I opened #2173 to address the invalid config and allow uploading `.well-known` without pinning back any actions.

ShubhamOulkar:
Not planned

PR #2170, opened 2026-01-29
fix: point netlify to astro dir

jonchurch:
For pr previews and branch deploys

The netlify build is currently failing, it cant parse the dir to deploy from.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,4 +1,4 @@
 [build]
-  base = astro
+  base = "astro"
   command = "npm run build"
-  publish = "astro/dist"
+  publish = "dist"


netlify.toml
```diff
@@ -1,4 +1,4 @@
 [build]
-  base = astro
+  base = "astro"
   command = "npm run build"
   publish = "astro/dist"
```
jonchurch: I think this needs to be relative to the base, so would just be "dist"

bjohansebas: try

jonchurch:
was at the dentist when I opened this, pushed the fix from the comment, deploys are working now

PR #2171, opened 2026-01-30
ci: remove Lighthouse audit workflow

bjohansebas:
I’m removing the CI for Lighthouse since we’re now going to use a plugin that Netlify provides (I’ve already talked to Rand and Jon about this). I’ve installed it already, as I now have access to Netlify, this is just to remove the workflow.

If we ever change hosting for previews, we can use this workflow again. This is just to reduce the number of CI workflows we have right now.


bjohansebas:
Fast-tracking is failing on the redesign branch, so this is to prevent it from showing as red.

ShubhamOulkar:
This is what I wanted before!

PR #2173, opened 2026-01-30
deploy: create tar archive for gh-pages manually

jonchurch:
[upload-pages-artifact](https://github.com/actions/upload-pages-artifact) excludes hidden files, so .well-known isnt being
published. see https://github.com/actions/upload-pages-artifact/issues/129

This PR drops that action to tar the _site dir ourselves and hands it
off to upload-artifact for publishing.

We probably don't even need to use that action if it ever has the config
changed (creating the tar is a one liner), but I'll open a PR to them with a `include-hidden-files` option
Edit: [PR to upstream](https://github.com/actions/upload-pages-artifact/pull/137) w/ the option

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -20,24 +20,31 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+
       - name: Setup Ruby
         uses: ruby/setup-ruby@8d27f39a5e7ad39aebbcbd1324f7af020229645c $ v1.287.0
         with:
-          bundler-cache: true # runs 'bundle install' and caches installed gems automatically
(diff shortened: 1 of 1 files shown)


bjohansebas (commented):
Could you create a POC to verify that this solution actually works? If not, I’ll do it over the weekend

jonchurch:
I didnt do a PoC or actually test this. 

What I did was copy the tar logic from the upload-pages-artifact action verbatim, but removed the hidden file exclusion. Ive definitely merged untested action changes that didnt work before hehe. Im confident in this one, but havent tested it. Will report back if I do test it

bjohansebas:
My plate is too full to work on a PoC right now. You’re the captain, so I trust you.

bjohansebas:
The new site won’t need this since we’re going to use Astro’s action, which already solves that problem. And since we’re close to launching the new site, I’m going to close this.

PR #2167, opened 2026-01-28
Redesign: Technical Setup and Project Base

g-francesca:
> [!NOTE]
> There’s no preview link, @bjohansebas working on fixing that

This PR introduces a complete Astro-based rebuild of the Express.js documentation site, establishing the foundational architecture, design system, and UI component library. The work includes:

- Astro 5 project setup
- Design system setup with tokens and primitives
- Content layer configuration with routing system
- i18n infrastructure for multi-language support
- Living documentation of all design foundations
---

### Astro 5 Configuration
- Initialized Astro 5 project with latest features including Content Layer API
- Configured Astro with MDX support
- Set up PostCSS with custom media queries, autoprefixer, and cssnano
- Established Node.js (≥24.13.0) and npm (≥11.0.0) requirements

### Development Tooling
- **ESLint:** Configured with Astro, TypeScript, JSX a11y, and Prettier integration
- **Prettier:** Set up with Astro plugin for consistent code formatting
- **VSCode:** Added workspace settings, recommended extensions, and launch configurations
- **Git:** Updated `.gitignore` and `.editorconfig` for better project hygiene
- **Package Manager:** Enforced npm usage with preinstall checks
---

### Design Tokens
A complete token system following semantic naming conventions, organized in modular CSS files:

**Color System**
- Primitive color scales: gray (50-950), sky, green, amber, red, white/black
- Semantic tokens for consistent usage:
  - Background: `--color-bg-{primary|secondary|tertiary|inverse|mute|success|warning|error}`
  - Text: `--color-text-{primary|secondary|tertiary|inverse|mute|success|warning|error}`
  - Border: `--color-border-{primary|secondary|tertiary|inverse|mute|success|warning|error}`
  - Icons: `--color-icon-{primary|secondary|tertiary|inverse}`
- Theme-aware colors with CSS custom properties supporting light/dark modes

**Typography**
- Type scale: xs (12px) → 6xl (54px)
- Font families: Inter (system), JetBrains Mono (code)
- Font weights: light (300) → bold (700)
- Line heights optimized for readability
- Responsive font sizing with fluid typography

**Spacing** 
- Consistent scale from 0.5 (2px) to 16 (64px)
- Based on 4px grid system
- Tokens: `--space-{0.5|1|2|3|4|6|8|12|16}`

**Breakpoints** 
- Mobile-first approach with custom media queries
- xs: < 768px (mobile)
- md: 768px - 1439px (tablet)
- lg: ≥ 1440px (desktop)
[...]

PR #2180, opened 2026-02-05
feat: mobile header and sidebar navigation

g-francesca:
Preview: https://deploy-preview-2180--expressjscom-preview.netlify.app/

This PR introduces a comprehensive sidebar navigation system for the Express.js documentation with multi-level navigation support, version-aware content filtering, and accessibility features.

## Overview

The Sidebar component provides a hierarchical navigation system that supports:

- **Multi-level navigation** (up to 5 levels deep)
- **Version-aware content filtering** with automatic item omission
- **Keyboard navigation** and focus management
- **Automatic active state detection** based on current page

## Menu Configuration

Menus are configured using TypeScript files in [`astro/src/config/`](astro/src/config/):

### Menu Structure

```typescript
export const menu: Menu = {
  sections: [
    {
      title?: string;              // Optional section title
      basePath?: string;           // Base path prepended to all hrefs
      versioned?: boolean;         // Whether URLs include version prefix
      omitFrom?: VersionPrefix[];  // Hide entire section from specific versions
      items: MenuItem[];           // Array of menu items
    }
  ],
  items?: MenuItem[];              // Items outside sections
};
```

### Menu Items

Menu items can be either **links** or **submenu triggers**:

```typescript
// Link item
{
  label: 'Overview',
  ariaLabel?: 'Overview page',   // Optional accessible label
  icon?: 'files',                 // Optional icon (for root level)
  href: '/docs/overview',
  omitFrom?: ['v3']              // Hide from specific versions
}

// Submenu item
{
  label: 'API Reference',
  icon?: 'code',
  submenu: {
    basePath: '/api-reference',
    versioned: true,
    sections: [...]
  },
  omitFrom?: ['v3']
}
```

### Sections

Related items can be grouped under titled sections:

```typescript
sections: [
  {
    title: 'Application',
    items: [
      { href: '/application/overview', label: 'Overview' },
      { href: '/application/app-all', label: 'app.all' },
      { href: '/application/app-delete', label: 'app.delete' },
    ],
  },
  {
    title: 'Request',
    items: [
      { href: '/request/overview', label: 'Overview' },
      { href: '/request/req-accepts', label: 'req.accepts' },
    ],
  },
];
```

### Nested Submenus (Multi-level Navigation)

Create up to 5 levels of navigation by nesting submenus:

```typescript
// Level 0: Root menu
{
  label: 'Docs',
  submenu: {
    // Level 1: First submenu
    sections: [
      {
        title: 'Getting Started',
        items: [
          { href: '/installation', label: 'Installation' },
          {
            label: 'Advanced',
            submenu: {
              // Level 2: Second submenu
              items: [
                { href: '/advanced/middleware', label: 'Middleware' }
              ]
            }
          }
        ]
      }
    ]
  }
}
```

### Adding/Omitting Items by Version

Use the `omitFrom` property to hide items from specific versions:

```typescript
// Item visible in all versions
{ href: '/application/overview', label: 'Overview' }

// Item hidden from v3 only
{
  href: '/application/app-delete',
  label: 'app.delete',
  omitFrom: ['v3']
}

// Item only visible in v3
{
  href: '/request/req-accepted',
  label: 'req.accepted',
  omitFrom: ['v5', 'v4']
}
```

### Omitting Entire Sections

Hide entire sections from specific versions:

```typescript
{
  title: 'New Features',
  omitFrom: ['v3', 'v4'],  // Only show in v5
  items: [
    { href: '/new-feature-1', label: 'Feature 1' },
    { href: '/new-feature-2', label: 'Feature 2' },
  ]
}
```

### How Version Filtering Works

1. **Detection**: The sidebar automatically detects the current version from the URL path (e.g., `/en/docs/v5/...`)
2. **Filtering**: Items and sections with `omitFrom` containing the current version are excluded
3. **Rendering**: Only applicable items are rendered in the DOM
4. **Version Switching**: When the user switches versions:
   - If the current page exists in the new version → navigates to the same page
   - If the current page is omitted in the new version → navigates to the first available page

## Version Switcher Integration

The sidebar includes an integrated version switcher that appears when navigating into versioned submenus:

```typescript
versions = [
  { id: 'v5', label: 'v5.x', isDefault: true },
  { id: 'v4', label: 'v4.x' },
  { id: 'v3', label: 'v3.x (deprecated)' },
];
```

**Key features:**

- Appears only in versioned navigation panels
- Automatically hides when navigating back to root level
- Handles graceful fallback when switching to versions that omit the current page
- Updates URL structure from `/en/docs/v4/page` to `/en/docs/v5/page`

### Keyboard Controls

- **`Escape`**: Navigate back one level (or close sidebar if at root)
- **`Tab`/`Shift+Tab`**: Navigate between focusable elements
- **Arrow keys**: Navigate between menu items
- **Focus trap**: Focus stays within sidebar when open

### ARIA Implementation

- `role="dialog"` and `aria-modal="true"` for sidebar
- `aria-expanded` on submenu triggers
- `aria-hidden` for inactive navigation panels
- `aria-label` for navigation landmarks
- Screen reader announcements for navigation changes

## Active State Detection

The sidebar automatically detects and highlights the current page:

1. **Path Matching**: Compares current URL with menu item hrefs
2. **Deep Navigation**: Opens nested submenus containing the current page
3. **Initial State**: Restores proper navigation level on page load
4. **Visual Indicator**: Adds `sidebar-nav-item--active` class to current page link

### Component Structure

```
Sidebar.astro                      # Container component
├── SidebarMenu.astro             # Menu renderer (recursive for levels)
│   └── SidebarItemsList.astro    # Items list renderer
│       └── SidebarNavItem.astro  # Individual item renderer
├── SidebarController.ts          # Navigation state management
├── SidebarVersionManager.ts      # Version switching logic
├── SidebarFocusTrap.ts          # Focus management
└── utils.ts                      # Helper functions
```

### Key Logic

- **`collectAllSubmenus()`**: Recursively collects all nested submenus during build
- **`shouldOmitItem()`/`shouldOmitSection()`**: Version-based filtering
- **`resolveHref()`**: Constructs full URLs with language, version, and basePath
- **`submenuContainsCurrentPath()`**: Checks if a submenu contains the active page
- **`calculateInitialActiveLevel()`**: Determines which navigation level to show on load

## Configuration Files

Create new menu configurations in [`astro/src/config/`](astro/src/config/):

- [`main-menu.ts`](astro/src/config/main-menu.ts) - Root level navigation
- [`docs-menu.ts`](astro/src/config/docs-menu.ts) - Documentation submenu
- [`api-reference-menu.ts`](astro/src/config/api-reference-menu.ts) - API reference submenu
- [`middleware-menu.ts`](astro/src/config/middleware-menu.ts) - Middleware submenu
- [`types.ts`](astro/src/config/types.ts) - TypeScript type definitions

## Example: Adding a New Menu Item

```typescript
// In astro/src/config/api-reference-menu.ts

{
  title: 'Application',
  items: [
    { href: `/application/overview`, label: 'Overview' },
    { href: `/application/app-all`, label: 'app.all' },
    // Add new item here
    {
      href: `/application/app-engine`,
      label: 'app.engine',
      omitFrom: ['v3']  // Optional: hide from v3
    },
  ]
}
```

## Notes

Please note that:

- This PR focuses exclusively on the mobile sidebar navigation.
- Language switcher and search integration are intentionally excluded and will be delivered in separate PRs.
- Menu configuration currently includes test scenarios for multiple edge cases; final configuration refinements will follow.


astro/package.json
```diff
@@ -38,13 +38,15 @@
   "devDependencies": {
     "@csstools/postcss-global-data": "^4.0.0",
     "@eslint/js": "^9.39.2",
+    "@iconify-json/fluent": "^1.2.38",
```
g-francesca, line 37: - [x] to decide whether we want to add this or just use a different icon from phosphor icon

astro/src/components/primitives/Button/Button.css
```diff
@@ -2,84 +2,86 @@
  * Button Component Styles
  */
 
-/* ============================================
-   BASE STYLES
-   ============================================ */
+@layer primitives {
```
g-francesca, line 5: Most of the diffs in the primitive CSS are due to the entire stylesheet being wrapped in a @layer rule

legacy/de/guide/migrating-5.md
```diff
@@ -99,7 +99,7 @@ Anfänglich wurde `del` statt `delete` verwendet, weil `delete` in JavaScript ei
 {% capture codemod-deprecated-signatures %}
 You can replace the deprecated signatures with the following command:
 
-```plain-text
+```plaintext
```
g-francesca, line 102: I was getting an error when using plain-text, for this reason you will find this changed whenever it was uded in legacy md files

astro/package.json
ShubhamOulkar: fluent:search-sparkle-20-regular looks like an AI-powered search the sparkle suggests smart results. Phosphor doesn't have the same icon. We can keep this Fluent icon for clarity. If fluent has equivalent icons like phosphor then removing phosphor is good choice.

astro/package.json
```diff
@@ -32,19 +32,21 @@
     }
   },
   "engines": {
-    "node": ">=24.13.0",
+    "node": "24.x",
```
ShubhamOulkar: Error (EBADDEVENGINES) happened in github codespace because devEngines.runtime required >=24.13.0 but Codespaces had v24.11.1, so npm refused to install.

I updated engines.node and devEngines.runtime.version to major 24.x.

astro/package.json
```diff
@@ -22,7 +22,7 @@
   "devEngines": {
```
ShubhamOulkar, line 22: This is redundent.

ShubhamOulkar:
Theme and i18n buttons are not accessible if user click on docs or api pages. It should be accessible everywhere.

|problem|suggestion fix|
|-----|------|
||Screenshot 2026-02-06 180032|


Edit: @g-francesca, language button is dropdown/popup menu of 10 items. We need to integrate this in mobile menu sidebar. Same goes with theme button. IDK your thoughts but we might need a little changes in sidebar design.

astro/package.json
g-francesca: Good idea... I see that Fluent has many more icons than Phosphor, I can give it a try, and in case, remove Phosphor.
Screenshot 2026-02-06 alle 10 57 26

astro/package.json
g-francesca: thanks!

g-francesca:
- [x] Note from @bjohansebas: docs routes shouldn’t have the docs suffix, for example, "en/docs/v5/starter/installing/" should become "en/v5/starter/installing/" or "en/starter/installing/". This is to avoid breaking too many existing links

g-francesca:
> Theme and i18n buttons are not accessible if user click on docs or api pages. It should be accessible everywhere.
> 
> 
> 
> Edit: @g-francesca, button is dropdown/popup menu of 10 items. We need to integrate this in mobile menu sidebar. Same goes with theme button. IDK your thoughts but we might need a little changes in sidebar design.

@ShubhamOulkar I can:

- [x] review the mobile design to make the theme switcher and language selector always accessible within the sidebar.

Regarding the Middleware button, I may need some clarification.
At the moment, on mobile, the sidebar is a slide-in navigation with 3 levels (no dropdowns/popups):

The theme switcher is in the first level as a simple toggle, and I can make it persistent across all navigation levels.

Middleware is in the second level and opens a third-level navigation with around 10 items

My understanding from the Figma comments was to avoid dropdown/popup menus and use a flat + drill-down navigation structure, which is why I implemented it this way.

When you say “integrate this in mobile menu sidebar”, do you mean moving Middleware to the first-level navigation, or changing the current multi-level structure?

Could you please confirm the intended UX? Thanks!

ShubhamOulkar:
> When you say “integrate this in mobile menu sidebar”, do you mean moving Middleware to the first-level navigation, or changing the current multi-level structure?

you are doing right for level navigation for pages.

Theme and language buttons should be on 1st level but stick/fix position at the bottom of the sidebar so that user can access everywhere. [ref image](https://private-user-images.githubusercontent.com/91728992/546190329-8817eb65-6fb9-4384-8abb-ef7afea308ba.png?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NzAzODI3NDUsIm5iZiI6MTc3MDM4MjQ0NSwicGF0aCI6Ii85MTcyODk5Mi81NDYxOTAzMjktODgxN2ViNjUtNmZiOS00Mzg0LThhYmItZWY3YWZlYTMwOGJhLnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjAyMDYlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwMjA2VDEyNTQwNVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWQwYzVjNGZiMzBhZWMwZWM5NTA4MjdlYzZkYmVmNDE3MWIwYTZiZGFjNDBiMWZiMGE2MmM1NDZiNjI3NTVhNjcmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0._W7glnuQauSnNjR8s1UmKPt80GvsXCV5uzc1RpIr5z4)

Also I didn't seen any website with levels in language/theme component. It is default to use dropdown/popups for this components.

ShubhamOulkar:
> * Note from @bjohansebas: docs routes shouldn’t have the docs suffix, for example, "en/docs/v5/starter/installing/" should become "en/v5/starter/installing/" or "en/starter/installing/". This is to avoid breaking too many existing links

Keep `/en/starter/installing`. Only blogs links are going to break due to differences in jekyll and astro(we need to use redirects for old blog links).

astro/src/components/patterns/Header/Header.astro
```diff
@@ -0,0 +1,63 @@
...
+        size="24"
+        color="var(--color-icon-primary)"
+        aria-hidden="true"
+      />
+    </div>
+  </Flex>
+</header>
+
+<script>
+  const menuToggle = document.getElementById('menuToggle');
```
ShubhamOulkar, line 57: I think we need to decide design of component structure for our project. `<script>` tag makes it messy and hard to maintain. 

I recommend following component design stucture 
```
---
// Component Script (JavaScript)
---
<!-- Component Template (HTML + JS Expressions) -->
```

cc @expressjs/docs-captains @expressjs/docs-collaborators

g-francesca:
> > When you say “integrate this in mobile menu sidebar”, do you mean moving Middleware to the first-level navigation, or changing the current multi-level structure?
> 
> you are doing right for level navigation for pages.
> 
> Theme and language buttons should be on 1st level but stick/fix position at the bottom of the sidebar so that user can access everywhere. [ref image](https://private-user-images.githubusercontent.com/91728992/546190329-8817eb65-6fb9-4384-8abb-ef7afea308ba.png?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NzAzODI3NDUsIm5iZiI6MTc3MDM4MjQ0NSwicGF0aCI6Ii85MTcyODk5Mi81NDYxOTAzMjktODgxN2ViNjUtNmZiOS00Mzg0LThhYmItZWY3YWZlYTMwOGJhLnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjAyMDYlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwMjA2VDEyNTQwNVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWQwYzVjNGZiMzBhZWMwZWM5NTA4MjdlYzZkYmVmNDE3MWIwYTZiZGFjNDBiMWZiMGE2MmM1NDZiNjI3NTVhNjcmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0._W7glnuQauSnNjR8s1UmKPt80GvsXCV5uzc1RpIr5z4)
> 
> Also I didn't seen any website with levels in language/theme component. It is default to use dropdown/popups for this components.

Ok got it, the image helped. Thank you

astro/package.json
g-francesca: I've tried using fluent icons only and this is the result. They look good to me. wdyt?
Screenshot 2026-02-06 alle 14 51 47
Screenshot 2026-02-06 alle 14 51 58

g-francesca:
> > > When you say “integrate this in mobile menu sidebar”, do you mean moving Middleware to the first-level navigation, or changing the current multi-level structure?
> > 
> > 
> > you are doing right for level navigation for pages.
> > Theme and language buttons should be on 1st level but stick/fix position at the bottom of the sidebar so that user can access everywhere. [ref image](https://private-user-images.githubusercontent.com/91728992/546190329-8817eb65-6fb9-4384-8abb-ef7afea308ba.png?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NzAzODI3NDUsIm5iZiI6MTc3MDM4MjQ0NSwicGF0aCI6Ii85MTcyODk5Mi81NDYxOTAzMjktODgxN2ViNjUtNmZiOS00Mzg0LThhYmItZWY3YWZlYTMwOGJhLnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjAyMDYlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwMjA2VDEyNTQwNVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWQwYzVjNGZiMzBhZWMwZWM5NTA4MjdlYzZkYmVmNDE3MWIwYTZiZGFjNDBiMWZiMGE2MmM1NDZiNjI3NTVhNjcmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0._W7glnuQauSnNjR8s1UmKPt80GvsXCV5uzc1RpIr5z4)
> > Also I didn't seen any website with levels in language/theme component. It is default to use dropdown/popups for this components.
> 
> Ok got it, the image helped. Thank you

@ShubhamOulkar, can you please look at this preview: https://www.awesomescreenshot.com/video/49169616?key=b28243e63366332bf09cc1b76070655a. Does it reflect the expected behaviour?

ShubhamOulkar:
> can you please look at this preview: https://www.awesomescreenshot.com/video/49169616?key=b28243e63366332bf09cc1b76070655a. Does it reflect the expected behaviour? 
 
Looks good. 

Just want to know, how will user change language on mobile screen? 
Most websites use dropdown/popups.
No need to do it in this PR.

 I will do code review tomorrow. Posted some comment on components design. Please share your reviews on it. 
Thank you for you support.

g-francesca:
> > can you please look at this preview: https://www.awesomescreenshot.com/video/49169616?key=b28243e63366332bf09cc1b76070655a. Does it reflect the expected behaviour?
> 
> Looks good.
> 
> Just want to know, how will user change language on mobile screen? Most websites use dropdown/popups. No need to do it in this PR.
> 
> I will do code review tomorrow. Posted some comment on components design. Please share your reviews on it. Thank you for you support.

Given the design, I believe a popup may work well for language selection on all breakpoints. On mobile, we may consider making it full-screen for easier use.

astro/src/components/patterns/Header/Header.astro
g-francesca: @ShubhamOulkar Are you suggesting we avoid `<script>` tags in general, or is there something specific about the `Header` implementation that feels messy?
The <script> tag here handles client-side interactivity. Did you mean we should extract this script to a separate file, or were you thinking of a different approach?

astro/src/components/patterns/Sidebar/Sidebar.astro
```diff
@@ -0,0 +1,88 @@
...
+  ],
+  defaultVersion = 'v5',
+  ...rest
+} = Astro.props;
+
+const lang = getLangFromUrl(Astro.url);
+const currentPath = Astro.url.pathname;
+const currentVersion = detectVersionFromUrl(currentPath, versions, defaultVersion);
+
+console.log('Astro pathname', currentPath);
```
bjohansebas:
```suggestion
```

astro/src/content/docs/en/5x/resources/community.md
bjohansebas, line 1: This resources folder should be global and not versioned, including the middlewares anyway, since that content comes from a script. We’ll later adapt that script so it points to the new location of these MDX files (which we’ll do after this PR), but in the meantime, this folder wouldn’t be part of the versioning. The rest of the versioning approach you have makes complete sense.

For the menu, I think we could adapt it so that if a certain submenu section doesn’t want to be versioned, or a specific page doesn’t want versioning, it simply points to the location without the versioning prefix.

astro/src/content/resources/en/middleware/cookie-session.md
bjohansebas, line 1: For version 3, we won’t be able to adapt the content, and I don’t think it’s worth the effort to update the guides for that version (almost no one uses it anymore). So for version 3, we could simply show the API reference, since that already exists.

For the menu, I think that if users select version 3, we could show a banner directly in the menu saying that there are no guides available for this version.

That said, we could still show the guides and have them point to the default version, I think that would be the ideal approach. Or what do you think would be the best way to handle this?

astro/postcss.config.mjs
```diff
@@ -1,11 +1,13 @@
 import postcssGlobalData from '@csstools/postcss-global-data';
 import postcssCustomMedia from 'postcss-custom-media';
+import postcssNested from 'postcss-nested';
```
bjohansebas, line 3: I don’t think this is necessary, CSS already has this built in, and the old website was already using it, so it should be safe to rely on it. All modern browsers support it

https://caniuse.com/css-nesting

astro/src/content/docs/en/3x/api/request/overview.md
```diff
@@ -0,0 +1,122 @@
...
+---
+
+<h2 id="request">Request</h2>
+
+The `req` object is an enhanced version of Node's own request object
+and supports all [built-in fields and methods](https://nodejs.org/api/http.html#http_class_http_incomingmessage).
+
+<section markdown="1">
+  {% include api/en/3x/req-params.md %}
+</section>
```
bjohansebas, line 13: For the future (and this shouldn’t be done in this PR, since we don’t want to make it any bigger) we could add a card-style layout, similar to the cards used on the Next.js website.

Image

bjohansebas (commented):
A few comments: in addition, there are 32 accessibility issues reported by Astro. It would be worth looking into them and fixing them

Image

Content changes are not necessary in this PR and should be avoided

Also, in the contribution file, I think it’s worth adding information about this, explaining how versioning works now.

astro/src/components/patterns/Header/Header.astro
bjohansebas: I think they’re referring to why, in the selector, the logic lives in a separate file, while here it’s inside the Astro file. My general thinking is that if the logic needed to make it work were fairly large, it should live in a separate file. But since this is just a small script, I don’t think it’s worth creating a new file for it.

astro/src/components/patterns/Header/Header.css
bjohansebas, line 1: Is there any reason to start using @layer?

bjohansebas (commented):
For the menu configuration files, could we put them into a folder called menu inside the config directory? I’m concerned about having more configuration files and them starting to get confusing, this would help with better organization.

astro/src/content/docs/en/5x/resources/community.md
g-francesca: @bjohansebas, at this point, do you think it makes sense to move Resources out of Docs and place it in the main navigation as a non-versioned section, like in these screenshots?
Screenshot 2026-02-09 alle 11 42 46
Screenshot 2026-02-09 alle 11 43 04

astro/src/content/resources/en/middleware/cookie-session.md
g-francesca: @bjohansebas Just to confirm my understanding of the requirements, before proposing a solution:

**Versioning:**

- **Getting Started** pages: versioned for v5/v4
- **Guides** pages: versioned for v5/v4
- **API Reference** pages: versioned for v5/v4/v3/v2
- **Advanced Topics** pages: versioned for v5/v4
- **Resources** pages: not versioned

**IA:**

- **Getting Started**, **Guides**, and **Advanced Topics** will be managed as sections under **Docs**.
- **Resources** will be placed under the root-level navigation item, alongside **Docs**, **API Reference**, **Support**, and **Blog** (currently it lives under Docs, to be confirmed if we want to keep it under Docs, or move to main nav level).

**Routing:**

**Docs pages**: all pages under Docs will follow the pattern:

- `/{lang}/{sectionName}/{pageName}` eg `en/starter/installing` (defaults to v5) 

PLUS:

- `/{lang}/{version}/{sectionName}/{pageName}` eg `en/v4/starter/installing` (for all supported versions)

**API Reference pages**

- `/{lang}/{version}/api` → index page eg `/en/v5/api` (as-is)
- `/{lang}/{version}/api/{method|property|event}/{name}` eg `/en/v5/api/methods/app-all` (new pages)

Can you please confirm?

astro/src/components/patterns/Header/Header.astro
g-francesca: At the moment, I’m keeping the <script> directly inside the Astro file (only where client-side interactivity is needed). The only exception is the Sidebar component, where I moved the script into a separate file because it had grown quite large.

I’d proceed this way. Please let me know if you agree and whether you’ve noticed any inconsistencies in the code I should address. Thanks!

astro/src/content/docs/en/3x/api/request/overview.md
g-francesca: agree and noted for future interactions

g-francesca:
> For the menu configuration files, could we put them into a folder called menu inside the config directory? I’m concerned about having more configuration files and them starting to get confusing, this would help with better organization.

Makes sense!

- [x] refactor config folder

astro/src/components/patterns/Header/Header.css
g-francesca: Right, thanks for asking @bjohansebas! I realized I missed noting this earlier.

I started using `@layer` to ensure a predictable cascade order, because when importing external local CSS files, component styles were sometimes overridden by global styles, and this was not the intended behavior.

As alternatives, we could use scoped styles:

- <style> blocks inside .astro files. Astro would automatically scope these, but I wanted to avoid large .astro files.
- CSS Modules (.module.css). This would also scope the styles automatically, and in that case, we might not need @layer for cascade control

Let me know what you think about that.

astro/postcss.config.mjs
g-francesca: @bjohansebas I added postcss-nested to allow more flexible nesting, since native CSS nesting has some current limitations:

- Nesting plain HTML selectors or multi-level descendant selectors can be tricky or unsupported in some browsers.
- BEM-style concatenations like `&--active` aren’t supported by native nesting yet.
- the `&` can not represent pseudo-elements

If we want to avoid adding postcss-nested, we could rewrite these cases using flat CSS selectors instead.

astro/package.json
g-francesca: I removed the `engines` field, as in our case using `devEngines` seemed more appropriate (https://docs.npmjs.com/cli/v11/configuring-npm/package-json#devengines
).
FYI, previous discussion about devEngines here: https://github.com/expressjs/expressjs.com/pull/2167#discussion_r2738354427

astro/src/content/resources/en/middleware/cookie-session.md
bjohansebas: > API Reference pages: versioned for v5/v4/v3/v2

For the v2 API, there will be a redirect to [https://github.com/expressjs/expressjs.com/tree/2x](https://github.com/expressjs/expressjs.com/tree/2x), as it’s currently set up on the site, so I don’t think we should add it to the dropdown menu.

> Resources will be placed under the root-level navigation item, alongside Docs, API 
Reference, Support, and Blog (currently it lives under Docs, to be confirmed if we want to keep it under Docs, or move to main nav level).

Good question about moving it to the root of the nav. I think it should live at the root of nav, @expressjs/docs-wg, WDYT?

> /{lang}/{version}/api → index page eg /en/v5/api (as-is)

It should be /{lang}/{version}/api → index page, e.g. /en/5x/api. That’s how the mini versioning we had was set up. (5x, 4x, 3x)


Otherwise, it looks good to me

g-francesca:
> A few comments: in addition, there are 32 accessibility issues reported by Astro. It would be worth looking into them and fixing them
> 
> Image
> Content changes are not necessary in this PR and should be avoided
> 
> Also, in the contribution file, I think it’s worth adding information about this, explaining how versioning works now.

Thanks for pointing this, it was mainly due to aria-label missing in config files. I've fixed them.
Yes, it makes sense to add info about versioning and content configuration in general. I’ll add some documentation once we’ve finished reviewing the navigation.

- [x] Add docs about content configuration and versioning to the contribution file

astro/src/content/api-reference/en/v5/express/express.raw.md
```diff
@@ -0,0 +1,42 @@
+---
+title: express.raw
+description: This is a built-in middleware function in Express. It parses incoming request
+menuTitle: express.raw
```
bjohansebas: Is there any reason for this? I see you modified it, but I’m not sure if you’re actually using it.

astro/src/components/patterns/Header/Header.css
bjohansebas: But either way, if you imported it (not a .module.css) from the Astro file, the styles would still be scoped only to that component, right? I don’t remember very well and I’m almost out of time to check.

And yeah, I understand not wanting to have very large files, so let’s use @layer.

astro/src/content/api-reference/en/v5/express/express.raw.md
g-francesca: no, that actually comes from a previous wip development and is no longer used, so removed it. Thanks for catching it

astro/src/content/resources/en/middleware/cookie-session.md
g-francesca: right, in my examples above, replace all `v5/v4/`... with `5x/4x`....
Ok for v2 API, I will not include it in the version switcher.

astro/src/components/patterns/Header/Header.css
g-francesca: Nope, seems like Astro does not automatically scope imported CSS unless it’s a `<style>` block inside the .astro file or a CSS Module. Ok, let's keep `@layer` then, ty

astro/package.json
ShubhamOulkar: Looks good to me

astro/src/components/patterns/Header/Header.astro
ShubhamOulkar: I misunderstood the script tag requirement for interactive components. I thought adding logic inside the --- section at the top was enough. Also only adds once if used multiple times👍. Thanks for making it clear.

astro/src/content/resources/en/middleware/cookie-session.md
g-francesca: I reviewed the navigation based on the requirements described above. To sum up:

- **Routes**: reviewed and aligned with the specifications above.
- **Content configuration**: updated so that `versioned` is no longer a boolean, but an array of supported versions. This allows us to selectively disable specific versions from a submenu.
- **Resources**: this is to be confirmed, but for now I’ve moved it to the root navigation level.

I also added documentation about versioning and content configuration to the [contributing doc](https://github.com/expressjs/expressjs.com/pull/2180/changes#diff-18fca66512b435b53e67c0c6f5a3ac8b68105ad48b0290adaaef611be3c74b50)

@bjohansebas

ShubhamOulkar: all routes looks good to me.

Just want to point out Docs on api page will confuse user. 
 
image

astro/src/pages/[lang]/[...slug].astro
```diff
@@ -1,37 +1,67 @@
...
+const breadcrumbs = await buildBreadcrumbs(lang, slug, 'docs', version);
+const { title, description } = page.data;
 ---
 
 <DocLayout
   title={title}
   description={description}
-  section={menu as MenuSection}
   breadcrumbs={breadcrumbs}
+  data-version={version}
```
Comment (ShubhamOulkar) [inline astro/src/pages/[lang]/[...slug].astro]: How we are going to use these data attributes? Need to pass as props on layouts.

astro/src/components/patterns/VersionSwitcher/VersionSwitcher.astro
```diff
@@ -0,0 +1,73 @@
...
+  {...rest}
+>
+  <Icon
+    name="fluent:tag-28-regular"
+    size="16"
+    color="var(--color-icon-secondary)"
+    aria-hidden="true"
+  />
+  <Select
+    id="version-select"
```
ShubhamOulkar: Please avoid duplication of `id` in HTML. This component is used two times so that id got duplicated.

astro/src/content/resources/en/middleware/cookie-session.md
g-francesca: Right... How do we want to manage the breadcrumbs in this case? As 5x > API > Application?

carlosstenzel:
Captura de Tela 2026-02-10 às 20 16 46

I like the language option in the navigation bar next to the search (similar to the Node.js website); I think it allows for better user experience.

astro/src/content/resources/en/middleware/cookie-session.md
g-francesca: FYI: API pages currently belong to the docs content collection and therefore use the same route handler as the rest of the documentation.

- **Content path:** `src/content/docs/{lang}/{version}/api/...`
- **Route file:** `src/pages/[lang]/[...slug].astro`

This results in URLs like:
`/en/5x/api/application/overview` which preserves the existing API URL structure (`/en/5x/api`).

Because of this setup, API pages currently show **Docs** in the breadcrumbs. I can easily remove that label specifically for API pages.

**Alternative approach:** We could also create a dedicated API content collection, for example:

- **Content path:** `src/content/api/{lang}/{version}/...`
- **Route file:** `src/pages/[lang]/[version]/api/[...slug].astro`

Since the content schema and layout are identical, I went with the first option to avoid unnecessary duplication and complexity. However, let me know if you’d prefer the second approach (separate API collection + dedicated route), and I can switch to that.

astro/src/pages/[lang]/[...slug].astro
Comment (g-francesca) [reply astro/src/pages/[lang]/[...slug].astro]: I double-checked this, and actually they're no longer used, so I removed them. Thanks for spotting.

g-francesca:
> Captura de Tela 2026-02-10 às 20 16 46
> I like the language option in the navigation bar next to the search (similar to the Node.js website); I think it allows for better user experience.

@carlosstenzel We can easily move it to the header, next to the search trigger. In that case, I would also consider moving the theme switcher there as well (same as node).

On mobile, Node moves the search trigger, theme switcher, and language selector into the dropdown menu (see screenshot). I’d keep these controls always visible in the header. We just need to ensure there’s enough touch spacing between the icons to maintain good usability.

wdyt?

@bjohansebas @ShubhamOulkar 

Screenshot 2026-02-11 alle 11 43 57

Screenshot 2026-02-11 alle 11 47 57

astro/src/components/patterns/VersionSwitcher/VersionSwitcher.astro
g-francesca: done, thx ✅

g-francesca:
> > Captura de Tela 2026-02-10 às 20 16 46
> > I like the language option in the navigation bar next to the search (similar to the Node.js website); I think it allows for better user experience.
> 
> @carlosstenzel We can easily move it to the header, next to the search trigger. In that case, I would also consider moving the theme switcher there as well (same as node).
> 
> On mobile, Node moves the search trigger, theme switcher, and language selector into the dropdown menu (see screenshot). I’d keep these controls always visible in the header. We just need to ensure there’s enough touch spacing between the icons to maintain good usability.
> 
> wdyt?
> 
> @bjohansebas @ShubhamOulkar
> 
> Screenshot 2026-02-11 alle 11 43 57 Screenshot 2026-02-11 alle 11 47 57

To help visualize how these actions would look in the header versus the sidebar, I’ve added them to both. This is a temporary setup and will be revisited once we make a final decision.

@carlosstenzel @ShubhamOulkar @bjohansebas

bjohansebas:
Yep, in my opinion the theme switcher and the language selector should always be visible, like Node.js has it, similar to the last change you made. For the search, I’m not sure how you’re planning to implement it, maybe a modal like in Node.js? Either way, I prefer having them easily accessible, since it reduces the number of clicks a user has to make to use those features.

I haven’t reviewed the code for the latest changes yet, I’ll do that tomorrow, but I wanted to share my opinion in advance, since the time difference might affect when I’m able to review again, usually in the afternoon/evening GMT-5.

One more note: we’ll need to fix that part, I know it was more of a demo of how you wanted it to look, but there’s a layout shift when switching between dark mode and light mode.

g-francesca:
> ep, in my opinion the theme switcher and the language selector should always be visible, like Node.js has it, similar to the last change you made. For the search, I’m not sure how you’re planning to implement it, maybe a modal like in Node.js? Either way, I prefer having them easily accessible, since it reduces the number of clicks a user has to make to use those features.
> 
> I haven’t reviewed the code for the latest changes yet, I’ll do that tomorrow, but I wanted to share my opinion in advance, since the time difference might affect when I’m able to review again, usually in the afternoon/evening GMT-5.
> 
> One more note: we’ll need to fix that part, I know it was more of a demo of how you wanted it to look, but there’s a layout shift when switching between dark mode and light mode.

Ok, so at this point, I think that we all agree to move both the theme switcher and the language selector into the header. Yep, I did notice the layout shift, I’m fixing it now, as we’ve reached a final decision on the placement.

Regarding the search: my plan was to stick to the Figma design, assuming this layout was already approved. The current proposal uses a full-screen search modal on mobile and tablet, and a lateral sidebar on desktop. Let me know if this works for you, or if you’d like to revisit the approach.

Thanks for the review, I’ll wait for your feedback today.

bjohansebas:
Okay, I like the code. The only thing I don’t think we should do is reload the page every time a different version is selected.

We can also complete the menu in a follow-up PR, since it’s quite a bit of manual work. What would be great to handle right away, though, is making sure the URL doesn’t change when selecting a new version.

I’m also trying to figure out what needs to be merged or added, since we haven’t been very good at adding APIs or keeping things up to date, which is something that needs to change  . Anyway, I’m tracking it here #2186 . It’s not something for Orama, but it is for the Express team.

g-francesca:
> Okay, I like the code. The only thing I don’t think we should do is reload the page every time a different version is selected.
> 
> We can also complete the menu in a follow-up PR, since it’s quite a bit of manual work. What would be great to handle right away, though, is making sure the URL doesn’t change when selecting a new version.
> 
> I’m also trying to figure out what needs to be merged or added, since we haven’t been very good at adding APIs or keeping things up to date, which is something that needs to change . Anyway, I’m tracking it here #2186 . It’s not something for Orama, but it is for the Express team.

@bjohansebas Thanks for the review! I have some concerns about removing URL changes for version switching:

- Not all pages exist in all versions: when switching to a version that doesn't have the current page, we need to redirect to a fallback page anyway (already implemented)
- No shareability: users can't share/bookmark links specific to a version
- Browser history broken: back/forward buttons won't work
- SEO problems: search engines can't index versions separately

I would keep the URL versioning but try to implement smooth SPA-like transitions between versions. What do you think?

bjohansebas:
No, I don’t mean removing the URL. I mean reloading the page when a version is selected, I don’t think we should do that. 

We can simply load the menu when a specific version is selected, without reloading the page.

g-francesca:
> No, I don’t mean removing the URL. I mean reloading the page when a version is selected, I don’t think we should do that.
> 
> We can simply load the menu when a specific version is selected, without reloading the page.

Got it, I misunderstood sorry! You basically want to give users the ability to explore what's available in different versions without reloading pages, right?

Example: I'm on `/en/5x/api/application/overview` → switch version to `v4` → URL stays the same (content stays v5) → lateral menu updates to show `v4` structure → page only reloads when clicking a menu item from v4.

If the example is correct and reflects your expectation, I have questions about a few critical edge cases:

**Active item:** Since I'm viewing v5 content but showing v4 menu, there would be no active item. Is this expected?
**Page refresh:** Should the menu remember the selected version (v4) or revert to match the URL (v5)?
**Recovery:** Is the version switcher the only way to return the menu back to v5?
**Internal links:** If I click a breadcrumb or a page link to another v5 page while the menu shows v4, should the menu stay on v4 or revert to v5?
**No visual change:** If the menu structure is identical between versions, nothing happens visually when switching. This may be perceived as an error. Is this acceptable?

Let me know so I can understand the user scenario you have in mind and implement this correctly. Ty
Of course, if you have any live examples to share, that would help.

bjohansebas:
> Example: I'm on /en/5x/api/application/overview → switch version to v4 → URL stays the same (content stays v5) → lateral menu updates to show v4 structure → page only reloads when clicking a menu item from v4.

yeah!

>: Since I'm viewing v5 content but showing v4 menu, there would be no active item. Is this expected?

yes it's expected

> Should the menu remember the selected version (v4) or revert to match the URL (v5)?

I think we should go back to menu 5, or to the version that’s in the URL

>  Is the version switcher the only way to return the menu back to v5?

Hmm, yes, by reloading the page or switching between URLs of the same version.

>  If I click a breadcrumb or a page link to another v5 page while the menu shows v4, should the menu stay on v4 or revert to v5?

revert to v5

> If the menu structure is identical between versions, nothing happens visually when switching. This may be perceived as an error. Is this acceptable?

I understand there might be some confusion, but if the guides or APIs are the same, it’s acceptable



>  if you have any live examples to share, that would help.

I’m not sure there’s any page that has implemented it yet

g-francesca:
> > Example: I'm on /en/5x/api/application/overview → switch version to v4 → URL stays the same (content stays v5) → lateral menu updates to show v4 structure → page only reloads when clicking a menu item from v4.
> 
> yeah!
> 
> > : Since I'm viewing v5 content but showing v4 menu, there would be no active item. Is this expected?
> 
> yes it's expected
> 
> > Should the menu remember the selected version (v4) or revert to match the URL (v5)?
> 
> I think we should go back to menu 5, or to the version that’s in the URL
> 
> > Is the version switcher the only way to return the menu back to v5?
> 
> Hmm, yes, by reloading the page or switching between URLs of the same version.
> 
> > If I click a breadcrumb or a page link to another v5 page while the menu shows v4, should the menu stay on v4 or revert to v5?
> 
> revert to v5
> 
> > If the menu structure is identical between versions, nothing happens visually when switching. This may be perceived as an error. Is this acceptable?
> 
> I understand there might be some confusion, but if the guides or APIs are the same, it’s acceptable
> 
> > if you have any live examples to share, that would help.
> 
> I’m not sure there’s any page that has implemented it yet

@bjohansebas Thanks for clarifying! Would you mind if I manage this in a different branch (created from this one)? Just to have the 2 different behaviours comparable.

Also, can you please clarify how you expect to handle this case:
I visit API menu -> I switch to v3 (I can do that since API page are versioned for v3) -> I interact with the menu and I go back to "Docs" menu.
Since docs pages are not versioned for v3, what should be the selected version in this case? Do we automatically move back to v5?

bjohansebas:
>  Would you mind if I manage this in a different branch (created from this one)? Just to have the 2 different behaviours comparable.

Do you want to handle this in another PR? I don’t have any problem with that, so you can merge this and continue working on the desktop part as well.

> Since docs pages are not versioned for v3, what should be the selected version in this case? Do we automatically move back to v5?

Here I think we should show the guides for the latest version and display a note saying that there are no guides for v3. We should also mention that v3 is in end-of-life and include a link to the support page.

g-francesca:
> > Would you mind if I manage this in a different branch (created from this one)? Just to have the 2 different behaviours comparable.
> 
> Do you want to handle this in another PR? I don’t have any problem with that, so you can merge this and continue working on the desktop part as well.
> 
> > Since docs pages are not versioned for v3, what should be the selected version in this case? Do we automatically move back to v5?
> 
> Here I think we should show the guides for the latest version and display a note saying that there are no guides for v3. We should also mention that v3 is in end-of-life and include a link to the support page.

mm yes, maybe better closing this PR.
Let me know if there's something else you would like to cover in this PR, or I can proceed with the merge. Thank you!

bjohansebas (approved):
From my side, everything looks good. We can proceed with the next PRs

astro/CONTRIBUTING.md
```diff
@@ -101,8 +107,52 @@ So you've found a problem that you want to fix, or have a site enhancement you w
 | `npm run dev`     | Start development server with hot reload |
 | `npm run build`   | Build production site to `./dist`        |
 | `npm run preview` | Preview production build locally         |
+| `npm run lint`    | Run ESLint to check for issues           |
+| `npm run check`   | Run type checking and format verification|
+
+## Submitting a pull request
+
+1. Create a new branch from `redesign`
```
ShubhamOulkar, line 117: Documenting `redesign` now (we'll change to `gh-pages` later). 
I'm reviewing the code now and will finish by tomorrow morning.

Please hold off on merging until then.

ShubhamOulkar:
I noticed that Body.astro, BodyMd.astro, BodySm.astro, and BodyXs.astro are nearly identical wrapper components.

We could significantly reduce code duplication by consolidating these into a single <Text> or <Body> component that accepts a size or variant prop (e.g., `<Body size="sm">` instead of `<BodySm>`). The same logic applies to the heading components (H1-H6), which could be unified into a single `<Heading level={1}>` component.

I've left them as is for now to match the current pattern, but it's worth considering for a future cleanup!

ShubhamOulkar (approved):

g-francesca:
> I noticed that Body.astro, BodyMd.astro, BodySm.astro, and BodyXs.astro are nearly identical wrapper components.
> 
> We could significantly reduce code duplication by consolidating these into a single or component that accepts a size or variant prop (e.g., `<Body size="sm">` instead of `<BodySm>`). The same logic applies to the heading components (H1-H6), which could be unified into a single `<Heading level={1}>` component.
> 
> I've left them as is for now to match the current pattern, but it's worth considering for a future cleanup!

Makes sense! I've noted this for future interaction. TY

PR #2169, opened 2026-01-29
chore: redesign in Astro

bjohansebas:
Preview Url :- https://deploy-preview-2169--expressjscom-preview.netlify.app/

> [!IMPORTANT]
> The conflicts will be resolved when we’re ready to release it to the public, so don’t worry about that—one of the captains will handle it when the time comes.

# Overview

The design aims to closely follow the original Figma as much as possible. However, due to time constraints, it was not feasible to keep it fully up to date, as Orama supported the redesign effort until April 3.

Throughout the development process, feedback was continuously gathered via Slack, and adjustments were made iteratively as the site evolved.

As part of this transition, and to provide a clear **before-and-after reference** for the API content, the `gh-pages` branch will be left untouched, while a new `main` branch has been introduced, which will become the default branch of the repository.

The following sections outline the **key decisions that were made** and explain why certain aspects of the site are implemented the way they are.

For more in-depth details on how the site works, everything is documented in the repository under the [`docs` folder](https://github.com/expressjs/expressjs.com/tree/redesign/docs).

---

## API Routes

At the beginning, each method and property had its own URL. Over time, we decided to group them by object (`req`, `res`, `app`, `router`) to reduce both build times and page size.

Previously, the site could reach up to **600MB** due to all generated i18n pages. With this change, the build process is now significantly faster and more efficient.

---

## Search

The original Figma proposal placed the search as a right-side panel. This was changed to a **modal-based search**, mainly because the original design didn’t account for having a Table of Contents (ToC) per page.

**Important note:**
This is the only part of the site that uses React. It was implemented following Orama’s recommendation to ensure a fast and high-quality search experience.

---

## Menu

Compared to the Figma design, the **language switcher** and **dark/light theme toggle** were moved next to the search.

This follows a pattern similar to the Node.js website.

---

## Versioning

The Figma design did not include page versioning. During development, we decided to introduce **versioning for documentation pages**.

How versioning works is documented in the repository under the `docs` folder.

---

## i18n

The versioning system remains intact.

Current issue:

* The Crowdin integration is not working properly (already in contact with their support).

For now:

* Non-English pages are **fallbacks from English**.

Once fixed:

* Proper translations will be available.

---

## Footer

The original footer design did not follow the foundation’s recommendations.

It was updated to include the proper copyright text, which was later also added to the current expressjs.com site.

---

## E2E testing

Setup playwright testing platform with GitHub actions. Total 48 tests are added for `hero, features, theme, i18n and footer` components.

[...]

PR #2183, opened 2026-02-12
fix: don't crop the OpenJS trademark on desktop

notsecret808:
This PR fixes the cropping of the OpenJS trademark. It is visible on the desktop viewport.

expressjs-openjs-trademark


jonchurch (changes requested):
Woah good attention to detail. This is largely a "my machine, my browser, my monitor" issue. 

I can only repro it on sub 100% zoom levels, and not consistently. As in, I see slight crop on 90%, 80%, but not 75%. It's a subpixel issue Im guessing, so even device DPI might be a factor idk.

That said, the overall issue is that the container is sizing itself via css, and the svg is attempting to fill it, but the width doesnt quite match the container. So we get a tiny bit of overflow.

A better fix is to size the svg itself via CSS.
```css
a.openjs-logo {
  display: inline-block;
  cursor: pointer;
}

a.openjs-logo svg {
  width: 125px;
  height: auto;
}
```


I'll let you do that if you like, but please first fix the line ending diff. I didnt realize this file was CRLF, but normalizing it to LF in this PR is not necessary. (making the change [from the web editor](https://github.com/expressjs/expressjs.com/edit/gh-pages/css/style.css) is honestly maybe the simplest, if your local setup is going to fight you to do conversion)

notsecret808:
Thank u for detailed feedback!

I have tested the suggested fix on the local build, but, unfortunately, it doesn't work. I think i would be better to use `display: flex` here instead. Also now line endings are fixed.

notsecret808:
@jonchurch could you please make a new review for me?

ShubhamOulkar:
> ```css
> a.openjs-logo {
>   display: inline-block;
>   cursor: pointer;
> }
> 
> a.openjs-logo svg {
>   width: 125px;
>   height: auto;
> }
> ```

Agreed with @jonchurch suggestions.

css/style.css
```diff
@@ -782,7 +782,7 @@ footer {
 
 /* icons */
 a.openjs-logo {
-  display: inline-block;
+  display: flex;
```
```suggestion
  display: inline-block;
```
This is only child component so it is not ideal here.

css/style.css
```diff
@@ -782,7 +782,7 @@ footer {
 
 /* icons */
 a.openjs-logo {
-  display: inline-block;
+  display: flex;
   width: 125px;
   height: 39px;
   cursor: pointer;
```
```suggestion
```
Anchors already have pointer cursor by default.

ShubhamOulkar (changes requested):
You can also use following for better scaling.
```
a.openjs-logo {
  display: inline-block;
  width: 125px;
  height: 39px;
}
a.openjs-logo svg {
  width: 100%;
  height: 100%;
} 
```

notsecret808:
Thank u, I will try it later and will update my PR.

css/style.css
notsecret808: Agreed. Now it's working as exprected. Sorry, I think that the problem was on my side during the building of the app.

ShubhamOulkar:
Hi 👋 @notsecret808 
I just noticed that the logo has a full-width clickable area. Could you please help us fix the width issue so that the logo is constrained to a maximum width of 1110px?

image

ShubhamOulkar (approved):
Tested on chrome and FF (80% to 400% zoom levels)

LGTM

ShubhamOulkar:
cc @jonchurch

notsecret808:
> Hi 👋 @notsecret808 I just noticed that the logo has a full-width clickable area. Could you please help us fix the width issue so that the logo is constrained to a maximum width of 1110px?
> 
> image

Hi @ShubhamOulkar !

The clickable area is already constrained to a maximum width of 1110px (see the screenshot). Should we also limit the icon’s clickable area on mobile to it's natural size?

image

bjohansebas:
Thanks for your contribution, but unfortunately we’re going to leave this as is. We’re very close to launching the new site, and I can’t remove the block from @jonchurch, so I’m going to close this since we’re no longer interested in fixing things on the old site or spending more time on something that’s about to be replaced.

I encourage you to check out the new site and see what you can fix before it launches (https://github.com/expressjs/expressjs.com/pull/2169)

PR #2184, opened 2026-02-12
fix(navmenu): enable scrolling at 400% zoom for accessibility

### LINKED ISSUE #1731: (a11y) -- menu is broken at higher zoom levels
zoom to 400%

only some menu items appear, and users cannot scroll to see all the menu items

Screenshot 2025-01-13 at 2 53 07 PM

omerbakr:
This change constrains the menu height relative to the fixed header using a CSS variable and enables vertical scrolling within the menu container. The layout and visual design remain unchanged.

Fixes #1731

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1,3 +1,7 @@
+:root {
+  --header-height: 57px;
+}
+
 * {
   box-sizing: border-box;
   -webkit-tap-highlight-color: transparent;
@@ -34,6 +38,7 @@ body {
 
 header {
   grid-area: header;
+  height: var(--header-height);
 }
 
 .content {
@@ -1699,6 +1704,9 @@ blockquote {
 
   #navmenu.opens {
     display: block;
+    max-height: calc(100dvh - var(--header-height));
+    overflow-y: auto;
+    overscroll-behavior: contain;
   }
 
   #navbar a {


bjohansebas:
@expressjs/docs-wg  could someone please take a look at this? My schedule is pretty full with the redesign and other topics.

Note: I don’t think we should spend more time fixing the current site’s issues. Let’s focus on making sure the redesign doesn’t have these problems and on launching it as soon as possible, rather than investing more time in the current site.

css/style.css
```diff
@@ -1699,6 +1704,9 @@ blockquote {
 
   #navmenu.opens {
     display: block;
+    max-height: calc(100dvh - var(--header-height));
+    overflow-y: auto;
+    overscroll-behavior: contain;
```
ShubhamOulkar, line 1709:
```suggestion
```

I think `overscroll-behavior` property is redundant because body get applied with `no-scroll` when menus open. But this property will stop outer body scroll if js will fail to add 'no-scroll' class. 

@omerbakr WDYT? Should we keep it or not?

ShubhamOulkar (approved):
LGTM, design breaks after 220% zoom level and scroll appears after 220% break point. [Ref bug fix (test removed)](https://www.awesomescreenshot.com/video/49794131?key=b2313949f3574e1c73468b63374c89b1)

css/style.css
omerbakr: You’re right, it works fine without it. I’m just +1 on keeping it as a tiny CSS safety net in case the `no-scroll` class doesn’t get applied - it’s only one line, so feels harmless to keep.

ShubhamOulkar:
Thank you @omerbakr for fixing this a11y issue.

PR #2192, opened 2026-02-21
feat: add PR link checking workflow and fix links

### LINKED ISSUE #1100: Remove Broken Links
I noticed when looking at a PR that there are some links which have broken over time. Egregiously, there are a few links to a website called `node-tricks.com` which is now serving up some kind of phishing uglyness. You can see this link listed under the Blogs on the [resources](https://expressjs.com/en/resources/books-blogs.html#blogs) page.

I propose we attempt to rid the site of 404'd and other links which have succumb to bit rot, and this issue is a place to track those efforts.

I used a tool called [`broken-link-checker`](https://github.com/stevenvachon/broken-link-checker) on only external links using the following command:

`npx broken-link-checker https://expressjs.com/ -roi`

**Warning: that command took 15 minutes to complete when I ran it**

It showed there's quite a few broken links, but this tool doesn't give me a good output of just broken links without writing a little code. I don't think it's really the tool we need (well someone could use it as a library and write something), but **we need a list of all the unique broken links, and where they occur on the site** so we can figure out the best way to fix them.

cc @expressjs/expressjs-com @wesleytodd @expressjs/triagers

bjohansebas:
After asking for some feedback on the Astro Discord about a link checker, they recommended [https://github.com/lycheeverse/lychee](https://github.com/lycheeverse/lychee) for this task. They also recommended [https://github.com/HiDeoo/starlight-links-validator](https://github.com/HiDeoo/starlight-links-validator), but it seems it only works with Starlight.

closes #1100


bjohansebas:
Okay @g-francesca and @expressjs/docs-wg, this is a pretty huge PR (something I wanted to avoid) and I'm going to merge it because I need it to unblock my next changes.

The most important thing to keep in mind about these changes is that the API now has its own Astro collection. It is no longer tied to the `docs` collection where everything was versioned and had i18n. Instead, the API is now only in the API collection, with no i18n.

I also added a fallback for pages that don't have a translation yet, pointing directly to the English version. This helps avoid a lot of 404s right now, since otherwise I wouldn't be able to keep moving forward. And honestly, I quite like the fallback :)

I know there are links between API pages that point to the wrong version; that will be fixed in #2193.

If you find any problems, feel free to open a PR to fix them or open an issue.

PR #2193, opened 2026-02-22
feat: add remark plugin to rewrite localized links in Markdown content

bjohansebas:
A remark plugin to make it easier to reference links that are versioned and obviously have i18n. This will also greatly help prevent links from breaking when content is translated in Crowdin.

TODOs:

* Update the content to remove the 5x or 4x prefixes and remove `{page.lang}` (after this PR, we can start better localizing our content).
* Add unit tests for this.
* Add references in the contributing guide explaining this and how it works.

Also, it would be great to have #2192  merged first to ensure that links are always working properly.

This was generated with GitHub Copilot and reviewed by me.


astro/src/content/docs/es/5x/guide/routing.md
bjohansebas, line 1: This is completely temporary, just to test that it’s working for i18n.

ShubhamOulkar:
This is a useful plugin. I’ll start looking into it once we reach a solid stage in the redesign. I just don’t want to overload myself right now.

.github/scripts/get-readmes.sh
```diff
@@ -95,6 +103,25 @@ LIST_END
...
+  }ge')
+
+  # Apply configured URL replacements (dead/moved links → working alternatives).
+  # The lookahead requires the match to end at a URL boundary, so a URL that is a
+  # prefix of another (e.g. the site root vs a sub-path) isn't corrupted.
+  for pair in "${URL_REPLACEMENTS[@]}"; do
+    FROM="${pair%%|*}"
+    TO="${pair#*|}"
+    CONTENT=$(FROM="$FROM" TO="$TO" perl -pe 's/\Q$ENV{FROM}\E(?=[\s)">]|$)/$ENV{TO}/g' <<<"$CONTENT")
+  done
```
krzysdz, line 123: All readmes are from expressjs. Do we really need this link replacing logic if we can just fix them at the source?

bjohansebas: Yes, but it's easier for me to fix it from here since I have the permissions and can modify the file directly, rather than waiting for the maintainers of those packages to merge those PRs.

bjohansebas: I'd rather not depend on that when I know we're slow at merging things, especially for something as simple as the PRs you opened.

bjohansebas (commented):
.

PR #2195, opened 2026-02-24
chore: add pre commit husky setup

ShubhamOulkar:
This PR adds Husky pre-commit hook configuration to the project is located within the `astro`.


carlosstenzel (approved):

carlosstenzel:
How or who has access to the Socket report?

bjohansebas:
>How or who has access to the Socket report?

As far as I know, Ulises and Wes have access. I’m not sure who is supposed to have access, I have no idea, honestly. I would assume the security team, but I’m not certain.

This is more of a precaution to prevent accidentally adding a malicious dependency just because the name looks similar, or something like that.

astro/.husky/pre-commit
```diff
@@ -0,0 +1,2 @@
+cd astro
+npx lint-staged
```
bjohansebas: Use `node --run`.

astro/package.json
```diff
@@ -12,7 +12,8 @@
     "format": "prettier --write .",
     "format:check": "prettier --check .",
     "check": "npm run format:check && npm run lint",
-    "fix": "npm run format && npm run lint:fix"
+    "fix": "npm run format && npm run lint:fix",
+    "prepare": "cd .. && husky astro/.husky"
```
bjohansebas: What is this for?

ShubhamOulkar: This automatically setup git hook using husky right after the dependencies are finished installing.

astro/.husky/pre-commit
ShubhamOulkar: Optimised in https://github.com/expressjs/expressjs.com/pull/2195/commits/2574c92fb6b5e06cb4f4bd60af5dd1ceb19b1f02

astro/package.json
bjohansebas: Once the site is in the root, won’t this be necessary anymore? I think Husky already handles that for us. I just want to confirm that this is temporary and not something that will always stay there, because in all the Husky configurations I’ve had, this is the first time I’ve seen it done this way (or maybe my memory is failing me already 😅 the exhaustion haha).

ShubhamOulkar: Yeah, this is temporary. Right now the project uses a subfolder structure, so Husky needs the path to the .git root. Once the site is moved to the root, we'll switch to the standard "prepare": "husky" command.

This will help to anyone contributing to redesign branch gets their local environment setup.

astro/package.json
```diff
@@ -12,7 +12,9 @@
     "format": "prettier --write .",
     "format:check": "prettier --check .",
     "check": "npm run format:check && npm run lint",
-    "fix": "npm run format && npm run lint:fix"
+    "fix": "npm run format && npm run lint:fix",
+    "prepare": "cd .. && husky astro/.husky",
+    "lint-staged": "lint-staged"
```
jonchurch: this is only necessary bc we are doing `node --run`

@bjohansebas is the goal to avoid npm? it could be `npm exec lint-staged` and we wouldnt need the passthrough `"lint-staged": "lint-staged"` script.

astro/package.json
jonchurch: @bjohansebas is your question whether the `prepare` script is temporary? Or the `cd ..` part of it?

prepare is necessary for husky to configure git to be aware of the git hooks. its something you can't keep in version control, the repo gitconfig to point to the script location (`.husky`). Their docs show this specific cd example for when the package.json is not at git root: https://typicode.github.io/husky/how-to.html#project-not-in-git-root-directory

It is one time setup, but setup that needs to be run whenever someone clones the repo fresh (like when a new contributor forks the repo).

It's one of the downsides to these kinda of git hook tools, there is setup that cannot be checked in and must be run on clone.

astro/.husky/pre-commit
jonchurch: @bjohansebas was the goal to avoid using npm? this approach requires we add a passthrough `"lint-staged": "lint-staged"` script, which isnt something Im used to seeing.

I wanted to suggest using the --cwd arg to lint-staged and dropping the `lint-staged` script, but npm exec (npx replacement) wont be able to detect the installed bin if we don't help it discover a package.json

```sh
npm exec lint-staged --cwd astro
```

Im being pedantic here a bit, but we could instead drop the cd and the script by setting the npm prefix, which is how you tell npm to act as if it is in that dir and do its resolution from there:

```sh
npm exec --prefix astro lint-staged
```
but lint-staged does care about cwd for its config lookup, so to avoid cd we'd need...
```sh
npm exec --prefix astro lint-staged -- --cwd astro
```
That sucks though and is cursed lol, so cd is fine for now.

jonchurch (commented):
In general I do not like husky or lint-staged. I won't block this, as it's mostly PTSD from working in a monorepo with several hundred devs in my last job.

It's definitely a nice to have convenience, but I prefer that folks solve these issues in their editors and in CI. There are multiple ways to commit to a repo and not have the changes go through the git hooks. Most relevant to us and this repo though is going to be folks making changes via the github UI itself, or when making suggestions inline on PRs. Its a small price for us to pay here, it introduces the possibility of unrelated diffs on PRs due to auto-fixing code that was not properly formatted before merge.

Its super annoying I know, but ideally CI blocks improperly formatted code from making it into main as the final gatekeeper. 

<details><summary>The PTSD...</summary> comes from folks having completely unrelated diffs show up in their PRs on other parts of the monorepo, and not being able to figure out why or how to get rid of them. The answer, after spending too much time on it, was that lint-staged will run against merge commits (bc they are commits) which usually hapens after you pull from upstream. So even though you think you're linting only what you've staged, it will also lint fix other people's code in a merge commit. If you dont know how to fix this (which is just committing w/ no-verify) then you basically get stuck with frustrating phantom changes you dont know how to escape. Even for devs at my previous FAANG job, almost no one knew how to deal with this as it is not obvious.

I forget what other issues I had with Husky itself lol, but damn that lint-staged mystery persisted and was a huge pain in the ass due to diffs on unrelatd team's code triggering codeowner requirements before merge... sigh

</details>

I don't think we'll have many of the issues that haunt my memories, but it's worth pointing out that these tools add in failure modes that do not exist without them, and we should understand that if we enforce at commit time code fixing, we should also enforce in CI the same rules.

ShubhamOulkar:
>we should understand that if we enforce at commit time code fixing, we should also enforce in CI the same rules. 

Yes, I will add ci before merging into main branch gh-pages.

astro/.husky/pre-commit
ShubhamOulkar: I accepted this because of performance. Since we are on Node 24, it is significantly faster than npx or npm run (it avoids the npm CLI startup overhead).

You're right that it requires the 'passthrough' script in package.json which is the trade off for that speed boost. I agree that the --prefix and --cwd alternatives feel a bit 'cursed' the current cd approach keeps it readable while staying fast. Once we move everything to the root, we can simplify this even further.

jonchurch:
Eh its okay @ShubhamOulkar, I dont intend to block this. Just wanted to climb onto my soap box a bit.

bjohansebas:
> we should understand that if we enforce at commit time code fixing, we should also enforce in CI the same rules.

> > Yes, I will add ci before merging into main branch gh-pages.

The linter has already been a required check to merge, I added that rule when the work started. Once the website is launched, the same change will be applied so that passing the linter is mandatory there as well (which only the captains can configure).

Phillip9587:
We could also adapt something like https://autofix.ci/ to solve this problem

ShubhamOulkar:
@Phillip9587 , this is good GH action. But don't know organisational rules for adaptation. Since it is related to read write permissions to third party. If you know how to deal with this, please help me to do it.

bjohansebas:
The use of the actions is left more at the discretion of the captains. In my opinion, it seems fine to use them. The only issue is that I think when they’re forks, it might not work correctly using only pull_request, right? If someone can clarify that for me, I can take it into consideration.

Phillip9587:
> The use of the actions is left more at the discretion of the captains. In my opinion, it seems fine to use them. The only issue is that I think when they’re forks, it might not work correctly using only pull_request, right? If someone can clarify that for me, I can take it into consideration.

It is not only an action but we would also need to install the autofix.ci Github App (https://github.com/marketplace/autofix-ci) for permissions. As i understand it it also works for forks. From their website: 

>Supports all pull requests (including from forks) and branches without branch protection.

> For security reasons, pull requests from forked repositories do not have any write permissions which they could use to update themselves. The autofix.ci app provides pull requests with tightly-locked self-update capabilities. Alternative approaches such as processing untrusted code in pull_request_target or passing around personal access tokens are highly dangerous to outright insecure.

bjohansebas (approved):
@ShubhamOulkar please rebase. There are now more files, so **Prettier** is taking about 11s, and there will be even more files. We should probably try **oxfmt** to reduce that time.

At least **lint-staged** only runs on the files that were modified, so it’s not really a problem.

ShubhamOulkar:
> We should probably try **oxfmt** to reduce that time.

OXC still do not support astro https://github.com/oxc-project/oxc/issues/19715. Happy to explore oxfmt as a follow-up improvement in a separate PR if the team wants to pursue it.

> At least **lint-staged** only runs on the files that were modified, so it’s not really a problem

yes, won't affect the pre-commit hook in practice

cc @bjohansebas

PR #2203, opened 2026-03-02
feat(redesign): Homepage Hero component

g-francesca:
[Preview](https://deploy-preview-2203--expressjscom-preview.netlify.app/en/)
This PR implements the homepage Hero component.

### What's included

- Background video (light/dark mode variants) with poster images for mobile fallback
- Code blocks using Expressive Code. It was installed primarily to get a built-in copy button on the Hero code blocks, something Shiki doesn't provide out of the box.
- Fully responsive layout

----- 
Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.


astro/src/components/patterns/Hero/Hero.astro
```diff
@@ -0,0 +1,313 @@
...
+import { Image } from 'astro:assets';
+import { Code } from 'astro-expressive-code/components';
+import { Icon } from 'astro-icon/components';
+import { H1, Button, Container, Flex, BodyMd, Grid, Col } from '@/components/primitives';
+import { getLangFromUrl, useTranslations } from '@/i18n/utils';
+
+const lang = getLangFromUrl(Astro.url);
+const t = useTranslations(lang);
+
+const EXPRESS_VERSION = '5.2.1';
```
bjohansebas: Currently we have an automation on the site to update the version. I’m not sure how to suggest it here, since it could be done with a fetch on each build or we could continue with the automation.

krzysdz (commented):
I was wondering what is the pause/play button supposed to do, because in my browser the (light) video did not play so it did nothing. The light video is encoded with 4:4:4 chroma subsampling, which is not supported by my browser's decoder. AVC/H264 4:4:4 is not even listed in [AMD](https://www.amd.com/en/products/graphics/radeon-for-creators/video-editing.html#tabs-b6a36ad588-item-4a96864c55-tab), [Intel](https://www.intel.com/content/www/us/en/docs/onevpl/developer-reference-media-intel-hardware/1-1/overview.html#DECODE-SUPPORT) or [Nvidia](https://developer.nvidia.com/video-encode-decode-support-matrix) video decode support matrix.

> Error Code: NS_ERROR_DOM_MEDIA_FATAL_ERR (0x806e0005)
Details: auto __cdecl mozilla::SupportChecker::AddMediaFormatChecker(const TrackInfo &)::(anonymous class)::operator()(void) const: Decoder may not have the capability to handle the requested video format with YUV444 chroma subsampling.

The video files are small, which is nice, but look very much compressed and so does the light static fallback image.

If someone blocks autoplay, then the initial button state is wrong - pause button must be clicked to play the video (later the state is synchronised).

The button, when hovered, is partially clipped by the code block in mobile view (open page with enough width to load in desktop mode and then make it smaller to force mobile mode).
pause button clipped on top

astro/src/components/patterns/Hero/Hero.astro
g-francesca: Right, I can fetch the latest version from the npm registry, so we're sure it's updated on every build. I can add a fallback (that can be just for example 5.x) just in case the npm registry is not reachable. wdyt?

g-francesca:
> I was wondering what is the pause/play button supposed to do, because in my browser the (light) video did not play so it did nothing. The light video is encoded with 4:4:4 chroma subsampling, which is not supported by my browser's decoder. AVC/H264 4:4:4 is not even listed in [AMD](https://www.amd.com/en/products/graphics/radeon-for-creators/video-editing.html#tabs-b6a36ad588-item-4a96864c55-tab), [Intel](https://www.intel.com/content/www/us/en/docs/onevpl/developer-reference-media-intel-hardware/1-1/overview.html#DECODE-SUPPORT) or [Nvidia](https://developer.nvidia.com/video-encode-decode-support-matrix) video decode support matrix.
> 
> > Error Code: NS_ERROR_DOM_MEDIA_FATAL_ERR (0x806e0005)
> > Details: auto __cdecl mozilla::SupportChecker::AddMediaFormatChecker(const TrackInfo &)::(anonymous class)::operator()(void) const: Decoder may not have the capability to handle the requested video format with YUV444 chroma subsampling.

@krzysdz  Can you please check now and tell me if the video plays now? The pause/play button is supposed to play or pause the video on demand, I've also reviewed the style so that, in case the video fails loading for any reasons, the button is not displayed.

> The video files are small, which is nice, but look very much compressed and so does the light static fallback image.

Ok, so in this case I can try to review video/images to find a better compromise between quality and size.
 
> If someone blocks autoplay, then the initial button state is wrong - pause button must be clicked to play the video (later the state is synchronised).

Right, thanks for reporting it. It's fixed now

> The button, when hovered, is partially clipped by the code block in mobile view (open page with enough width to load in desktop mode and then make it smaller to force mobile mode). pause button clipped on top

As the video are not loaded on mobile for performance reasons, I've actually hidden the button on mobile.

astro/src/components/patterns/Hero/Hero.astro
bjohansebas: Yes, I think that sounds good. I think it would be fine to use [https://npm.antfu.dev/express](https://npm.antfu.dev/express) to simplify things.

astro/src/pages/[lang]/index.astro
Comment (bjohansebas) [inline astro/src/pages/[lang]/index.astro:1]: For the others, there are still some things to add, as well as the banner

astro/src/components/patterns/Hero/Hero.astro
bjohansebas: Now that I think about it, we don’t need that many automations anymore. Astro can handle them for us, which is pretty cool. I’m just saying this for pulling in the READMEs from our middlewares.
https://github.com/withastro/docs/blob/8e216293f74f38cc64db427622de74060403d5a8/src/content.config.ts#L168

g-francesca: Nice! I didn't know that. I've reviewed the content config to use the content collection loader

g-francesca:
> The video files are small, which is nice, but look very much compressed and so does the light static fallback image.

@krzysdz I've reviewed video/poster images to find a better compromise between size and quality. Let me know how they look to you now. TY

astro/src/components/patterns/Hero/Hero.astro
```diff
@@ -0,0 +1,323 @@
...
+    const observer = new IntersectionObserver(
+      (entries) => {
+        entries.forEach((entry) => {
+          if (entry.isIntersecting) {
+            video.play().catch(() => {});
+            observer.disconnect();
+          }
+        });
+      },
+      { rootMargin: '5rem' }
```
ShubhamOulkar:
```suggestion
      { rootMargin: '80px' }
```
I didn't review all the code, but this caught my attention. `rem` is not supported by `rootMargin`. Please fix it. Also, you can verify it in your console.

Actually it is spec bug https://github.com/w3c/IntersectionObserver/issues/308

ShubhamOulkar (changes requested):

astro/src/components/patterns/Hero/Hero.astro
krzysdz: > Uncaught DOMException: IntersectionObserver constructor: rootMargin must be specified in pixels or percent.

The [specification](https://w3c.github.io/IntersectionObserver/#dom-intersectionobserverinit-rootmargin) says that it has to be _"a string of 1-4 components, each either an [absolute length](https://drafts.csswg.org/css-values-3/#absolute-length) or a percentage"_.

Because of this exception more things are broken for me (e.g. the resize listener)

ShubhamOulkar:
Also Light theme looks weird.

image

EDIT :- It is because of my browser 3g throttle. We might have performance issue. Let's test it in separate PR. Now theme looks OK to me. 

image

krzysdz:
I can confirm that now both videos can be played by my browser.

The exception thrown by the `IntersectionObserver` aborts the script execution, so some more things are broken - e.g. I can start the video as "desktop", resize to mobile (video keeps playing, but play/pause button is gone) and then changing the theme doesn't work on the video allowing for light video in dark mode or video with black background in light mode (with black text on black background).

astro/src/components/patterns/Hero/Hero.astro
g-francesca: right, it's fixed now. thanks for catching it

g-francesca:
> Also Light theme looks weird.
> 
> image
> EDIT :- It is because of my browser 3g throttle. We might have performance issue. Let's test it in separate PR. Now theme looks OK to me.
> 
> image

~~mmm yes, I could reproduce that with slow connection. I've noted this for later fix. Thanks~~

I spotted the issue @ShubhamOulkar, should be solved now.

astro/astro.config.mjs
```diff
@@ -2,14 +2,21 @@
 import { defineConfig } from 'astro/config';
 import mdx from '@astrojs/mdx';
 import icon from 'astro-icon';
+import expressiveCode from 'astro-expressive-code';
 
 // https://astro.build/config
 export default defineConfig({
-  site: 'https://expressjs.com',
```
bjohansebas, line 8: We still need the site.

astro/src/pages/[lang]/index.astro
```diff
@@ -1,29 +1,20 @@
...
 ---
 
 <Layout>
-  <Container>
-    <H1>{t('home.welcome')}</H1>
-    <Body>{t('home.workInProgress')}</Body>
-    <ul>
-      <li>
-        {t('home.reviewDesignSystemPrefix')}
-        <a href={`/${lang}/ds-foundations`}>{t('home.reviewDesignSystemLink')}</a>
```
Comment (bjohansebas) [inline astro/src/pages/[lang]/index.astro:24]: I really like having that page, could we add it to the menu under the Resources section?

g-francesca:
> I can confirm that now both videos can be played by my browser.
> 
> The exception thrown by the `IntersectionObserver` aborts the script execution, so some more things are broken - e.g. I can start the video as "desktop", resize to mobile (video keeps playing, but play/pause button is gone) and then changing the theme doesn't work on the video allowing for light video in dark mode or video with black background in light mode (with black text on black background).

@krzysdz  I fixed the `IntersectionObserver`, so you should not reproduce the other related bugs. About the theme switcher not updating the video accordingly, is it a problem you can reproduce in specific cases? Or it's always like this?

astro/src/components/patterns/Hero/Hero.astro
bjohansebas, line 1: And when the user has animations disabled, does this respect that as well?

astro/src/pages/[lang]/index.astro
Comment (g-francesca) [reply astro/src/pages/[lang]/index.astro:24]: I added the ds-foundations page for internal reference, but of course if you find this useful, we can add it. I leave the decision to you, let me know in case :)

astro/astro.config.mjs
g-francesca: I removed it by mistake. Thanks for the catch. Fixed now.

astro/src/components/patterns/Hero/Hero.astro
g-francesca: @bjohansebas do you mean blocking the autoplay when animation is disabled?

bjohansebas: yep!

g-francesca: Yes, it's blocked if user preference is set to `reduced` motion

krzysdz:
> About the theme switcher not updating the video accordingly, is it a problem you can reproduce in specific cases? Or it's always like this?

It works now after the `IntersectionObserver` fix.

astro/src/components/patterns/Hero/Hero.astro
```diff
@@ -0,0 +1,333 @@
...
+    const prefersReducedMotion = window.matchMedia('(prefers-reduced-motion: reduce)').matches;
+    if (prefersReducedMotion) {
+      updatePoster();
+      video.pause();
+      video.removeAttribute('autoplay');
+      controlButton.style.display = 'none';
+      return;
+    }
+
+    const isMobile = window.innerWidth < breakpointMd;
```
ShubhamOulkar: why this logic is repeated? https://github.com/expressjs/expressjs.com/pull/2203/changes#diff-959176fc6245a55e46051932f45e47460870ef1acf3d57f64100ee864bd4a36dR250

astro/src/components/patterns/Hero/Hero.astro
```diff
@@ -0,0 +1,333 @@
...
+
+      updatePoster();
+
+      if (isNowMobile) return;
+
+      const currentSrc = video.querySelector('source')?.src;
+      const targetSrc = getVideoSource();
+
+      if (currentSrc && targetSrc && !currentSrc.endsWith(targetSrc)) {
+        const source = video.querySelector('source');
```
ShubhamOulkar, line 161: We dont need to call DOM multiple times. https://github.com/expressjs/expressjs.com/pull/2203/changes#diff-959176fc6245a55e46051932f45e47460870ef1acf3d57f64100ee864bd4a36dR183

astro/src/components/patterns/Hero/Hero.astro
```diff
@@ -0,0 +1,333 @@
...
+        source.src = initialSource;
+        video.load();
+      }
+    }
+
+    // Preloading the alternate theme video
+    const alternateVideoSrc = getCurrentTheme() === 'dark' ? videoDesktopLight : videoDesktop;
+    if (alternateVideoSrc) {
+      const link = document.createElement('link');
+      link.rel = 'preload';
```
ShubhamOulkar: Why not to use two simple `prefetch` links in the html?

```
<link rel="prefetch" href="/hero-background-light.mp4" as="video">
<link rel="prefetch" href="/hero-background.mp4" as="video">
```
Instead of dynamically adding a preload link for the alternate theme video, we might consider using static prefetch links (possibly desktop-only) for both theme variants and simply switch the <source> src on theme change. Since this is decorative background media, prefetch would be lower priority and less aggressive than preload, reducing bandwidth pressure while still keeping theme switching smooth.

ShubhamOulkar: You removed preload in https://github.com/expressjs/expressjs.com/pull/2203/commits/89531de6434f204aabf23674b8d5f323769432d4

ShubhamOulkar (changes requested):

astro/src/components/patterns/Hero/Hero.astro
```diff
@@ -0,0 +1,271 @@
...
+          autoplay
+          muted
+          loop
+          playsinline
+          preload="metadata"
+          data-src={videoSrc}
+        >
+          <source type="video/mp4" />
+        </video>
+        {videoSrcLight && (
```
ShubhamOulkar, line 75: Now it is rendering two videos, need fix here.

Image

ShubhamOulkar (changes requested):

ShubhamOulkar:
😅 It’s 8:41 PM, I’m hungry, and I’m just clicking anywhere at this point. That’s it for today. 👋 Bye!

astro/src/components/patterns/Hero/Hero.astro
g-francesca: Do you refer to the logic to disable video on mobile right? I moved it in a function that is called on load and on resize.

astro/src/components/patterns/Hero/Hero.astro
g-francesca: @ShubhamOulkar I added the 2 `<video>` elements intentionally. Only one visible at a time, depending on current theme. I did this to avoid a new network request on every theme switch.

Since one of two video is hidden via CSS, browsers do not composite or paint the hidden element. The files are also small (~600 KB each) and only loaded on desktop, so I thought  the memory overhead could be acceptable.

Of course, if you have suggestions on how to improve this, I'm happy to revisit.

ShubhamOulkar: Yes, I realised this after sometime (while eating). If performance is improved then I am ok with it.

ShubhamOulkar: My question to you 🤔 do we are going to add this background on all pages? Like api, blogs pages

g-francesca: The other pages do not include this component with the background video, that was developed for the homepage only.

bjohansebas (approved):
LGTM!

ShubhamOulkar (approved):
Sorry for a delay

SGTM!

g-francesca:
> Sorry for a delay
> 
> SGTM!

np! merging now, thanks!

PR #2208, opened 2026-03-13
Adjust layout content width for responsive sidebar

GroophyLifefor:
Adjusts content width behavior across breakpoints to better accommodate the sidebar layout and maintain consistent spacing in responsive views.

Video (before/after):

https://github.com/user-attachments/assets/92e71e3e-3463-4fd6-bdd3-8b204a752a39

### DIFF
--- a/astro/src/styles/base/_global.css
+++ b/astro/src/styles/base/_global.css
@@ -272,15 +272,16 @@ must be updated as well.
 */
 
 .layout-content {
-  max-width: calc(100dvw - 16px); 
+  max-width: calc(100dvw - 16px);
 }
 
 @media (--md-up) {
   .layout-content {
     max-width: calc(100dvw - 116px);
   }
-  
-  .layout-wrapper:has([data-toggle-nested-button][data-visible="true"][aria-expanded="true"]) .layout-content {
+
+  .layout-wrapper:has([data-toggle-nested-button][data-visible='true'][aria-expanded='true'])
+    .layout-content {
     max-width: calc(100dvw - 372px);
   }
 }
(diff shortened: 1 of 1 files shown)


bjohansebas:
It's not using the full width

image

GroophyLifefor:
> It's not using the full width

@bjohansebas The 16px margin was solving an initial issue, but the media queries I added later had already resolved it. I don’t think the 16px margin is still necessary, could you test all the pages again?

astro/src/styles/base/_global.css
```diff
@@ -248,3 +248,40 @@ main {
...
+@media (--md-up) {
+  .layout-content {
+    max-width: calc(100dvw - 100px);
+  }
+
+  .layout-wrapper:has([data-toggle-nested-button][data-visible='true'][aria-expanded='true'])
+    .layout-content {
+    max-width: calc(100dvw - 356px);
+  }
+}
```
g-francesca, line 254: @GroophyLifefor a suggestion here to avoid hardcoded max-width values, that depends on sidebar width. You can add (outside the media query):

```
.layout-content {
  min-width: 0;
}
```

The overflow problem was caused by the code block in the screenshot (and long code blocks in general), that forced the container to expand beyond viewport. This happened because the container had `min-width: auto` by default (as it's a flexbox item). 

Image

GroophyLifefor: Great point, when I tested It's not only fixing 284-287, It's also fix the core problem. Gonna do that.

bjohansebas (approved):
lgtm, thanks!

PR #2209, opened 2026-03-16
docs: add send options and methods changes to Express 5 migration guide

### LINKED ISSUE #1868: Add the changes in send to the migration guide
### Description

I just noticed that we haven’t documented all the changes made in send; we’ve only documented the change that send.mime no longer exists, but there were other changes that directly affect Express.

https://github.com/pillarjs/send/blob/master/HISTORY.md#100-beta1--2022-02-04

Discovered in https://github.com/expressjs/express/issues/6450

### Expectations

We need to document it

chakri11-git:
## Description

Fixes #1868

Added missing documentation to the Express 5 migration guide for 
breaking changes from the `send` package that directly affect 
Express users upgrading from v4 to v5.

## Changes made

**`en/guide/migrating-5.md`**

- Added new section `res.sendFile() and express.static() options` 
  documenting removed options:
  - `hidden` → use `dotfiles` instead
  - `from` → use `root` instead

- Added new section `Removed send methods` documenting methods 
  no longer exported by the `send` package:
  - `send.etag()` → use `etag` option instead
  - `send.index()` → use `index` option instead
  - `send.maxage()` → use `maxAge` option instead
  - `send.root()` → use `root` option instead

- Updated table of contents with links to both new sections

## Motivation

Developers upgrading from Express 4 to Express 5 who relied on 
these options and methods would encounter silent breakage with 
no guidance in the migration guide. This PR fills that gap.

### DIFF
--- a/en/guide/migrating-5.md
+++ b/en/guide/migrating-5.md
@@ -56,8 +56,10 @@ You can find the list of available codemods [here](https://codemod.link/express)
   <li><a href="#res.send.body">res.send(body, status)</a></li>
   <li><a href="#res.send.status">res.send(status)</a></li>
   <li><a href="#res.sendfile">res.sendfile()</a></li>
+  <li><a href="#res.sendFile.options">res.sendFile() and express.static() options</a></li>
   <li><a href="#router.param">router.param(fn)</a></li>
   <li><a href="#express.static.mime">express.static.mime</a></li>
+  <li><a href="#send-methods">Removed send methods</a></li>
   <li><a href="#express:router-debug-logs">express:router debug logs</a></li>
 </ul>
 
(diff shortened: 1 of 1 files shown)


krzysdz:
I haven't reviewed it thoroughly yet, but I'm pretty sure that some of the changes to `send` that you've listed do not directly affect Express users.

chakri11-git:
> I haven't reviewed it thoroughly yet, but I'm pretty sure that some of the changes to `send` that you've listed do not directly affect Express users.

Thank you for the feedback! You're right — after reviewing 
this more carefully, I can see that send.etag(), send.root(), 
send.index(), and send.maxage() were internal methods of the 
`send` package used by Express internally, not methods that 
Express users called directly in their applications.

I'll update the PR to remove that section and keep only the 
changes that are directly relevant to Express users (the 
res.sendFile() and express.static() options changes).

Sorry for the noise and thanks for the review!

PR #2238, opened 2026-03-31
docs: update more docs for the new website

bjohansebas:


ShubhamOulkar:
@bjohansebas why some of the advanced and guide pages are versioned? If we are going to version them then we need to add redirects for these old pages. Also breadcrumbs are not generating correct due to these differences.

Screenshot 2026-03-31 133635

bjohansebas:
We don't need redirects, there's a fallback that allows non-versioned pages and uses the latest version (i.e., 5.x) as the reference. You can check it by trying the URLs.

https://deploy-preview-2238--expressjscom-preview.netlify.app/en/5x/starter/installing/
and https://deploy-preview-2238--expressjscom-preview.netlify.app/en/starter/installing/, no need the redirects :)

ShubhamOulkar:
> We don't need redirects, there's a fallback that allows non-versioned pages and uses the latest version (i.e., 5.x) as the reference. You can check it by trying the URLs.
> 
> https://deploy-preview-2238--expressjscom-preview.netlify.app/en/5x/starter/installing/ and https://deploy-preview-2238--expressjscom-preview.netlify.app/en/starter/installing/, no need the redirects :)

This results in a poor user experience. You can see the issues in the breadcrumbs and sidebar navigation below 👇
image

It would be better to use redirects from `/en/starter/installing` to `/en/5x/starter/installing` instead of generating each non-versioned page. Also, note that we are currently generating English fallbacks for these non-versioned pages, which adds 8 × 10 more pages during the build.😊

PR #2246, opened 2026-03-31
[experiment] feat: refactor Hero component to use WebGL background

bjohansebas:


ShubhamOulkar (commented):
I’ve tried a similar approach before using SVG with a continuous `raf` loop, and it ended up being quite heavy on the main thread especially for background animations.

A few concerns with this approach, continuous WebGL/canvas animation for a  background is hurting performance, particularly on mobile and lower-end devices. There’s an initial blank state before the canvas renders, which is causing background blank visual flash. Replicating the current video background with code is time-consuming and hard to match visually (my main reason to skip svg implementation).

In general, these techniques are better suited for interactive elements, not passive backgrounds. For a quick check, just open the preview and monitor GPU usage you’ll likely notice the difference immediately. I’d recommend sticking with the current 1st load image then show video based approach ensures measurable performance gains.

g-francesca:
@bjohansebas the visual effect looks nice to me! Still, I see that there's a drop in performance/increased GPU cost. Also, I see there's no way to play/pause the animation.
I would probably investigate more on how to improve performance with this approach, or stick to the old approach.

krzysdz:
@bjohansebas changing theme on mobile (and probably with `prefers-reduced-motion`) does not update the background after today's changes.
With WebGL disabled (e.g. lockdown mode on Apple devices) the fallback is a plain white/black background, but I think it still looks OK.

bjohansebas:
@krzysdz fixed in https://github.com/expressjs/expressjs.com/pull/2283

PR #2258, opened 2026-04-03
docs: unify api pages

bjohansebas:


ShubhamOulkar:
Naturally we avoid long pages. BUt why you made these changes? my opinion is to maintain two pages for properties and methods.

Issue #2270, opened 2026-04-09
404 page in Additional examples


ahmxad:
### Description

By clicking on _[prisma-fullstack](https://github.com/prisma/prisma-examples/tree/latest/pulse/fullstack-simple-chat) - Fullstack app with Express and Next.js using [Prisma](https://www.npmjs.com/package/prisma) as an ORM_ on the link **https://expressjs.com/en/starter/examples.html**, it shows a 404 page not found error. This is the page: _https://github.com/prisma/prisma-examples/tree/latest/pulse/fullstack-simple-chat_

### Expectations

Not very user-friendly.

krzysdz:
For reference - the "Pulse" examples in `prisma-examples` have been deleted in https://github.com/prisma/prisma-examples/pull/7654.

PR #2276, opened 2026-04-15
docs(5x): clarify app.listen() callback differs from http.Server.listen() (#2255)

### LINKED ISSUE #2255: 5.x app.listen([port[, host[, backlog]]][, callback]) is not identical to Node’s http.Server.listen().
### Description

Documentation of [app.listen([port[, host[, backlog]]][, callback])](https://expressjs.com/en/5x/api.html#app.listen) states:

> This method is identical to Node’s [http.Server.listen()](https://nodejs.org/api/http.html#http_server_listen).

However, this is not true. In particular, calls to the callback differ.

See, for example expressjs/express#6191.

The new behaviour is documented in the migration guide: https://expressjs.com/en/guide/migrating-5.html

But the [5.x app.listen documentation](https://expressjs.com/en/5x/api.html#app.listen) still incorrectly indicates the express function is identical to the node http.Server.listen() function and does not document the details of callback calling that are unique to Express 5.x. This makes it difficult to understand what is happening when there are errors from the server (e.g. port already in use), if one is not following the migration guide.

It would be helpful if the 5.x documentation was updated to remove the statement that Express' listen() is identical to nodes http.Server.listen() and to document the Express specific handling of the callback.

 but the difference is not documented. 

### Expectations

The 5.x documentation of app.listen() should not state that app.listen() is identical to nodes' https.Server.listen() and it should document the Express specific handling of the callback arguments.

SAY-5:
Closes #2255.

The 5.x \`app.listen()\` documentation currently claims the method is *\"identical to Node's [http.Server.listen()](https://nodejs.org/api/http.html#http_server_listen)\"*. As reported in the issue, that's no longer true in Express 5: the optional \`callback\` is invoked with Express-specific arguments (see [expressjs/express#6191](https://github.com/expressjs/express/issues/6191)), and errors like \`EADDRINUSE\` are surfaced through the returned \`http.Server\`'s \`'error'\` event rather than through the callback parameters. The change is mentioned in the migration guide but not on the API reference page where most people land first when they hit a startup error.

This PR:
- Softens *\"identical to Node's\"* to *\"the forms accepted by this method are the same as\"*, which is the actual contract — the call shapes match, the callback semantics don't.
- Adds a \"Callback signature\" subsection that explicitly says the callback parameters are Express-specific, links to the 5.x migration guide and #6191, and shows the recommended pattern of attaching \`server.on('error', ...)\` to detect \`EADDRINUSE\` and friends.
- Applies the wording fix to both the \`(path, [callback])\` and \`([port[, host[, backlog]]][, callback])\` overloads.

Docs only — no code or schema changes.


krzysdz:
See https://github.com/expressjs/expressjs.com/pull/2278#issuecomment-4255195791. Everything that I wrote there applies to this PR too.

SAY-5:
Closing — the `redesign` branch has restructured these files and the changes here don't apply to the new layout. Apologies for targeting the wrong branch. If the underlying issues (#2255, #1941, #2270) still apply to the redesign, happy to re-do the fixes against that branch.

PR #2277, opened 2026-04-15
docs: note that app.METHOD() uses full-path matching, not prefix (#1941)

### LINKED ISSUE #1941: Mention different path matching strategies

SAY-5:
Closes #1941.

The shared \`routing-args\` partial used by \`app.use()\`, \`app.get()\`, \`app.post()\`, \`app.METHOD()\` (and friends) links to a single \"Path examples\" section. That section is sourced from \`app-use.md\` and only demonstrates the **prefix** matching used by \`app.use()\`: \`/abcd\` matches \`/abcd\` *and* \`/abcd/foo\`. The HTTP method handlers use **full-path** matching, but the docs never call out the difference, so users following the link from \`app.get()\` get the wrong mental model — exactly what's reported in #1941.

This PR adds a one-paragraph note next to the existing \"Path examples\" link in the routing-args partial (both 4.x and 5.x copies) explicitly stating that the linked examples illustrate \`app.use()\`'s prefix matching, and that \`app.get('/abcd', …)\` matches \`/abcd\` only — not \`/abcd/foo\`.

No content moved or restructured: the actual examples stay where \`app.use()\` readers expect them, and every other handler that pulls in this partial now points to them with accurate framing.

Docs only. Touches \`_includes/api/en/4x/routing-args.html\` and \`_includes/api/en/5x/routing-args.html\`.


krzysdz:
See https://github.com/expressjs/expressjs.com/pull/2278#issuecomment-4255195791. Everything that I wrote there applies to this PR too.

SAY-5:
Closing — the `redesign` branch has restructured these files and the changes here don't apply to the new layout. Apologies for targeting the wrong branch. If the underlying issues (#2255, #1941, #2270) still apply to the redesign, happy to re-do the fixes against that branch.

PR #2290, opened 2026-04-21
docs(5x): fix app.listen — remove incorrect 'identical to Node' claim

### LINKED ISSUE #2255: 5.x app.listen([port[, host[, backlog]]][, callback]) is not identical to Node’s http.Server.listen().

Sumit-Mayani:
Fixes #2255

## Problem

The 5.x API docs state `app.listen()` is **"identical to Node's http.Server.listen()"** — this is inaccurate. Express 5 changed the callback behavior: server errors (e.g. `EADDRINUSE`) are now passed to the callback instead of being thrown. This is already documented in the [migration guide](https://expressjs.com/en/guide/migrating-5.html#app.listen) but missing from the API reference.

## Changes

- Removed the false "identical to Node" claim from both `app.listen(path)` and `app.listen([port])` signatures
- Added accurate Express 5 error-handling description with a code example showing the `(error) => {}` callback pattern


Sumit-Mayani:
Closing this PR as it requires deeper Express 5 internals knowledge to verify accurately. Will revisit with more thorough research.

PR #2291, opened 2026-04-21
docs(5x): add per-route size limit example for express.json()

### LINKED ISSUE #1421: Document customizing `express.json()` size limit for specific routes
In the following issue, it is requested to customize the body size limit for `express.json()` dynamically based on the API route:

> @marbuser (Apr 11, 2019): 
> I've got some endpoints and 90% of them I want to have a limit of around '2MB'. However, there is 1 specific route, my upload route, where I want to have an upload limit of '10MB'.

- https://github.com/expressjs/express/issues/3932

There are some solutions provided in the thread, but maybe a simpler option would be this one inspired by [this Stack Overflow answer](https://stackoverflow.com/questions/74387929/increase-body-limit-for-specific-route-with-expressjs):

```js
// Parse application/json
// - limit on /upload: 5MB
// - limit on other routes: default of 100KB
app.use('/upload', express.json({ limit: '5mb' }));
app.use(/^(?!\/upload$)/, express.json());
```

This may be a nice example to document on [the `express.json()` docs section](https://expressjs.com/en/5x/api.html#express.json)

Sumit-Mayani:
Closes #1421

## Problem

`express.json()` docs show the `limit` option in the table (default `100kb`) but give no example of applying **different limits per route** — a very common need that currently sends users to Stack Overflow.

## Change

Added a short inline example after the options table showing how to pass `express.json({ limit: '10mb' })` directly to a specific route handler while keeping the default limit globally.


Sumit-Mayani:
Closing this PR to keep contributions focused. Will revisit with more thorough research.

PR #2292, opened 2026-04-22
ci: pin GitHub Actions to SHAs

jonchurch:
## The Problem
We want to ensure that our actions are pinned so we dont get rekt in a takeover

## The Solution

SHA pin all the actions being used on the redesign branch. 

SHAs were resolved via `gh api repos/OWNER/REPO/commits/TAG --jq .sha`

The main branch wasnt all pinned, so the lack of pinning wasnt necessarily new to the redesign, but this PR ensures that we pin them all.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,12 +31,12 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
           persist-credentials: false
 
       - name: Set up Node.js
-        uses: actions/setup-node@v6
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
         with:
           # node-version-file: ".nvmrc" use when .nvmrc is on root of the repo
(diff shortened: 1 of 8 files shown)


krzysdz (approved):
Verified all manually (only if hashes match versions; not a code audit). Some actions use different versions in different jobs and some don't use the latest version:

- `actions/setup-node` - v4 is used once, rest (3) uses v6 (latest version)
- `actions/upload-artifact` - v7 is available, v4 is used in 2 jobs, v6 is used in 1 job
- `actions/download-artifact` - v8 is available, v4 is used in 1 job
- `actions/deploy-pages` - v5 is available, v4 is used in 1 job
- `actions/cache` - v5 is available, v4 is used in 1 job

I did not check compatibility or review changes (some releases have only a few days and if we pin versions for security reason, the code probably should be audited when the version changes).

`github/codeql-action/*` actions probably don't require pinning by hash, because releases (and corresponding tags) are [immutable](https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases). A patch version tag (e.g. `v4.35.2`) should always point to the same commit if we trust GitHub to have implemented this feature correctly.

bjohansebas (approved):
LGMT! thanks!

PR #2303, opened 2026-05-05
docs: add verification step for Express installation

aryandevtyagi10:
### What
Added a "Verify your installation" section to the installing guide.

### Why
New users have no way to confirm Express installed correctly.
This adds a quick verification step using express-generator.

### Changes
- Added "Verify your installation" section at the end of 
  en/starter/installing.md

I certify that this contribution is my own work and submitted 
under the project's open source license.


krzysdz (changes requested):
`express-generator` is not `express`. Installing `express` won't install `express-generator`. Besides that `npx` offers you to install a package if it is not already installed, so using it to check whether a package is installed is not a great idea. [`npm ls`](https://docs.npmjs.com/cli/v11/commands/npm-ls) lists installed packages and I'd say is the recommended way to check what is installed.

Probably the best way to verify if Express is installed correctly is following the next step and creating a simple "Hello world" server.

aryandevtyagi10:
Thanks for the feedback! You're right — updated the 
verification to use `npm ls express` instead. Let me 
know if any further changes are needed.

PR #2309, opened 2026-05-16
docs: update res.redirect examples to use default 302

### LINKED ISSUE #2308: res.redirect: Code examples should show default HTTP status code 302 instead of 301
### Description

Was migrating from Express 4 to 5 recently. Followed the docs and replaced `res.redirect(url)` with `res.redirect(301, url)` as per the code examples in the migration guide. 

My bad on not checking that 301 is a permanent redirect. Only found out later when I could not log out from my application.

Read the fine text and saw that the status code, if not specified, defaults to 302, which is a temporary redirect. Updated my application code but had to clear browser cache.

### Expectations

Code examples for `res.direct(status, url)` should use the default status code, which is 302.
- https://expressjs.com/en/5x/api.html#res.redirect
- https://expressjs.com/en/guide/migrating-5.html#res.redirect
- https://expressjs.com/en/4x/api.html#res.redirect

Code examples should use default values to prevent confusion, as developers would likely jump straight to them instead of reading through paragraphs of text, especially when the method parameters are few and seemingly straightforward.

debojeetmitra:
Updates the `res.redirect` examples in the migration guide to reflect the default temporary redirect behavior (302) instead of explicitly using 301.

This helps reduce confusion during migration from Express 4 to 5 and avoids unintentionally encouraging permanent redirects.

closes https://github.com/expressjs/expressjs.com/issues/2308


bjohansebas (approved):

krzysdz:
@bjohansebas this is supposed to "fix" https://github.com/expressjs/expressjs.com/issues/2308

bjohansebas:
thanks @krzysdz and @debojeetmitra!

debojeetmitra:
Thank you for the review and merge! Looking forward to contributing more to Express.js in the future.

PR #2326, opened 2026-05-19
chore: update package-lock for latest npm 11

jonchurch:
This should stabilize the lockfile, Im getting a lockfile diff on npm i w/ npm 11.12.1, which we have pinned in the package.json now.

I think the diff here is bc npm 11 had some bugs or something w/ peer dep bookkeeping, so the lockfille was generated in an earlier version of 11, and since like 11.10 something whatever bugs existed were fixed. 

This diff is just from running `npm i` here with 
```sh
node -v
# 24.15.0
npm -v
# 11.12.1
```


bjohansebas (approved):
I wish npm had an option to freeze the lockfile instead of having to use npm ci. The same thing happened to me in webpack, so I’m not sure this will help much because it seems like it will always get regenerated

PR #2334, opened 2026-05-19
fix: version switcher 404 on non versioned pages

### LINKED ISSUE #2330: Changing API version in menu redirects to 404 page when using non-versioned page
### Steps to reproduce

1. Visit https://expressjs.com/en/blog/ or https://expressjs.com/en/support/
2. In side menu select "API" (or "Docs")
3. In the API (or "Docs") submenu click on version selector (showing 5.x)
4. Select any version (including 5.x)

### What happens

The page immediately redirects to `/<lang>/<selected api version/<support, blog or whatever else was here>` (e.g. `https://expressjs.com/en/4x/blog`), which does not exist.

### What should happen

- Links in the API (or "Docs") submenu should point to relevant docs for the selected version.
- Page should not reload and if it has to (~~I think this was an intentional choice~~), then it should do so only on pages that are already versioned.
  **EDIT:** The page should not reload according to https://github.com/expressjs/expressjs.com/pull/2180#issuecomment-3908458651

jonchurch:
closes #2330 

# Problem

There were two different handlers for version switching, and the buggy one was handling the version switch.

In the "buggy" one, it assumed we were on a versioned page, which was not always the case. So when switching from the blog `en/blog`, it would update the URL to be  `/en/4x/blog` and we'd 404

# Fix

Drop the other dumber handler, wire up the `SidebarVersionManager` to listen to the correct event.

The `SidebarVersionManager` correctly implements the case where we are on a non versioned page and switch versions from the sidebar, it will drop us into the sidebar section we have open at the correct version. But it was listening for the wrong event, so never had a chance to trigger the correct logic.

The other handler was dumber and did not have that logic, but was the one handling the event.

### DIFF
--- a/src/components/patterns/Sidebar/SidebarVersionManager.ts
+++ b/src/components/patterns/Sidebar/SidebarVersionManager.ts
@@ -10,15 +10,12 @@ export class SidebarVersionManager {
   }
 
   setup(): void {
-    const versionSwitchers = this.sidebar.querySelectorAll(
-      '[data-version-switcher] select, [data-version-select]'
-    );
+    const versionSwitchers = this.sidebar.querySelectorAll('[data-version-switcher]');
 
     versionSwitchers.forEach((switcher) => {
-      switcher.addEventListener('change', (e) => {
-        const select = e.target as HTMLSelectElement;
-        this.handleVersionChange(select.value);
-      });
+      switcher.addEventListener('select-change', ((e: CustomEvent<{ value: string }>) => {
+        this.handleVersionChange(e.detail.value);
+      }) as EventListener);
     });
   }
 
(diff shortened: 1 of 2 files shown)


src/components/patterns/Sidebar/SidebarVersionManager.ts
```diff
@@ -129,17 +126,4 @@ export class SidebarVersionManager {
...
-  }
-
-  setVersion(version: string): void {
-    const versionSelects =
-      this.sidebar.querySelectorAll<HTMLSelectElement>('[data-version-select]');
-    versionSelects.forEach((select) => {
-      select.value = version;
-    });
-    this.handleVersionChange(version);
-  }
```
jonchurch, line 144: These are unused, so I removed them.

bjohansebas (approved):

PR #2335, opened 2026-05-19
docs: use consistent API reference titles

krzysdz:
For some reason the title for 3.x in English docs was in Spanish.


krzysdz:
The 3.x description is also much different from others, but it is not user visible (except for search engines and things that show Open Graph previews).

bjohansebas (approved):

PR #2289, opened 2026-04-21
docs: remove broken prisma-fullstack example link

Sumit-Mayani:
Fixes #2270

The `prisma-fullstack` link points to `prisma/prisma-examples/tree/latest/pulse/fullstack-simple-chat` which returns a 404 — that path no longer exists in the prisma-examples repo.

Removed the broken entry from all 10 locale pages (`en`, `de`, `es`, `fr`, `it`, `ja`, `ko`, `pt-br`, `zh-cn`, `zh-tw`). The `prisma-rest-api-ts` entry below it remains as it points to a valid URL.

PR #2345, opened 2026-05-22
docs: remove prisma-fullstack example link(it not exist)

WuMingDao:
It link jump to this: 
image

And I have not found the same full-stack use express:

image


krzysdz:
Duplicate of #2289

PR #2311, opened 2026-05-17
docs: mention TypeScript generator options

popsiclelmlm:
Related to expressjs/express#7231

Summary:
- clarify that `express-generator` creates a JavaScript application skeleton
- add a short TypeScript section to the generator page
- reuse the existing community caveat before linking to a community TypeScript generator

Tests:
- `npm test -- --quiet en/starter/generator.md`

PR #2352, opened 2026-05-27
docs: mention TypeScript generator option

jhcpeixoto:
Refs expressjs/express#7231

Adds a short note to the English Express application generator docs for both 4.x and 5.x explaining that the official generator creates a JavaScript skeleton and linking to the community TypeScript generator suggested in the issue.

Checks:
- npm run check
- npm run build


jhcpeixoto:
I checked the failing linkChecker job. The new express-generator-typescript link returns HTTP 200, and the job log points to existing site-wide links such as https://slack-invite.openjsf.org/ returning 403 rather than the link added in this PR.

krzysdz:
Duplicate of #2311

PR #2353, opened 2026-05-27
docs: document hidden directory dotfiles behavior in v5 migration

### LINKED ISSUE #1868: Add the changes in send to the migration guide

jhcpeixoto:
Closes #1868

## Summary
- clarify that the Express 5 res.sendFile() dotfiles option applies to hidden directories in the path, not just hidden target files
- add an example matching the hidden mount directory case from expressjs/express#6450

## Verification
- npm run check
- npm run build
- git diff --check


jhcpeixoto:
I checked the failing linkChecker job. This PR adds text only and no new links. The failures are existing site-wide links such as https://slack-invite.openjsf.org/ returning 403, LinkedIn profile links, old Node API anchors, and translated-page local paths; they are not introduced by this change.

bjohansebas (approved):

PR #2333, opened 2026-05-19
fix: add missing redirects

krzysdz:
Cherry picked 879c8d82f6b647554ce5d936841247b31d245a96 from #2324

Fixes #2327
Fixes #2328
Fixes #2332
Fixes #2343

PR #2354, opened 2026-05-27
fix: redirect legacy default-language URLs

### LINKED ISSUE #2343: Old URLs without a language code in the URL no longer work
This is one of the two major types of broken link visible in the table I made at https://github.com/expressjs/expressjs.com/issues/2332#issuecomment-4499884195. (The other is https://github.com/expressjs/expressjs.com/issues/2338.) Per what I see in the Wayback Machine, prior to the recent redesign these URLs redirected to the /en/ version, which seems sensible.

jhcpeixoto:
Closes #2343\n\nAdds explicit redirects for legacy no-language documentation URLs to their current English equivalents, including the reported /starter/static-files.html path.\n\nVerification:\n- npx prettier --check src/config/redirect.js\n- git diff --check\n- npm run build\n- checked generated redirect pages for /starter/static-files.html and /guide/using-middleware.html


krzysdz:
Duplicate of #2333

Issue #2332, opened 2026-05-19
Old URLs ending in ".html" are 404s; should they redirect instead?


ExplodingCabbage:
[Via the Wayback Machine](https://web.archive.org/web/20260425190126/https://expressjs.com/en/advanced/best-practice-performance.html), I see that https://expressjs.com/en/advanced/best-practice-performance.html was a valid URL less than a month ago. Now it is a 404 page. It looks like the content that was once there is now located at https://expressjs.com/en/advanced/best-practice-performance/ instead.

Okay, fine, but this is super-unhelpful to me as a (first-time) end user. Google results currently link me exclusively to the 404 pages, and from a quick sample I would guess that every single one of the 2500 links to the Express docs on Stack Overflow is now a dead link. (I would also not be surprised if this were an incredibly potent act on anti-SEO; I don't know if historical pagerank will flow to the new docs URLs. It would be unfortunate if the official docs dropped off the Google results for Express-related searches as a result!)

I have not carefully dug into the recent changes that led to this, but is it not possible to simply have URLs ending in `.html` redirect to fix it?

ExplodingCabbage:
(Presumably this resulted from the new website I think you launched [yesterday](https://expressjs.com/en/blog/2026-05-18-a-new-look-for-express/)? I would consider this to be on-fire-priority, given the potentially permanent SEO implications of not addressing it quickly.)

krzysdz:
Yes, it should be possible to redirect them. I'll try to do this in a few hours, maybe separately from #2324.

See also #2327, #2328

jonchurch:
Thanks for the report, I thought we had taken care of this but apparently not

> I have not carefully dug into the recent changes that led to this, but is it not possible to simply have URLs ending in .html redirect to fix it?

This is something we can fix with Cloudflare Redirect Rules as a quick mitigation, working on it now

jonchurch:
This is live now in cloudflare, the `.html` stripping

We may want to re evaluate this, the directory style we ahve now is the astro default, we could also configure astro to format urls in file mode, to get back to the exact structure we had previously and not require CDN level redirs.

Please let us know if you notice a .html related 404!

ExplodingCabbage:
[...]

PR #2356, opened 2026-05-28
Switch from directory to file format in Astro

krzysdz:
The "file" format generates `name.html` files instead of default `name/index.html`. This would make old links work, even without redirects on Cloudflare level, while still allowing the current link format with no extension. The only observable difference between current version and this PR is that the URLs no longer end with a `/` (GitHub/Netlify serve a 301).

References:

- https://docs.astro.build/en/reference/configuration-reference/#buildformat
- #2332
- @jonchurch's [suggestion on Slack](https://openjs-foundation.slack.com/archives/C090970SCLE/p1779221992561499?thread_ts=1779218541.473839&cid=C090970SCLE)

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -23,6 +23,9 @@ const site = NETLIFY_PREVIEW_SITE || 'https://expressjs.com';
 export default defineConfig({
   redirects,
   site,
+  build: {
+    format: 'file',
+  },
   markdown: {
     // Link localization (Markdown links + raw HTML/JSX `<a href>`). Configuration —
     // localized sections, versioned sections, default version, and the "global" pages
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ on:
 # Cancel in progress workflows
 # in the scenario where we already had a run going for that PR/branch/tag but then triggered a new run
 concurrency:
-  group: '${{ github.workflow }} ✨ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}'
+  group: "${{ github.workflow }} ✨ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}"
   cancel-in-progress: true
 
 permissions:
@@ -36,8 +36,8 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
         with:
-          node-version-file: '.nvmrc'
-          cache: 'npm'
+          node-version-file: ".nvmrc"
+          cache: "npm"
 
       - name: Install Node.js dependencies
         run: npm ci
@@ -97,11 +97,11 @@ jobs:
         with:
           args: |
             --root-dir $PWD/dist
+            --fallback-extensions html
             --exclude-path dist/llms.txt
             --exclude-path dist/llms/
             --exclude-private
             --exclude-loopback
-            --remap "https://expressjs\.com\/((?:[^\/]+\/)*[^\/\.]+)\/?$ file://$PWD/dist/\$1/index.html"
-            --remap "https://expressjs\.com\/(.*\.html)$ file://$PWD/dist/\$1"
+            --remap "^https://expressjs\.com\/ file://$PWD/dist/"
             dist/
           fail: true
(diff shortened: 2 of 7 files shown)


krzysdz:
**EDIT:** Fixed by #2390

<details>
<summary>Original comment about <code>/index</code> in search results</summary>

This breaks search results for API pages, which for some reason end with `/index`. I'm not sure if this will fix itself (after sync) or if slightly changing source file structure is necessary (~~krzysdz@64ecc18438bf35e892679199189df5a011079e17~~ krzysdz@981a172a34f4a7a13de2c18d1326cb1c07d94049). I think that the current search behaviour is unexpected and should be changed whether this PR is merged or not.

@bjohansebas is there any reason why API docs live as `index.mdx` in separate directories instead of just `name.mdx` like other pages?

**EDIT:** Either API docs files must be moved and renamed or the code responsible for uploading pages to Orama must be changed to strip `/index`:

https://github.com/expressjs/expressjs.com/blob/ec1d0372968c6912548229831956787d1e07ab9d/scripts/orama-documents.mjs#L52-L72

What should work instead of renaming files (~~krzysdz@f1151f6c6d7ff3f02d212ac3f97bcc8984a6f6aa~~ krzysdz@4b1e61b3a67d09d657a1ee109558187c31d72f25):

```diff
diff --git a/scripts/orama-documents.mjs b/scripts/orama-documents.mjs
index 0b9a9668..5dd9831a 100644
--- a/scripts/orama-documents.mjs
+++ b/scripts/orama-documents.mjs
@@ -106,7 +106,7 @@ export const getApi = async (lang) => {
           const fullPath = join(baseDir, file);
           const raw = await readFile(fullPath, 'utf-8');
           const { data, content } = matter(raw);
-          const pathSegment = relative(baseDir, fullPath).replace(/\.mdx?$/, '');
+          const pathSegment = relative(baseDir, fullPath).replace(/(\/index)?\.mdx?$/, '');

           return {
             title: data.title ?? basename(file).replace(/\.mdx?$/, ''),
```

**EDIT 2:** This applies also to `/resources/index.mdx` and `/resources/middleware/index.mdx`, so doing it while uploading docs to Orama seems like a safer solution.

</details>

krzysdz:
As a bonus, with the changes from this PR we can just enable [`--include-fragments`](https://lychee.cli.rs/guides/cli/#--include-fragments) and the results won't be full of false positives.

In fact using lychee with `--include-fragments` on this PR is how I found #2363 and #2387.

astro.config.mjs
```diff
@@ -23,6 +23,9 @@ const site = NETLIFY_PREVIEW_SITE || 'https://expressjs.com';
 export default defineConfig({
   redirects,
   site,
+  build: {
+    format: 'file',
+  },
   markdown: {
```
bjohansebas, line 29: Okay, quite a bit of time has passed. Other than the `.html` extensions, have we run into any additional issues?

To be honest, I feel like we should move forward with the current approach, especially considering the redirects that have already been put in place (which I don't have the ability to remove anyway).

krzysdz: > Okay, quite a bit of time has passed. Other than the `.html` extensions, have we run into any additional issues?

I don't think so. The thing that I found after creating this PR is that it just works with `--include-fragments` in lychee, which I think that we should enable. Fragment checking should also work with the current directory based version with more [remapping](https://lychee.cli.rs/recipes/local-folder/).

> To be honest, I feel like we should move forward with the current approach, especially considering the redirects that have already been put in place (which I don't have the ability to remove anyway).

The file format works **with and without** redirects (`/page` and `/page.html` just work, `/page/` receives a 301 redirect to `/page`).

I initially wrote that this also lets us avoid 301s that add `/`, but while most links in menu don't have a trailing `/` ([`/api/express/`](https://github.com/expressjs/expressjs.com/blob/b3e09cd9abacf43dde376b637a769b53859db5ec/src/config/menu/api.ts#L21) is the only one ending with `/`), links in content now [have `/` appended](https://github.com/expressjs/expressjs.com/blob/b3e09cd9abacf43dde376b637a769b53859db5ec/src/plugins/rewrite-localized-links-core.mjs#L138-L152) since #2193 was merged (2 days ago).

bjohansebas:
Google has already adapted to the new format, so I think we can close this. Thanks anyway!

PR #2358, opened 2026-05-29
chore(5x): remove deprecated --save flag from install examples

ARYAN0-work:
I removed the --save flag from the npm install express command on the homepage and added a note explaining the default dependency behavior for npm 5+. I also updated the related tests.

I did this because while exploring the site, I noticed npm install express was written in three different places with different commands. This can be confusing for users wondering what the difference is and which one is correct. I updated the commands to be consistent and added a note to make the behavior clear.

### DIFF
--- a/src/content/docs/en/5x/starter/installing.mdx
+++ b/src/content/docs/en/5x/starter/installing.mdx
@@ -39,9 +39,3 @@ To install Express temporarily and not add it to the dependencies list:
 ```bash
 npm install express --no-save
 ```
-
-<Alert type="info">
-  For npm version 5.0.0 and above, <code>npm install express</code> automatically adds the package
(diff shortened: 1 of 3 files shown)


src/content/docs/en/5x/starter/installing.mdx
```diff
@@ -39,3 +39,9 @@ To install Express temporarily and not add it to the dependencies list:
 ```bash
 npm install express --no-save
 ```
+
+<Alert type="info">
+  For npm version 5.0.0 and above, <code>npm install express</code> automatically adds the package
+  to your <code>dependencies</code>. Explicit use of the <code>--save</code> flag is optional and
+  mainly kept for reference purposes.
+</Alert>
```
bjohansebas:
```suggestion
```

bjohansebas (approved):

PR #2227, opened 2026-03-30
docs: add pompelmi to additional middleware modules

SonoTommy:
## Summary

This PR adds [pompelmi](https://github.com/pompelmi/pompelmi) to the “Additional middleware modules” section in `en/resources/middleware.md`.

Pompelmi is an open-source file upload security toolkit for Node.js with Express support. It helps inspect uploaded files before storage with checks such as malware scanning hooks, MIME spoofing detection, and risky archive inspection.

## Why this fits this page

- It is directly relevant to Express middleware usage around `multipart/form-data` and file uploads.
- It addresses a practical security concern at the middleware layer: validating and scanning untrusted uploads before they are stored or processed.
- The project already documents Express usage and provides an Express adapter for this workflow.

## Scope

- docs only
- one new entry in the third-party middleware list
- no changes to Express core or official Express-maintained middleware

Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.

PR #2375, opened 2026-06-08
docs(middleware): add @inferdi/express

maxrendel:
## Summary

- Adds `@inferdi/express` to the Additional middleware modules list.
- `@inferdi/express` provides request-scoped dependency injection for Express 5 using explicit InferDI scopes.

## Test plan

- [x] Updated only `src/content/pages/en/resources/middleware/index.mdx`.
- [x] Did not edit localized pages directly; translations are handled through Crowdin.
- [x] Link points to the package documentation in the InferDI monorepo.

### DIFF
--- a/src/content/pages/en/resources/middleware/index.mdx
+++ b/src/content/pages/en/resources/middleware/index.mdx
@@ -40,4 +40,5 @@ These are some additional popular middleware modules.
 | Middleware module                                   | Description                                                                                                                                  |
 | --------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------- |
 | [helmet](https://github.com/helmetjs/helmet)        | Helps secure your apps by setting various HTTP headers.                                                                                      |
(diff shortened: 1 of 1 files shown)


krzysdz:
Right now we don't have any criteria which middleware is listed there, but the section description talks about _popular middleware_. I don't think that a  project with 35 stars, 43 commits and a history of just 1 month can be considered popular. The Express version of your package has only 1 version (from an hour ago) and 0 installs.

See also https://github.com/expressjs/expressjs.com/pull/2227#pullrequestreview-4031120692 and https://github.com/expressjs/expressjs.com/pull/1516#issuecomment-2108907327 for explanations why we don't accept such PRs at this moment.

maxrendel:
@krzysdz thanks for the explanation. I understand the concern about popularity, maintenance burden, and the lack of clear inclusion criteria for third-party middleware.

Just to clarify: is the current policy that no new third-party middleware entries should be added until the middleware section is redesigned or explicit criteria are defined?

If so, I’m fine with keeping this PR closed and revisiting later when `@inferdi/express` has more adoption and the project has clearer requirements for this page.

Thanks for reviewing.

krzysdz:
> Just to clarify: is the current policy that no new third-party middleware entries should be added until the middleware section is redesigned or explicit criteria are defined?

Yes. Until we have explicit criteria there should be no new entries of the list.

PR #2395, opened 2026-06-18
docs: add Deno to the package manager tabs

bartlomieju:
👋 Bartek from the Deno team here.

This adds **Deno** as a fifth option to the `<PackageManagerCommand>` tabs, alongside npm / yarn / pnpm / bun. Since the equivalents are derived at build time from the authored npm command (via `convertPackageManagerCommand`), Deno now appears on every install snippet across the site and in every language, with no per-page changes.

Mappings added (Deno is a drop-in package manager for the npm ecosystem):

- `npm install <pkg>` → `deno add <pkg>` (e.g. `deno add express`)
- `npm install` (no packages) → `deno install`
- `npm uninstall <pkg>` → `deno remove <pkg>`
- `npm init` → `deno init`
- `npm create <starter>` → `deno create --npm <starter>`
- `npm run <script>` → `deno task <script>`
- `npm exec <pkg>` / `npx <pkg>` → `deno x <pkg>`

Following the existing pattern, commands with no clean Deno equivalent return `null`, so the Deno tab is simply omitted there (e.g. `npm install <pkg> --no-save`, global installs).

I verified locally with `astro dev`: the install page renders a Deno tab showing `deno add express` (and `deno init` for `npm init`), the `--no-save` snippet correctly omits Deno, and `prettier --check` passes. Every Deno command was checked against the Deno 2.8 CLI.

Two files changed: the converter (`src/utils/package-manager.ts`) and the `MANAGERS` list in the component's tab-switching script.

Totally fine to close this if you'd prefer not to add another tab. Happy to adjust any of the mappings.

_Disclosure: I work on Deno, so I have an interest here. The goal is parity with the package managers you already list, not promotion. This PR was prepared with AI assistance under human supervision: I review every change myself and personally respond to any comments or review feedback._


src/utils/package-manager.ts
```diff
@@ -102,6 +106,9 @@ export function convertPackageManagerCommand(command: string): CommandMap {
           ? null
           : join('pnpm add', isDev && '--save-dev', isGlobal && '--global', pkgs, extras),
         bun: join('bun add', isDev && '--dev', isGlobal && '--global', pkgs, extras),
+        // Deno adds to deno.json/package.json; it has no `--no-save`, and global
+        // tool installs use a different verb (`deno install -g`), so skip those.
+        deno: hasNoSave || isGlobal ? null : join('deno add', isDev && '--dev', pkgs, extras),
```
krzysdz: If deno has no `--no-save` then why do you return `true` instead of `null` (like you wrote in the description)[^1]? For consistency it would be better to change this to chained ternary operators to match the structure used for other package managers (it would also fix the problem described in the previous sentence).

According to [your docs](https://docs.deno.com/runtime/reference/cli/add/) `deno add` is an alias to `deno install` - using `deno install` would allow you to use this way of "assembling" the command using `join` to include global installs. And nothing prevents you from using `deno install` when `isGlobal` is true and `deno add` when it's false.

[^1]: LLMs generating code frequently write one thing in description and something completely else in code.

krzysdz (changes requested):

src/utils/package-manager.ts
bartlomieju: Thanks for the review, I did another pass on this one and corrected the mistake here and in `deno x` invocation. I also removed the comment here since it's not really needed.

src/utils/package-manager.ts
```diff
@@ -117,6 +124,7 @@ export function convertPackageManagerCommand(command: string): CommandMap {
         yarn: isGlobal ? join('yarn global remove', pkgs) : join('yarn remove', pkgs),
         pnpm: join('pnpm remove', isGlobal && '--global', pkgs),
         bun: join('bun remove', isGlobal && '--global', pkgs),
+        deno: isGlobal ? null : join('deno remove', pkgs),
```
krzysdz: I'm pretty sure that Deno allows [uninstalling packages installed globally](https://docs.deno.com/runtime/reference/cli/uninstall/#deno-uninstall---global-%5Bscript_name%5D).

BTW Deno docs say that `add` is an alias of `install` and `remove` is an alias of `uninstall`, but aliases usually accept the same parameters.

krzysdz (changes requested):

bjohansebas (changes requested):
Does Deno now actually install the package into `node_modules` so it can be executed with Node? I'm asking because Bun does, and that's why it's listed there. I wasn't sure about Deno because I didn't know whether it does that now, and officially Express doesn't support it (even though we all know it runs fine).

src/utils/package-manager.ts
bartlomieju: Thanks for the catch, actually it looks like `deno remove` doesn't accept `--global`. I'll fix it for Deno v2.9 and can open a follow up PR once it lands if that works for you.

bartlomieju:
> Does Deno now actually install the package into `node_modules` so it can be executed with Node? I'm asking because Bun does, and that's why it's listed there. I wasn't sure about Deno because I didn't know whether it does that now, and officially Express doesn't support it (even though we all know it runs fine).

@bjohansebas yes! You can use `deno` as a standalone package manager, it can set up `node_modules/` that is compatible with npm output, it's pretty fast too: https://deno.com/blog/v2.8#deno-now-defaults-to-npm.

src/utils/package-manager.ts
bartlomieju: Good catch, thanks! Fixed — global removals now map to `deno uninstall --global <pkg>` (mirroring the `yarn global remove` branch) instead of being dropped. You're right that `remove`/`uninstall` are aliases, but only `uninstall` takes `--global`, so I used that for the global case.

krzysdz:
> it **can** set up `node_modules/` that is compatible with npm output

If I'm reading this correctly it **can, but doesn't** by default:

> By default, Deno instead resolves npm packages from a central global cache and does not create a node_modules directory.

https://docs.deno.com/runtime/fundamentals/node/#control-node_modules

bartlomieju:
> > it **can** set up `node_modules/` that is compatible with npm output
> 
> If I'm reading this correctly it **can, but doesn't** by default:
> 
> > By default, Deno instead resolves npm packages from a central global cache and does not create a node_modules directory.
> 
> https://docs.deno.com/runtime/fundamentals/node/#control-node_modules

Yes I wasn't precise. Currently `node_modules/` is set up automatically if you have `package.json` (which Deno understands and can add it), or if you use `deno.json` for dependencies then you need `nodeModulesDir: "auto"` (or `"manual"`). These options are just causing confusion and will be phased out in Deno 3 to always create `node_modules/` by default.

bjohansebas (approved):
LGTM. An Express project naturally has a `package.json`, and if that automatically installs `node_modules` and can be run with Node, I'm more than happy with that.

I'll wait a bit to see if @expressjs/express-tc @expressjs/docs-wg  has any thoughts on this, because right now I feel like I probably shouldn't have added Bun in the first place. That said, Bun is also a package manager, which is why I added it, and Deno is a package manager as well, so I feel comfortable adding it too.

krzysdz:
> An Express project naturally has a `package.json`, and if that automatically installs `node_modules` and can be run with Node, I'm more than happy with that.

If someone uses [`deno init`](https://docs.deno.com/runtime/reference/cli/init/) to initialise a project, then there will be no `package.json`, so no `node_modules/` and the project won't be able to run with Node. I don't see any option in [`deno init` docs](https://docs.deno.com/runtime/reference/cli/init/) that would create `package.json` instead of/in addition to `deno.json`.

It works as a package manager only in already existing projects and can't be used to initialise one.

bartlomieju:
Good distinction. The Deno tab here is the equivalent of the `npm install express` step, not `npm init` — and in a project that has a `package.json` (which any Express app does), `deno install express` writes to `package.json` and produces an npm-compatible `node_modules/`, so `node app.js` runs exactly like the npm/yarn/pnpm/bun tabs. You're right that `deno init` scaffolds a Deno-native project instead, but that's the init step, which this tab doesn't replace. Happy to add a one-line note that Deno users in a fresh dir should start from a `package.json` (or just `deno run` it) if you think that's clearer.

PR #2400, opened 2026-06-21
chore: add cspell for spell checking

bjohansebas:


src/layouts/Layout.astro
```diff
@@ -124,7 +124,7 @@ const lang = getLangFromUrl(Astro.url);
     <meta name="twitter:site" content="@UseExpressJS" />
     <meta name="twitter:title" content={pageTitle} />
     <meta name="twitter:description" content={description} />
-    <meta name="twitter:domail" content="expressjs.com" />
+    <meta name="twitter:domain" content="expressjs.com" />
```
bjohansebas, line 127: lol

project-words.txt
```diff
@@ -0,0 +1,144 @@
...
+Comeau's
+cookieparser
+cookiesession
+couchnode
+crandmck
+Davide
+dbuser
+Deutsch
+doctable
+dont
```
krzysdz: This doesn't look like something that should be ignored

bjohansebas: agree

project-words.txt
```diff
@@ -0,0 +1,144 @@
...
+Embrey
+Español
+expressjs
+expressjscom
+expressstaticroot
+extensionless
+extnames
+Filip
+Flexbox
+Fmysubdomain
```
krzysdz, line 39: I can't find `Fmysubdomain` anywhere

bjohansebas: https://github.com/expressjs/expressjs.com/blob/61ef8d165b069266c3a41fdfdc1b41152acfd546/src/content/api/4x/api/response/index.mdx?plain=1#L282

project-words.txt
```diff
@@ -0,0 +1,144 @@
...
+abbbbbcd
+abbbcd
+abbcd
+abxcd
+Almeida
+Angelini
+Arcd
+azser
+basicauth
+behaviour
```
krzysdz: Should we decide on the _version_ of English? I personally mostly use the British spelling (I'm European), but I think that the documentation mostly uses American English (USA).

project-words.txt
krzysdz: Oh, I had "match whole word" option enabled when searching.

project-words.txt
```diff
@@ -0,0 +1,144 @@
...
+Italiano
+javascripts
+johndoe
+jongleberry
+jshtml
+jshttp
+Kamil
+kawaii
+Keygrip
+Konichiwa
```
krzysdz: This is a typo. The spelling with [double N](https://en.wikipedia.org/wiki/Konnichiwa) is correct.

bjohansebas: good catch

project-words.txt
```diff
@@ -0,0 +1,144 @@
...
+Ooops
+openjs
+oracledb
+Orama
+OSSF
+Oulkar
+persica
+pillarjs
+plantae
+portno
```
krzysdz: This is the only occurrence of "portno" and should be changed to "port number"

https://github.com/expressjs/expressjs.com/blob/61ef8d165b069266c3a41fdfdc1b41152acfd546/src/content/api/3x/api/request/index.mdx?plain=1#L133

project-words.txt
bjohansebas: the docs are mostly in American English, so we should stick to that to avoid having to rewrite a lot of terms

project-words.txt
bjohansebas: fixed thanks!

PR #2414, opened 2026-06-28
docs: add app.query() method to API documentation

GroophyLifefor:
Add `app.query()` documentation to Application API

Documents the new `app.query()` route method for handling HTTP QUERY requests ([RFC 10008](https://www.rfc-editor.org/rfc/rfc10008.html)), a usage example, and an info alert noting the method was standardized on June 15, 2026 (Fresh Method).

### DIFF
--- a/src/content/api/5x/api/application/index.mdx
+++ b/src/content/api/5x/api/application/index.mdx
@@ -677,10 +677,14 @@ Express supports the following routing methods corresponding to the HTTP methods
       <code>copy</code>
     </li>
     <li>
-      <a href="#appdelete"><code>delete</code></a>
+      <a href="#appdelete">
+        <code>delete</code>
+      </a>
     </li>
     <li>
-      <a href="#appget"><code>get</code></a>
+      <a href="#appget">
+        <code>get</code>
+      </a>
     </li>
     <li>
       <code>head</code>
@@ -715,7 +719,9 @@ Express supports the following routing methods corresponding to the HTTP methods
       <code>patch</code>
     </li>
     <li>
-      <a href="#apppost"><code>post</code></a>
+      <a href="#apppost">
+        <code>post</code>
+      </a>
     </li>
   </ul>
   <ul>
@@ -723,10 +729,14 @@ Express supports the following routing methods corresponding to the HTTP methods
       <code>purge</code>
     </li>
     <li>
-      <a href="#appput"><code>put</code></a>
+      <a href="#appput">
+        <code>put</code>
+      </a>
     </li>
     <li>
-      <a href="#appquery"><code>query</code></a>
+      <a href="#appquery">
+        <code>query</code>
+      </a>
     </li>
     <li>
       <code>report</code>
@@ -966,7 +976,7 @@ app.put('/', (req, res) => {
 
 ### app.query()
 
-<Signature runtime={{ "Node.js": ">=20.19.3 <21 || >=22.2.0" }}>
+<Signature runtime={{ 'Node.js': '>=20.19.3 <21 || >=22.2.0' }}>
   <Fragment slot="attributes">
     <Param name="path" type="String | RegExp | Array">
       The path for which the middleware function is invoked; can be any of: a string representing a
--- a/src/content/api/5x/api/application/index.mdx
+++ b/src/content/api/5x/api/application/index.mdx
@@ -668,95 +668,33 @@ PUT, POST, and so on, in lowercase. Thus, the actual methods are `app.get()`,
 
 Express supports the following routing methods corresponding to the HTTP methods of the same names:
 
-<div class="methods-columns">
-  <ul>
-    <li>
-      <code>checkout</code>
-    </li>
-    <li>
-      <code>copy</code>
-    </li>
-    <li>
-      <a href="#appdelete">
-        <code>delete</code>
-      </a>
-    </li>
-    <li>
-      <a href="#appget">
-        <code>get</code>
-      </a>
-    </li>
-    <li>
-      <code>head</code>
-    </li>
-    <li>
-      <code>lock</code>
-    </li>
-    <li>
-      <code>merge</code>
-    </li>
-    <li>
-      <code>mkactivity</code>
-    </li>
-  </ul>
-  <ul>
-    <li>
-      <code>mkcol</code>
-    </li>
-    <li>
-      <code>move</code>
-    </li>
-    <li>
-      <code>m-search</code>
-    </li>
-    <li>
-      <code>notify</code>
-    </li>
-    <li>
-      <code>options</code>
-    </li>
-    <li>
-      <code>patch</code>
-    </li>
-    <li>
-      <a href="#apppost">
-        <code>post</code>
-      </a>
-    </li>
-  </ul>
-  <ul>
-    <li>
-      <code>purge</code>
-    </li>
-    <li>
-      <a href="#appput">
-        <code>put</code>
-      </a>
-    </li>
-    <li>
-      <a href="#appquery">
-        <code>query</code>
-      </a>
-    </li>
-    <li>
-      <code>report</code>
-    </li>
-    <li>
-      <code>search</code>
-    </li>
-    <li>
-      <code>subscribe</code>
-    </li>
-    <li>
-      <code>trace</code>
-    </li>
-    <li>
-      <code>unlock</code>
-    </li>
-    <li>
-      <code>unsubscribe</code>
-    </li>
-  </ul>
+<div class="columns-2">
+
+- `checkout`
+- `copy`
+- [`delete`](#appdelete)
+- [`get`](#appget)
+- `head`
+- `lock`
+- `merge`
+- `mkactivity`
+- `mkcol`
+- `move`
+- `m-search`
+- `notify`
+- `options`
+- `patch`
+- [`post`](#apppost)
+- `purge`
+- [`put`](#appput)
+- [`query`](#appquery)
+- `report`
+- `search`
+- `subscribe`
+- `trace`
+- `unlock`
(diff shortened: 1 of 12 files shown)


src/content/api/5x/api/application/index.mdx
```diff
@@ -961,6 +961,58 @@ app.put('/', (req, res) => {
...
+
+Routes HTTP QUERY requests to the specified path with the specified callback functions.
+For more information, see the [routing guide](/guide/routing/).
+
+<Alert type="alert">
+
+The HTTP QUERY method was standardized on June 15, 2026 as [RFC 10008](https://www.rfc-editor.org/rfc/rfc10008.html).
+Client and intermediary support is still limited — verify that your target environments support the QUERY verb before relying on it in production.
+
+</Alert>
```
bjohansebas, line 946: Also, include the Node.js version.

src/content/docs/zh-cn/5x/guide/routing.mdx
bjohansebas, line 1: Avoid updating guides in other languages.

GroophyLifefor: Ah, not to know that, fixing

src/content/api/5x/api/application/index.mdx
```diff
@@ -961,6 +961,58 @@ app.put('/', (req, res) => {
 });
 ```
 
+### app.query()
+
+<Signature>
```
bjohansebas: I've been thinking about adding a "since" field, but at the package level. It's difficult to say which Express version a feature was introduced in, so instead it would indicate the version of the specific package it was added in, or in cases like this, the Node.js version it requires.

GroophyLifefor: Since this feature mostly based on Node Version, I would prefer to see minimum node version.

bjohansebas: Yeah, but I mean it more as a general concept, since it won't always be tied to the Node.js version. Sometimes it will be the version of a package, like `router`, for example. But that's something for the docs team to decide. I should finish that part this week.

krzysdz (changes requested):
All supported methods are listed in `app.METHOD()` section and I can't see it modified here. If someone checked the docs for `router`, they would be directed to the [table in `app.METHOD()`](https://github.com/expressjs/expressjs.com/blob/472119366bbffc206e0deba32277d607cf9f6e13/src/content/api/5x/api/application/index.mdx?plain=1#L667) for the list of supported methods.

BTW. [`HEAD`](https://httpwg.org/specs/rfc9110.html#HEAD), [`OPTIONS`](https://httpwg.org/specs/rfc9110.html#OPTIONS) and [`PATCH`](https://www.rfc-editor.org/info/rfc5789/), which AFAIK also are a part of the HTTP standard do not have their own sections in `app` (like `DELETE`, `GET`, `POST`, `PUT` do)

src/content/api/5x/api/application/index.mdx
GroophyLifefor: image
image

from my research I found as 21.7.2 as minimum and tested with "var { METHODS } = require('node:http'); METHODS.includes('QUERY')" script.

GroophyLifefor:
> All supported methods are listed in `app.METHOD()` section and I can't see it modified here. If someone checked the docs for `router`, they would be directed to the [table in `app.METHOD()`](https://github.com/expressjs/expressjs.com/blob/472119366bbffc206e0deba32277d607cf9f6e13/src/content/api/5x/api/application/index.mdx?plain=1#L667) for the list of supported methods.
> 
> BTW. [`HEAD`](https://httpwg.org/specs/rfc9110.html#HEAD), [`OPTIONS`](https://httpwg.org/specs/rfc9110.html#OPTIONS) and [`PATCH`](https://www.rfc-editor.org/info/rfc5789/), which AFAIK also are a part of the HTTP standard do not have their own sections in `app` (like `DELETE`, `GET`, `POST`, `PUT` do)

I didn't add it there on purpose because I created a new section for `query`, if I hadn't created a section, I would have just added `query` to the list there.

bjohansebas:
I think having a separate section makes sense. I say that because it's essentially a combination of `POST` and `GET` in some ways, and this is also a case where we should educate developers. Many people don't know it exists, so it's a good addition for the query use case. It also helps distinguish it from Express 4, which, as far as I know, doesn't support it.

src/content/api/5x/api/application/index.mdx
krzysdz: There are problems with 21.7.2 - https://github.com/nodejs/node/issues/51562#issuecomment-2067811981. It is properly supported since 22.2.0 (https://github.com/nodejs/node/issues/51562#issuecomment-2151103116), but the patch was not backported to v21 (https://github.com/nodejs/node/issues/51562#issuecomment-2151421058).

bjohansebas:
> I didn't add it there on purpose because I created a new section for query, if I hadn't created a section, I would have just added query to the list there.

Also, add it to that table

src/content/api/5x/api/application/index.mdx
bjohansebas: yep

krzysdz:
> It also helps distinguish it from Express 4, which, as far as I know, doesn't support it.

I think that it does. The support depends only on Node.js version if I remember this correctly.

**EDIT:** `QUERY` test changes in 4.x https://github.com/expressjs/express/pull/6513

bjohansebas:
> I think that it does. The support depends only on Node.js version if I remember this correctly

To be honest, I never tested it. But unless the methods package was updated, then we would have support

Lol, it's been too long. My memory does fail sometimes 😂

GroophyLifefor:
> > It also helps distinguish it from Express 4, which, as far as I know, doesn't support it.
> 
> I think that it does. The support depends only on Node.js version if I remember this correctly.
> 
> **EDIT:** `QUERY` test changes in 4.x [expressjs/express#6513](https://github.com/expressjs/express/pull/6513)

from my research, in v4, uses "methods" npm package [line 18](https://github.com/expressjs/express/blob/715101bd27ec9c14a7ccc8a37c476f330d98ad53/lib/application.js#L18), and list of that http methods not includes "query" [index.js](https://www.npmjs.com/package/methods?activeTab=code)

edit: at the end, v4 *should* not support it, but not tested
2n edit: tested with empty project and looks working

krzysdz:
> unless the methods package was updated, then we would have support

It uses whatever Node.js lists in [`http.METHODS`](https://nodejs.org/docs/latest/api/http.html#httpmethods) or if `http.METHODS` does not exist (pre 0.11.9) it defaults to a hardcoded list:

https://github.com/jshttp/methods/blob/f89e491ab3f2a1e8c418648b890da2e96ef15edb/index.js#L22-L33

bjohansebas:
In that case, it looks like we do have support in Express 4 after all. Would you like to update its documentation as well?

GroophyLifefor:
> In that case, it looks like we do have support in Express 4 after all. Would you like to update its documentation as well?

I think it would be best if you made the decision on this. If you decide, I can make the updates quickly.

bjohansebas:
>  I can make the updates quickly.

yes please

krzysdz (commented):
We should not mention Node.js versions other than 22.2+, because it just doesn't work correctly on older versions (see https://github.com/expressjs/expressjs.com/pull/2414/changes#r3488667921).

I think that the Node.js version requirement could be mentioned in the list (something like `<code>query</code> - requires Node.js v22.2.0 or higher`).

bjohansebas:
Yep, I'll make that change.

krzysdz:
Should we mention that a parser (e.g. `express.urlencoded()`) must be used to access `req.body` with the actual query content?

bjohansebas:
> Should we mention that a parser (e.g. express.urlencoded()) must be used to access req.body with the actual query content?

I'm not entirely sure how it works yet. I'm going to start testing it first, and then I'll make the necessary changes to get it working.

krzysdz:
> > Should we mention that a parser (e.g. express.urlencoded()) must be used to access req.body with the actual query content?
> 
> I'm not entirely sure how it works yet. I'm going to start testing it first, and then I'll make the necessary changes to get it working.

It behaves like `POST` (requires `Content-Type` and body), but can't modify anything and can be cached like `GET`.

krzysdz (changes requested):
v22.**2**.0, not v22.0.0 (https://github.com/nodejs/node/issues/51562#issuecomment-2151103116)

bjohansebas:
It was in Node.js 22.0.0.
https://github.com/nodejs/node/releases/tag/v22.0.0
https://github.com/nodejs/node/commit/25a6fb6a07b34487a2d8d55cfcac3f0af19b0e2c

krzysdz:
But this bugfix https://github.com/nodejs/node/pull/52701 landed in v22.2.0.

krzysdz:
The bugfix was backported to v20.19.3, but I don't know if it has other patches required for `QUERY` support.

**EDIT:** In v20.19.2 llhttp received an update to 9.2.0, so there's a chance that `QUERY` is supported in Node.js 20.x

krzysdz:
Just tested - Node.js v20.19.3 works and v20.19.2 doesn't

Simple `QUERY` method test passing on Node.js v20.19.3

krzysdz (approved):

bjohansebas (approved):

bjohansebas:
thanks @GroophyLifefor @krzysdz !

RobinTail:
Thank you @GroophyLifefor && @krzysdz 

The types are on the way: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/75187

CC @bjohansebas

PR #2415, opened 2026-06-30
fix: triager guide broken link

mkco0:


krzysdz:
This part of this file is [updated automatically](https://github.com/expressjs/expressjs.com/blob/main/.github/scripts/get-contributing.sh) from https://github.com/expressjs/discussions/blob/master/docs/GOVERNANCE.md.

https://github.com/expressjs/expressjs.com/blob/3295cd3ab33dfb5a395379d0ec84955ff71c790f/src/content/pages/en/resources/contributing.mdx?plain=1#L16

It looks like the script has problems with relative URLs. The URL was last updated in source in https://github.com/expressjs/discussions/pull/489, and the changes have been pulled automatically in https://github.com/expressjs/expressjs.com/pull/2315.

mkco0:
I see, thanks for the clarification about the [script](https://github.com/expressjs/expressjs.com/blob/main/.github/scripts/get-contributing.sh), I wasn't aware of that. I'll close my pull request and open an issue while trying to figure out what's wrong with the script.

krzysdz:
> I'll close my pull request and open an issue while trying to figure out what's wrong with the script.

Here is the part that modifies links. It adds always the same prefix, but URLs that do not start with `/` are relative to the directory of the source file 

https://github.com/expressjs/expressjs.com/blob/3295cd3ab33dfb5a395379d0ec84955ff71c790f/.github/scripts/get-contributing.sh#L82-L84

PR #2418, opened 2026-07-01
chore: add homepage and bugs URLs to package.json

DakshSinghDhami:
Adds homepage and bugs.url fields


krzysdz:
This is not an NPM package. Why would it need those fields, if you have to find the repository first to see them?

https://github.com/expressjs/expressjs.com/blob/9ac528e469b0e81cc1dc3bed031758c4ad746c2b/package.json#L4

PR #2430, opened 2026-07-04
fix(sitemap): update filter to include only canonical locale-prefixed URLs

bjohansebas:
…


bjohansebas:
We were including pages with redirects, which we shouldn't be.

PR #2432, opened 2026-07-05
fix(seo): declare canonical URLs for unversioned and untranslated page copies

bjohansebas:
Technically, these pages contain the same content, just for i18n purposes. Because of that, we keep a canonical pointing to the English page so Google can index it more effectively.

It would also be good to exclude them from the sitemap, but I'd consider that a future improvement since I haven't found a clean way to do it yet.

Also, the reason I chose the unversioned URLs to point to the latest version (v5) is that Google was already treating the v5 URLs as the canonical ones.

By the way, I have access to Google Search Console, so I can actually verify this.

### DIFF
--- a/src/layouts/Layout.astro
+++ b/src/layouts/Layout.astro
@@ -11,6 +11,8 @@ interface Props {
   description?: string;
   authors?: { name: string; github?: string }[];
   image?: string;
+  /** Canonical pathname override, for pages served as a copy of another URL. */
+  canonicalPathname?: string;
 }
 
 const {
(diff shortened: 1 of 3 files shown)


src/layouts/Layout.astro
```diff
@@ -90,18 +99,19 @@ const lang = getLangFromUrl(Astro.url);
...
-          href={
-            new URL(
-              replaceLanguageInPath(Astro.url.pathname, altLang),
-              Astro.site || Astro.url.origin
-            )
-          }
-        />
-      ))
+      !isBlogOrApi &&
+        ['x-default', ...languageCodes].map((altLang) => (
```
bjohansebas, line 103: This is so that if none of the available languages match the user's language, it falls back to English.

PR #2434, opened 2026-07-05
fix: serve homepage at root instead of meta-refresh redirect to /en/

bjohansebas:
I'm restoring the behavior from before the redesign by removing the automatic redirect to English. Google still recommends the root URL by default for obvious reasons, since it was the homepage for many years.

imagen

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -73,13 +73,15 @@ export default defineConfig({
     icon(),
     react(),
     sitemap({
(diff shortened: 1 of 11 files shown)


ShubhamOulkar:
I think we should fix this file too "src/pages/[lang]/[...slug].astro".

krzysdz:
When I write the URL, I never add `/en/` and I block refresh (`accessibility.blockautorefresh` in Firefox `about:config`), so having an actual home page at `/` would be nice.

PR #2435, opened 2026-07-05
feat: add signature metadata handling for improved documentation extractio

bjohansebas:
This would improve the inputs we provide to Orama and the llms.txt files.

The changes in Orama won't be visible until Orama recovers from its outage, since its portal isn't even accessible right now.

imagen

-->


bjohansebas (commented):
Run tests

PR #2442, opened 2026-07-11
fix: prevent chat source citation clipping

### LINKED ISSUE #2408: Display of citations in AI question responses is broken
See how the third link here is cropped on the right, and there's no way to scroll right to see if there are additional citations:

Image

vibhor-aggr:
## Summary

Closes #2408.

Prevents AI answer source/citation cards from being clipped horizontally by:
- allowing the source list to show a thin horizontal scrollbar instead of hiding the scroll affordance
- preventing individual source cards from shrinking in the flex row
- reserving right-side padding so the final card is not obscured by the fade overlay
- allowing long citation text or URLs to wrap inside the card

## Verification

- `npx prettier --check src/components/patterns/Searchbox/ChatSources.css`
- `npm run lint -- src/components/patterns/Searchbox/ChatSources.css` (exits 0; CSS is ignored by the current ESLint config)
- `git diff --check`
- `npm run build`

### DIFF
--- a/src/components/patterns/Searchbox/ChatSources.css
+++ b/src/components/patterns/Searchbox/ChatSources.css
@@ -29,21 +29,19 @@
     gap: var(--space-3);
     margin-top: var(--space-8);
     margin-bottom: 0;
-    overflow-x: scroll;
+    overflow-x: auto;
     scroll-behavior: smooth;
-    -ms-overflow-style: none;
-    scrollbar-width: none;
+    scrollbar-width: thin;
     padding-left: 0;
+    padding-right: var(--space-16);
     list-style: none;
-
-    &::-webkit-scrollbar {
-      display: none;
-    }
   }
 
   .chat-sources-item {
     display: flex;
+    flex: 0 0 auto;
     max-width: 100%;
+    min-width: 0;
     height: 100%;
     align-items: center;
     gap: var(--space-2);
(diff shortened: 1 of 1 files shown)


ShubhamOulkar (changes requested):
Hi @vibhor-aggr, thank you for bringing up and fixing this issue.

Please add the `::before` sudo element for `chat-sources-scroll-container` so that the gradient will appear on the left. @krzysdz pointed it out in the issue comment.

Adjust padding for `chat-sources` so that it will produce a non-overlapping scroller UI. As shown below.

https://github.com/user-attachments/assets/d61cc6a5-5f67-4b0c-8aad-308b16fdc524

You can refer to the Web Content Accessibility Guidelines at https://www.w3.org/WAI/WCAG22/Techniques/general/G206

vibhor-aggr:
Addressed in `5b63b8f3`:

- added a mirrored `::before` fade on the left edge
- reserved `var(--space-16)` padding on both inline edges so the first and last cards do not overlap either fade
- layered both fades consistently while keeping them non-interactive

Verification passed with Prettier, `git diff --check`, the full 2,307-page `npm run build`, and Playwright geometry checks at 800px and 375px widths. Both edge clearances match the 64px fade width.

ShubhamOulkar (approved):

ShubhamOulkar:
thank you @vibhor-aggr

PR #2451, opened 2026-07-18
docs(routing): clarify wildcard behavior and restructure route parameters

### LINKED ISSUE #1891: Missing documentation for wildcard parameters
### Description

I am unable to find any documentation indicating how wildcard parameters work in v5. The only mention I've been able to find is in the [v5 Migration guide](https://expressjs.com/en/guide/migrating-5.html#path-syntax) which states:

> The wildcard * must have a name, matching the behavior of parameters :, use /*splat instead of /*

The [Routing docs](https://expressjs.com/en/guide/routing.html) do not explain how this syntax works. The closest it says there is that "the characters ?, +, *, [], and () are handled differently than in version 4, please review the [migration guide](https://expressjs.com/en/guide/migrating-5.html#path-syntax) for more information." which only states what I put above.

The section dedicated to Route Parameters on the same page also does not explain that you can swap out the `:` with `*` to consume more of the path.

The [`req.params`](https://expressjs.com/en/5x/api.html#req.params) section of the docs also doesn't indicate that you even can use `*` params. It also does not indicate that these params will be an array of strings that were split on `/`

### Expectations

The path syntax documentation and any docs about routing and handling route parameters should indicate that wildcard parameters are an option and explain how they work and how someone should be using them.

Akshay4754:
## Summary

This PR improves the discoverability of wildcard parameters in the Express 5 **Route parameters** documentation.

Instead of duplicating the existing wildcard documentation, it adds a short note that points readers to the existing **Wildcards** section.

Fixes #1891

### DIFF
--- a/src/content/api/4x/api/request/index.mdx
+++ b/src/content/api/4x/api/request/index.mdx
@@ -399,12 +399,6 @@ app.use('/admin', function (req: Request, res: Response, next: NextFunction) {
 
 ### req.params
 
-<Alert type="alert">
-
-In Express 4, wildcards (`*`) are unnamed, whereas in Express 5, wildcards must be explicitly named (e.g., `/*splat`). Please check the [migration guide to 5.x](https://expressjs.com/en/guide/migrating-5/#path-route-matching-syntax) before upgrading your routes.
-
-</Alert>
-
 <Signature type="Object" />
 
 This property is an object containing properties mapped to the [named route "parameters"](/guide/routing/#route-parameters). For example, if you have the route `/user/:name`, then the "name" property is available as `req.params.name`. This object defaults to `{}`.
--- a/src/content/api/4x/api/request/index.mdx
+++ b/src/content/api/4x/api/request/index.mdx
@@ -401,7 +401,7 @@ app.use('/admin', function (req: Request, res: Response, next: NextFunction) {
 
 <Alert type="alert">
 
-Wildcard parameters (`*`) are introduced in Express 5. See the [migration guide to Express 5](https://expressjs.com/en/guide/migrating-5/#path-route-matching-syntax) for details.
+In Express 4, wildcard matches are exposed as unnamed parameters (for example, `req.params[0]`) and return a string. Express 5 introduces named wildcard parameters that are exposed as arrays. See the [migration guide](https://expressjs.com/en/guide/migrating-5/#path-route-matching-syntax) for details.
 
 </Alert>
 
--- a/src/content/api/4x/api/request/index.mdx
+++ b/src/content/api/4x/api/request/index.mdx
@@ -399,9 +399,9 @@ app.use('/admin', function (req: Request, res: Response, next: NextFunction) {
 
 ### req.params
 
-<Alert type="alert">
+<Alert type="info">
 
-In Express 4, wildcard matches are exposed as unnamed parameters (for example, `req.params[0]`) and return a string. Express 5 introduces named wildcard parameters that are exposed as arrays. See the [migration guide](https://expressjs.com/en/guide/migrating-5/#path-route-matching-syntax) for details.
+In Express 4, wildcards (`*`) are unnamed, whereas in Express 5, wildcards must be explicitly named (e.g., `/*splat`). Please check the [migration guide to 5.x](https://expressjs.com/en/guide/migrating-5/#path-route-matching-syntax) before upgrading your routes.
 
 </Alert>
 
--- a/src/content/api/4x/api/request/index.mdx
+++ b/src/content/api/4x/api/request/index.mdx
@@ -399,7 +399,7 @@ app.use('/admin', function (req: Request, res: Response, next: NextFunction) {
 
 ### req.params
 
-<Alert type="info">
+<Alert type="alert">
 
 In Express 4, wildcards (`*`) are unnamed, whereas in Express 5, wildcards must be explicitly named (e.g., `/*splat`). Please check the [migration guide to 5.x](https://expressjs.com/en/guide/migrating-5/#path-route-matching-syntax) before upgrading your routes.
 
--- a/src/content/api/4x/api/request/index.mdx
+++ b/src/content/api/4x/api/request/index.mdx
@@ -399,6 +399,12 @@ app.use('/admin', function (req: Request, res: Response, next: NextFunction) {
 
 ### req.params
 
+<Alert type="alert">
+
+Wildcard parameters (`*`) are introduced in Express 5. See the [migration guide to Express 5](https://expressjs.com/en/guide/migrating-5/#path-route-matching-syntax) for details.
+
+</Alert>
+
 <Signature type="Object" />
 
 This property is an object containing properties mapped to the [named route "parameters"](/guide/routing/#route-parameters). For example, if you have the route `/user/:name`, then the "name" property is available as `req.params.name`. This object defaults to `{}`.
--- a/src/content/docs/en/5x/guide/routing.mdx
+++ b/src/content/docs/en/5x/guide/routing.mdx
@@ -554,7 +554,7 @@ The methods on the response object (`res`) in the following table can send a res
 ## app.route()
 
 You can create chainable route handlers for a route path by using `app.route()`.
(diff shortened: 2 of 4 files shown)


ShubhamOulkar (approved):

src/content/docs/en/5x/guide/routing.mdx
```diff
@@ -275,6 +275,8 @@ app.get('/users/:userId/books/:bookId', (req, res) => {
 });
 ```
 
+Wildcard parameters (`*`) are also supported in Express 5. Unlike named parameters (`:`), they capture one or more path segments and expose them as an array in `req.params`. See the [Wildcards](#wildcards) section above for details and examples.
```
ShubhamOulkar:
```suggestion
<Alert type="info">
Wildcard parameters (`*`) are only supports in Express 5. Unlike named parameters (`:`), they capture one or more path segments and expose them as an array in `req.params`. See the [Wildcards](#wildcards) section above or [migration guide](https://expressjs.com/en/guide/migrating-5/#path-route-matching-syntax) for details and examples.
<Alert>
```

Add abouve note on page: https://expressjs.com/en/5x/api/request/#reqparams

Some extra, but not necessary. Add alert not on `4x` doc https://expressjs.com/en/4x/api/request/#reqparams
```
<Alert type="alert">
Wildcard parameters (`*`) are introduced in Express 5. Unlike named parameters (`:`), they capture one or more path segments and expose them as an array in `req.params`. See the [migration guide to 5x](https://expressjs.com/en/guide/migrating-5/#path-route-matching-syntax) for details.
<Alert>
```

ShubhamOulkar (changes requested):

src/content/docs/en/5x/guide/routing.mdx
Akshay4754: Thanks for the review!

I've addressed the requested changes by:
- Wrapping the wildcard parameter note in an `Alert` component.
- Adding the wildcard parameter note to the Express 5 `req.params` documentation with a link to the migration guide.
- Adding the suggested notice to the Express 4 `req.params` documentation.

Please let me know if there are any further changes or improvements you'd like me to make. Thanks!

krzysdz: I'm not sure if this wording is clear. In 4.x there are wildcard parameters, but with no explicit names. These can still be accessed from `req.params` (so I'd argue that they are parameters) under integer keys (in most cases there's just one unnamed parameter so - `req.params[0]`). In 4.x wildcards return strings instead of `string[]` like in 5.x.

Akshay4754: Thanks for pointing that out! That's a good clarification. I'll update the wording to distinguish the Express 4 behavior (`req.params[0]` returning a string) from the named wildcard parameters introduced in Express 5.

ShubhamOulkar: Yes, I think this is best fit.
```
<Alert type="info">
In Express 4, wildcards (`*`) are unnamed, whereas in Express 5, wildcards must be explicitly named (e.g., `/*splat`). Please check the [migration guide to 5.x](https://expressjs.com/en/guide/migrating-5/#path-route-matching-syntax) before upgrading your routes.
</Alert>
```

Akshay4754: Thanks! I've updated the wording accordingly.

src/content/api/4x/api/request/index.mdx
```diff
@@ -399,6 +399,12 @@ app.use('/admin', function (req: Request, res: Response, next: NextFunction) {
 
 ### req.params
 
+<Alert type="info">
```
ShubhamOulkar:
```suggestion
<Alert type="alert">
```
Putting an alert box is useful here.

Akshay4754: Done, thanks! I've applied the suggested change.

ShubhamOulkar (approved):

krzysdz (changes requested):
5.x docs (API and routing guide) still claim that only Express 5 supports wildcards.

src/content/docs/en/5x/guide/routing.mdx
Akshay4754: Kindly tell if anything else is required. Thanks

src/content/docs/en/5x/guide/routing.mdx
```diff
@@ -275,6 +275,12 @@ app.get('/users/:userId/books/:bookId', (req, res) => {
 });
 ```
 
+<Alert type="info">
+
+Wildcard parameters (`*`) are only supported in Express 5. Unlike named parameters (`:`), they capture one or more path segments and expose them as an array in `req.params`. See the [Wildcards](#wildcards) section above or the [migration guide](https://expressjs.com/en/guide/migrating-5/#path-route-matching-syntax) for details and examples.
```
bjohansebas: I don't think we need to say "Express 5" anymore. The guides are versioned, so that redundancy isn't necessary

bjohansebas (changes requested):
Hmm, I need to take a closer look at this. I feel like we're breaking the reading flow, so I'll briefly block it for now

src/content/docs/en/5x/guide/routing.mdx
Akshay4754: Thanks for the feedback! I've simplified the wording by removing the redundant "Express 5" reference from the versioned guides while keeping the migration guide reference.

Akshay4754:
Hi @bjohansebas  , @ShubhamOulkar ,I've addressed the requested feedback in the latest commits, including the reading-flow concerns and the Express 4 migration notes. Whenever you have time, I'd appreciate another look and suggest if any changes needed . Thanks!

krzysdz:
I was wondering when did the "Wildcards" section referenced in `<Alert>` appear, because I'm pretty sure it wasn't there when #1891 was opened. Had it been there at that time, the issue most likely would not have been created. To be clear, this doesn't mean that I'm opposed to this PR.

Information about wildcards was introduced to the 5.x routing guide in #2263.

bjohansebas:
I think reorganizing it the way I did improves the readability of the docs, as well as the format for LLMs. The idea is to focus the guides only on the current versions; for everything else, we already have the migration guides if users run into incompatibilities.

bjohansebas (approved):

src/content/docs/en/5x/guide/routing.mdx
```diff
@@ -318,6 +254,74 @@ See the [path route matching syntax](/guide/migrating-5#path-route-matching-synt
...
+  res.send(`File: ${req.params.filepath.join('/')}`);
+});
+```
+
+To also match the root path, wrap the wildcard in braces:
+
+```js
+// Matches / , /foo , /foo/bar , etc.
+app.get('/{*splat}', (req, res) => {
+  // GET / => req.params.splat = []
```
krzysdz: Is this correct? In this case there should be no `"splat"` in `req.params`. I tested it some time ago in #2090.

bjohansebas: yep, good catch

src/content/docs/en/5x/guide/routing.mdx
```diff
@@ -356,6 +356,8 @@ app.get('/order{/:id}', (req: Request, res: Response) => {
 });
 ```
 
+The examples above behave the same regardless of the [`strict routing` setting](/api/application/#application-settings). As with any other route, that setting only changes whether a trailing slash (such as `/order/`) is treated as a different path.
```
krzysdz: This is not really clear to me - is it about the _trailing slash (such as `/order/`)_ in path or the requested URL[^1]. It's confusing also because on one hand the examples show that **with or without** strict routing the trailing `/` (in path) makes a difference and on the other hand the second sentence of this paragraph says that it's what strict routing affects.

[^1]: I know how this works, but it's not explained well and confusing for anyone who does not already know it.

krzysdz (approved):

bjohansebas:
thanks @Akshay4754

Akshay4754:
Thanks! I appreciate the review and all the guidance.

ShubhamOulkar:
Great @Akshay4754 @bjohansebas @krzysdz 🫰

PR #2464, opened 2026-07-30
docs: add missing package.json "type": "module" step to TypeScript in…

jnzlab:
## Description

The TypeScript setup section provides a `tsconfig.json` using `"module": "nodenext"` and demonstrates ESM `import` syntax in `src/app.ts`, but doesn't mention that `package.json` also needs `"type": "module"` set. Without it, TypeScript treats `.ts` files as CommonJS by default and throws `ECMAScript imports and exports cannot be written in a CommonJS file under 'verbatimModuleSyntax'. (ts1295)` when following the guide as written. This adds the missing step so the example works out of the box.

## Changes

- Added a new step between the middleware-types warning and the `tsconfig.json` block explaining that `"module": "nodenext"` requires `package.json` to declare `"type": "module"`
- Included a `package.json` code snippet showing the required field
- Added a warning callout explaining the exact error users will hit if this step is skipped

## Type of change

- [x] Documentation fix/improvement

### DIFF
--- a/src/content/docs/en/5x/starter/installing.mdx
+++ b/src/content/docs/en/5x/starter/installing.mdx
@@ -52,6 +52,25 @@ for example `@types/cors` alongside `cors`.
 
 </Alert>
 
+Since the `tsconfig.json` below uses `"module": "nodenext"`, TypeScript determines whether each
+file is treated as an ES Module or CommonJS based on the `"type"` field in your `package.json`.
+Add `"type": "module"` to your `package.json` so that ES Module `import`/`export` syntax is
+recognized:
+
+```json title="package.json"
+{
+  "type": "module"
+}
+```
+
+<Alert type="warning">
+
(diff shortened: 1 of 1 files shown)


krzysdz:
There is also another problem - while Node.js [**can** detect](https://nodejs.org/docs/latest-v26.x/api/packages.html#syntax-detection) and [run](https://nodejs.org/docs/latest-v26.x/api/packages.html#determining-module-system) `.js`/[`.ts`](https://nodejs.org/docs/latest-v26.x/api/typescript.html#determining-module-system:~:text=%2Ets%20files%20will%20have%20their%20module%20system%20determined%20the%20same%20way%20as%20%2Ejs%20files) files as ES Modules, `npm init` sets `"type": "commonjs"` [**by default**](https://docs.npmjs.com/cli/v12/commands/npm-init#init-type), effectively blocking the detection mechanism.

PR #1474, opened 2024-03-20
feat: add docs about open redirect prevention

wesleytodd:
Adds details about open redirects and how to prevent them.

PR #2465, opened 2026-07-31
docs: fix ReferenceError in open-redirect example (Url -> URL)

ofri-peretz:
## Summary

The open-redirect validation example on *Production Best Practices: Security* calls `new Url(req.query.url)`, but the JavaScript global is `URL`. As written the constructor throws `ReferenceError: Url is not defined` on every request, the `catch` swallows it, and the handler responds `400 "Unsupported redirect"` for every URL — including valid ones — so the redirect never runs. Anyone copying the snippet gets a handler that silently rejects all redirects instead of validating them.

One-line fix: `Url` → `URL`.

The same snippet appears in the translated mirrors, but per `docs/i18n.md` I've only touched the English source so Crowdin can propagate it.

Verified locally: `npm run check` green, unit tests 52/52, and a full `astro build` succeeds with the corrected snippet in the built page.

### DIFF
--- a/src/content/pages/en/advanced/best-practice-security.mdx
+++ b/src/content/pages/en/advanced/best-practice-security.mdx
@@ -64,7 +64,7 @@ Here is an example of checking URLs before using `res.redirect` or `res.location
 ```js
 app.use((req, res) => {
   try {
-    if (new Url(req.query.url).host !== 'example.com') {
+    if (new URL(req.query.url).host !== 'example.com') {
       return res.status(400).end(`Unsupported redirect to host: ${req.query.url}`);
     }
   } catch (e) {


krzysdz (approved):
Oops. @marco-ippolito found this in https://github.com/expressjs/expressjs.com/pull/1474#discussion_r1532245354, the suggestion was applied in 401c6a6039a6c52d43f057c6443d78c39dcea0c4 and then before merging there was a [force push](https://github.com/expressjs/expressjs.com/pull/1474#event-12216707839), which reformatted the code and accidentally [reverted](https://github.com/expressjs/expressjs.com/compare/401c6a6039a6c52d43f057c6443d78c39dcea0c4..19e7de375ecbcbc8f427c7ad6b0e25b8b151e81d) this change.

ofri-peretz:
Thanks @krzysdz for digging up the history, and @bjohansebas for merging.

Worth recording that the credit for spotting it belongs to @marco-ippolito in #1474 — the fix was applied there and then lost to a force-push before merge. I only noticed the symptom. Glad it's back.

PR #2221, opened 2026-03-28
perf: optimize hero media using WebM/WebP and reduce asset size

ShubhamOulkar:
Refactor hero background video logic to improve reliability, theme switching, and performance.

### Changes
- Moved video visibility control from CSS to JS to avoid dual-video rendering issues
- Fixed poster flickering and incorrect poster persistence during theme changes
- Removed `display: none` usage for videos/posters and switched to `opacity` for smoother transitions and to prevent flicker
- Poster is now hidden only after video is ready using `readyState` (`HAVE_FUTURE_DATA`)
- Handled already-loaded videos to avoid missing `canplay` events on theme switch
- Simplified logic using `getAssets()` to remove duplication
- Removed `video.src` checks (unreliable) in favor of `readyState` 
- Removed mobile blocking logic (`getComputedStyle` hack) 🚀🚀🚀
- Enabled background video animation on mobile 🥳🥳🥳
- Maintains accessibility by respecting `prefers-reduced-motion`

### Result
- Correct poster/video sync across theme changes
- Smooth poster to video transition using opacity
- Stable playback even after multiple theme toggles
- **Improved mobile support and cleaner logic** 💥

Asset | Before | After | Reduction
-- | -- | -- | --
Video (MP4 → WebM) | 651 KB | 224 KB | ~65%
Image (JPG → WebP) | 93 KB | 57 KB | ~40%

PR #2485, opened 2026-08-08
chore: rewrite HTML tables in `migrating-4.mdx` to markdown

krzysdz:
Tables in `migrating-4.mdx` are simple an have been rewritten using markdown syntax, which makes them work better with plugins (the HTML was missing `<thead>` and `<tbody>`).

The tables in "Express behind proxies" are quite large and contain a lot embedded markdown (code blocks, lists). Unfortunately, as far as I know it is not possible to write a markdown table with lists or code blocks without using HTML for either the table or the content.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12612,9 +12612,9 @@
       "license": "MIT"
     },
     "node_modules/undici": {
-      "version": "7.28.0",
-      "resolved": "https://registry.npmjs.org/undici/-/undici-7.28.0.tgz",
-      "integrity": "sha512-cRZYrTDwWznlnRiPjggAGxZXanty6M8RV1ff8Wm4LWXBp7/IG8v5DnOm74DtUBp9OONpK75YlPnIjQqX0dBDtA==",
+      "version": "7.29.0",
+      "resolved": "https://registry.npmjs.org/undici/-/undici-7.29.0.tgz",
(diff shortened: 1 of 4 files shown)


krzysdz:
One thing that I noticed while making these changes: #2221 removed a table from [5.x writing middleware](https://github.com/expressjs/expressjs.com/blob/b9d510698d15f3fd21f4a13fc97fe9f292f9b105/src/content/docs/en/5x/guide/writing-middleware.mdx?plain=1#L22) docs, but [left it in 4.x](https://github.com/expressjs/expressjs.com/blob/b9d510698d15f3fd21f4a13fc97fe9f292f9b105/src/content/docs/en/4x/guide/writing-middleware.mdx?plain=1#L22-L49). The table served as an explanation of some things shown on an illustration, but the illustration has been updated to include these descriptions, so 4.x version is [labelled twice](https://expressjs.com/en/4x/guide/writing-middleware/#mw-fig).

obraz

ShubhamOulkar (approved):
#2493 fixes scrolling issues, also we discussed on rewriting to markdown tables. migrating-4 tables are in markdown, if you want then do it for tables on behind-proxies page. Removing manual addition of `table-scroller` is essential before merging the PR.

krzysdz:
Removed the manually added `.table-scroller`. Tables from `behind-proxies.mdx` cannot be rewritten to pure markdown.

I looked for inline HTML and other than this (cannot be removed), some other HTML that can't be removed (e.g. links wrapping Astro components) it can be found only in:

- 3.x docs
- an unordered list in behind-proxies (4.x and 5.x, see krzysdz@0afd2707168feb00583385b4945f8a29cfe02690)
- one blog post (see https://github.com/expressjs/expressjs.com/pull/2493#issuecomment-5511478870)

PR #2488, opened 2026-08-12
fix: let the doc bottom navigation wrap on narrow viewports

### LINKED ISSUE #2486: Previous/next navigation can be too wide on small screens
Example on https://expressjs.com/en/guide/migrating-4/ (with #2485 ([preview](https://deploy-preview-2485--expressjscom-preview.netlify.app/en/guide/migrating-4/)), because without it tables cause the same problem):

Screenshot of a mobile preview of "Migrating to Express 4" guide. Previous and Next buttons in bottom navigation both have long descriptions and are placed on the same line. They are too wide, causing a gap on the right in remaining sections on the website.

### Possible solutions

1. `flex-wrap: wrap-reverse` in [`.doc-nav`](https://github.com/expressjs/expressjs.com/blob/61e6f789388cf1f56476e9c05550451bcf3347c4/src/components/patterns/DocBottomNav/DocBottomNav.css#L2-L9) - I think that "Next" being above "Previous" should make the navigation easier (most of the time people are reading from start to the end)
   Same preview as earlier, but now the Next button shows above the Previous button and both have arrows, which were not visible earlier. The page is sized correctly.
2. `flex-wrap: wrap` in [`.doc-nav`](https://github.com/expressjs/expressjs.com/blob/61e6f789388cf1f56476e9c05550451bcf3347c4/src/components/patterns/DocBottomNav/DocBottomNav.css#L2-L9)
   Screenshot similar to the one above (point 1), but the Previous button is above Next.

bilashcse:
## Problem

On narrow viewports the previous/next links at the bottom of a doc page cannot fit on one line, and because `.doc-nav` is `display: flex` with no `flex-wrap`, the row overflows instead of wrapping.

Measured on the live site at a 400px wide viewport on `/en/guide/migrating-4/`: `.doc-nav` is 340px wide with a `scrollWidth` of 399px, so the *Next* link is clipped. The same thing happens around 950px, where both sidebars are visible and the content column is only about 250px wide.

Closes #2486.

## Fix

Add `flex-wrap: wrap-reverse` to `.doc-nav`, the first option suggested in the issue. When the two links cannot share a line, *Next* wraps onto its own line above *Previous*, which keeps the forward link closest to the end of the content.

With the rule applied, the overflow measured above goes from 59px to 0px. At a 1314px viewport nothing changes: both links stay on one line at the same offsets.

## Tests

`tests/e2e/doc-bottom-nav.spec.ts` covers three things on `/en/guide/migrating-4/`: `.doc-nav` does not overflow at a 400px viewport, *Next* sits above *Previous* at that width, and both links stay side by side on one line at 1400px. The first two assertions fail against `main` today.

A note on verification: the Playwright job runs against the PR deploy preview, so I could not run the new spec from my fork. The numbers above come from measuring the elements in a browser on the live site with the rule injected, and the CI run on this PR should exercise the spec for real.

## Docs

`docs/design-system.md` gains a short *Doc Bottom Navigation* section describing the wrapping behaviour, including a note that DOM order stays previous-then-next so keyboard and screen reader order is unaffected.

### DIFF
--- a/tests/e2e/doc-bottom-nav.spec.ts
+++ b/tests/e2e/doc-bottom-nav.spec.ts
@@ -1,0 +1,54 @@
+import { test, expect } from '@playwright/test';
+
+// Regression coverage for https://github.com/expressjs/expressjs.com/issues/2486
+const DOC_PATH = '/en/guide/migrating-4/';
+
+test.describe('Doc bottom navigation', () => {
+  test('should not overflow horizontally on a narrow viewport', async ({ page }) => {
+    await page.setViewportSize({ width: 400, height: 900 });
+    await page.goto(DOC_PATH);
+
+    const nav = page.locator('.doc-nav');
+    await expect(nav).toBeVisible();
+
+    const overflow = await nav.evaluate((el) => el.scrollWidth - el.clientWidth);
+    expect(overflow).toBeLessThanOrEqual(1);
+  });
+
+  test('should move the next link above the previous link when they cannot share a line', async ({
+    page,
+  }) => {
+    await page.setViewportSize({ width: 400, height: 900 });
+    await page.goto(DOC_PATH);
+
+    const prev = page.locator('.doc-nav__link--prev');
+    const next = page.locator('.doc-nav__link--next');
+
+    await expect(prev).toBeVisible();
+    await expect(next).toBeVisible();
+
+    const prevBox = await prev.boundingBox();
+    const nextBox = await next.boundingBox();
+
+    if (!prevBox || !nextBox) {
+      throw new Error('expected both doc nav links to be laid out');
+    }
+
+    expect(nextBox.y + nextBox.height).toBeLessThanOrEqual(prevBox.y);
+  });
+
+  test('should keep both links on a single line on a wide viewport', async ({ page }) => {
+    await page.setViewportSize({ width: 1400, height: 900 });
+    await page.goto(DOC_PATH);
+
+    const prevBox = await page.locator('.doc-nav__link--prev').boundingBox();
+    const nextBox = await page.locator('.doc-nav__link--next').boundingBox();
+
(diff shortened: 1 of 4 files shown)


bilashcse:
CI results are in, so the verification note in the description can be treated as resolved: Playwright now runs 57 tests (48 before this PR, plus the 3 new ones on chromium, firefox and webkit) and all 57 pass against the deploy preview. `lint` and `build` are green too.

The extra commits after the first three are only lint fixes: a trailing blank line prettier did not want, and two rewordings because cspell's dictionary does not have "viewports". Happy to squash or to add the word to `project-words.txt` instead if you would prefer that.

tests/e2e/doc-bottom-nav.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+    await page.goto(DOC_PATH);
+
+    const nav = page.locator('.doc-nav');
+    await expect(nav).toBeVisible();
+
+    const overflow = await nav.evaluate((el) => el.scrollWidth - el.clientWidth);
+    expect(overflow).toBeLessThanOrEqual(1);
+  });
+
+  test('should move the next link above the previous link when they cannot share a line', async ({
```
ShubhamOulkar, line 18: I don’t think these tests are particularly useful, as we generally don’t test layout behaviours such as “the next link should move above the previous link when they can’t fit on the same line” or “the content should not overflow horizontally on a narrow viewport.”

These are more like responsive/layout implementation details rather than behaviours we need to explicitly test. I’d suggest removing these tests to keep the test suite focused on meaningful and maintainable behaviours.

ShubhamOulkar (changes requested):
Hi @bilashcse, thank you for opening PR.  please remove unnecessary changes that are not related to CSS issue.

Issue #1881, opened 2025-04-22
5.x docs are missing some documentation added in late 4.x versions


krzysdz:
@paulhiggs in https://github.com/expressjs/express/issues/6479 reported that `express.text()` is missing from 5.x docs.

`express.text()` and `express.raw()` were added in Express 4.17.0. The 4.x documentation reflects that (18dfac1bbd053fb99cba9ff137cb5a5e0c67d5fe, 958e581dbea384dca6bf4a62020762bb0f1ab1dc, b49ab1b94a34e37c791ec661b66ffac5a37803a5), but 5.x documentation has not been updated after changes from 4.17.0/4.17.1 were merged into 5.0/5.x branches (https://github.com/expressjs/express/commit/121fe9982b62d2b1760ec9cb148c19bc5f17255b).

TODO:

- [x] Add `express.text()` to 5.x API docs (#1883)
- [x] Add `express.raw()` to 5.x API docs (#1883)
- [x] Review 4.x documentation changes and check if more changes to late 4.x versions that were later included in 5.x are missing in 5.x docs (it has already been tracked in #1479)
  - [x] 4.18 added `priority` option to `res.cookie()` and `root` option to `res.download()` (5c98ee4e949fc69b9dbc89b839c137b7a2eea3fc) (#1886)
  - [x] #1402 improved `req.acceptsLanguages()` docs, but only the 4.x version (#1886)
  - [x] 5e918ea3a98edcc9a4f9698ca0994f3c6fc1f4e3 added warning boxes to `res.render()` and `app.render()` (#1886)
  - [x] fcaca7fcc6f182a8821352a94a88f9fc10320cbb added warnings about `locals` (#1886)
  - [x] #1456 documented the `partitioned` option of `res.cookie()` (#1886)
  - [x] 4.x and 5.x had slightly different (and incompatible) examples for `router.param(fn)`, which doesn't even exist in Express 5 (#1884)
  - [x] a5ca5b0edf437c066f7a1dfa5dc11bc4cb9981e1 changed `req.body` documentation (4.x), removing references to `multer` and `body-parser` and presenting built-in middleware. (#1886)
  - [x] #1063 expanded `res.cookie()` docs, showing how to set multiple cookies (#1886)

Other differences include:

[...]

PR #2489, opened 2026-08-12
docs: finish the remaining 5.x/4.x API reference consistency items

bilashcse:
Closes the three remaining unchecked items in #1881.

## Problem

The 4.x and 5.x API reference pages drifted apart after the redesign, which makes it hard to tell a
real API difference from a formatting difference when switching versions:

1. Result comments in code samples use double quotes in many places (`// => "text/html"`) while the
   rest of the site, and the Prettier `singleQuote` setting, use single quotes.
2. The `res.type()` sample had a typo — `res.type('png'); // => image/png:` — with the value unquoted
   and a stray trailing colon (the typo #1881 links to).
3. `req.host` and `req.hostname` were sorted differently: 5.x lists `req.host` first (alphabetical,
   like every other section), 4.x lists `req.hostname` first.
4. The 5.x page `description` frontmatter said "The request object …" / "The response object …" while
   4.x and the opening sentence of both pages say "The `req` object …" / "The `res` object …".

## Fix

- Normalised 33 result comments across the 4.x and 5.x application/request/response pages to single
  quotes. Double quotes are kept where the sample legitimately shows JSON (`// => callback({ "user": "tobi" })`)
  or a raw HTTP header (`// Host: "example.com:3000"`).
- Fixed the `res.type('png')` sample to `res.type('png'); // => 'image/png'`.
- Moved the 4.x `req.host` section above `req.hostname` so both versions are alphabetical. I picked
  the 5.x order because every other section on the page is alphabetical — happy to flip it the other
  way if you prefer the deprecated alias to stay last.
- Aligned the 5.x request/response `description` frontmatter with 4.x and with the first sentence of
  each page.

No prose or behaviour was changed, only formatting, ordering and the one typo.

## Tests

Added `tests/unit/api-docs-style.test.mjs`, which runs in the existing `npm run test:unit` step of
the `lint` job. It walks every `.mdx` file under `src/content/api/4x` and `src/content/api/5x` and
asserts that:

- no result comment quotes a plain string value with double quotes;
- no result comment ends with stray punctuation (this is what would have caught the `image/png:` typo);
- `req.host` is documented before `req.hostname` in both versions;
- the request/response `description` frontmatter starts with "The req object represents" /
  "The res object represents".

Each assertion fails on `main` before this change and passes after it.

## Docs

Added a "Style Conventions" subsection to `docs/content.md` under **API Reference** describing the
three conventions and pointing at the test that enforces them, so the next person editing these pages
knows the rules without having to diff 4.x against 5.x.

### DIFF
--- a/docs/content.md
+++ b/docs/content.md
@@ -134,6 +134,20 @@ src/content/api/
 - Versioned: `/en/5x/api/application` → Express 5.x application API
 - Non-versioned: `/en/api/application` → Defaults to Express 5.x
 
+### Style Conventions
+
+The reference pages for every version follow the same conventions, so that switching between
+Express 4.x and 5.x only shows real API differences:
+
+- Property and method sections are listed alphabetically, so `req.host` comes before `req.hostname`.
+- Result comments in code samples use single quotes, matching the Prettier `singleQuote` setting:
+  `res.type('png'); // => 'image/png'`. Double quotes are kept only where the sample shows JSON or a
+  raw HTTP header.
+- The frontmatter `description` of the request and response pages repeats the opening sentence of the
+  page, which refers to the objects as `req` and `res` rather than spelling out their full names.
+
+`tests/unit/api-docs-style.test.mjs` checks these conventions, and it runs as part of `npm run test:unit`.
+
 ## Versioning
 
 The Express.js documentation supports multiple versions. Content is organized by version in both the `docs` and `api` collections.
--- a/src/content/api/4x/api/application/index.mdx
+++ b/src/content/api/4x/api/application/index.mdx
@@ -560,7 +560,7 @@ app.get('title');
 
 app.set('title', 'My Site');
 app.get('title');
-// => "My Site"
+// => 'My Site'
 ```
 
 ### app.get()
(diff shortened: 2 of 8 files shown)


bilashcse:
Heads up on overlap: #2487 is also open against #1881 and touches two of the same files.

Where we agree, the changes are identical (the `res.type('png')` typo, and the 5.x request/response `description` frontmatter switching to "The req object ..." / "The res object ...").

Where we differ is `req.host` / `req.hostname`. #2487 makes 5.x match 4.x: it renames the 5.x `req.host` section to `req.hostname`, drops the `example.com:3000` and `[::1]:3000` samples, and reports `'example.com'`. That would undo a real Express 5 behaviour change, since in 5.x `req.host` keeps the port and `req.hostname` is the one that strips it. This PR instead keeps both 5.x sections as they are and moves the 4.x `req.host` section above `req.hostname`, so the ordering is alphabetical in both versions without touching what either property is documented to return.

This PR also covers the first checkbox in #1881, which #2487 does not: 33 result comments across the 4.x and 5.x application/request/response pages move from double to single quotes, plus the unit test and the docs/content.md section.

Happy to defer or rebase on whichever you would rather merge first.

PR #2493, opened 2026-08-30
Fix rehype accessible table plugins

### LINKED ISSUE #2484: A page in the documentation seems messy in small screens
I was surfing expressJS documentation with my phone, found this link https://expressjs.com/en/5x/guide/behind-proxies/ a bit untidy. This happens in a width < 300px.

It seems that example section takes full width, in contrast to the rest of content which is taking only one column (50% of width) leaving half of the page in a gap

Image

ShubhamOulkar:
This plugin was failing because it only handled standard Hypertext Abstract Syntax Tree format table nodes and missed tables authored as raw HTML inside MDX/Markdown (mdxJsxFlowElement, mdxJsxTextElement).

The fix registers the plugin in both Markdown and MDX pipelines and updates the transform to support both raw HTML and standard table nodes, ensuring consistent scrollable table behavior.

This fixes the missing wrapper for raw HTML tables: https://deploy-preview-2493--expressjscom-preview.netlify.app/en/5x/guide/behind-proxies/

and markdown-generated tables: https://deploy-preview-2493--expressjscom-preview.netlify.app/en/resources/middleware/

closes https://github.com/expressjs/expressjs.com/issues/2484


astro.config.mjs
```diff
@@ -68,8 +69,7 @@ export default defineConfig({
         borderWidth: 'var(--border-width-1)',
       },
     }),
-    accessibleTablesIntegration(),
-    mdx(),
+    mdx({ rehypePlugins: [rehypeAccessibleTables] }),
```
krzysdz: Something in this PR breaks syntax highlighting and adding `<a>` to headers (all plugins?) and I think this may be the culprit.

krzysdz (changes requested):
Somehow this PR breaks multiple things (other plugins?).

The tables in "Moving to Express 4" don't have `scope` on `<th>` elements, ~~because `<tbody>` is missing~~:

Image

If I'm reading the code (and old comments) correctly these should have `scope="row"` (according to code, not really how it should be). On a closer look `scope` seems to be missing **everywhere** (e.g. in the [routing guide](https://deploy-preview-2493--expressjscom-preview.netlify.app/en/5x/guide/routing/#response-methods)).

As a side note, I've noticed that the HTML `<table>`s have `<p>`s inside cells if the text is multiline. In the big tables ("behind proxies") this is may not be obvious, but in "Migrating to Express 4" only some rows have `<p>` elements and this causes them to have much bigger padding than other rows - rewriting them to Markdown (#2485) lets us avoid this and fixes the table structure (missing `<thead>`).

astro.config.mjs
ShubhamOulkar: Good catch, and thanks for the note. I only checked that scroll bars were being added for both HTML and Markdown-generated tables, and I m +1 on rewriting the two tables in #2485 to Markdown (Markdown tables render as semantic HTML in MDX and will produce thead/tbody structures). That approach makes sense and I support including those Markdown table changes from #2485. Converting every raw-HTML table across the site isn’t economical to do immediately we will keep converting pages over time as we come across them.

ShubhamOulkar: > Something in this PR breaks syntax highlighting and adding <a> to headers (all plugins?) and I think this may be the culprit.

fixed in 6cde5c8e1de6da71f570a65b43ed618892698ff2

src/plugins/rehype-accessible-tables.mjs
```diff
@@ -45,28 +73,7 @@ export function rehypeAccessibleTables() {
       };
 
       parent.children.splice(index, 1, wrapper);
-      // Skip the inserted wrapper so we don't revisit the table inside it
       return [SKIP, index + 1];
```
krzysdz, line 76: Since now there is only a single `visit()` call, we **must not** skip visiting the children of inserted node (the `<table>` may be visited twice, but there is a `parentAlreadyWrapped` check) or the `scope` won't be added. However, there's [a line in `unist-util-visit` readme](https://github.com/syntax-tree/unist-util-visit/blob/5d601df684ca7341646d6b57eb0f20fdfe277bc2/readme.md?plain=1#L159-L160) that makes me think that using something else than `SKIP` when replacing a node may lead to problems:

> Replacing `node` itself, if `SKIP` is not returned, still causes its descendants to be walked (which is a bug).

Maybe it would be safer to use a separate `visit()` call for adding the `scope`? I know that this comes with a performance hit (the website already takes a while to build), but it sounds like a better idea.

Even with this fixed, the tables in "Moving to Express 4" don't have the `scope`, because there is neither `<thead>` **nor `<tbody>`** in the source file - the `<table>` is directly followed by `<tr>`s:

https://github.com/expressjs/expressjs.com/blob/f7b0338cff90a9b8ff1d3a6ea645eeecb1aed5f3/src/content/pages/en/guide/migrating-4.mdx?plain=1#L46-L51

Will you copy the rewritten tables from #2485 to this PR, or should I just remove the added lines with wrapper (from "behind proxies") and make #2485 just a rewrite of those 2 HTML tables to markdown?

ShubhamOulkar: > Maybe it would be safer to use a separate visit() call for adding the scope? I know that this comes with a performance hit (the website already takes a while to build), but it sounds like a better idea.

I am aware of this and may be optimize will improve build time. https://docs.astro.build/en/guides/integrations-guide/mdx/#optimize

> make https://github.com/expressjs/expressjs.com/pull/2485 just a rewrite of those 2 HTML tables to markdown?

yes, rewrite tables from behind proxies. Rewritten markdown tables in "Moving to Express 4" LGTM.

ShubhamOulkar:
https://github.com/expressjs/expressjs.com/pull/2493/commits/4e49d373403538c743bd1bfeaaac9fcaa116d667 Uses separate passes to keep AST traversal predictable while supporting both HAST and MDX JSX tables. Build time remains unchanged.

src/plugins/rehype-accessible-tables.mjs
```diff
@@ -1,72 +1,157 @@
...
-            rehypePlugins: [rehypeAccessibleTables],
-          },
-        });
-      },
-    },
+      /*
+       * Don't continue traversing into the table after moving it.
+       * Scope processing has already happened in the previous passes.
+       */
+      return;
```
krzysdz: Returning `undefined` will cause the `visit()` to check also the children nodes. `SKIP` should be returned if we don't want to traverse the children nodes.

I tested adding `return SKIP` here, in [line 136](https://github.com/expressjs/expressjs.com/blob/4e49d373403538c743bd1bfeaaac9fcaa116d667/src/plugins/rehype-accessible-tables.mjs#L136), after [line 116](https://github.com/expressjs/expressjs.com/blob/4e49d373403538c743bd1bfeaaac9fcaa116d667/src/plugins/rehype-accessible-tables.mjs#L116) and after [line 93](https://github.com/expressjs/expressjs.com/blob/4e49d373403538c743bd1bfeaaac9fcaa116d667/src/plugins/rehype-accessible-tables.mjs#L93) - everything still works, no observable improvement in build time (the plugins probably don't affect it very much).

krzysdz:
This wrapper from a table in 5.1 release post can also be removed.

https://github.com/expressjs/expressjs.com/blob/e2152150e105a7e9c6bc7f89e7a5a51a2e0705b9/src/content/blog/2025-03-31-v5-1-latest-release.md?plain=1#L103

It doesn't do anything, because the `table-scroller` inside scrolls instead.

Firefox dev tools inspector: div with style "overflow-x: auto" that has a table-scroller dive inside. The table-scroller div is scrollable, because the table that it wraps overflows it. The outer div does nothing.

src/plugins/rehype-accessible-tables.mjs
ShubhamOulkar: > SKIP should be returned if we don't want to traverse the children nodes.

It's bug you mentioned earlier.

ShubhamOulkar:
removed html wrapper in https://github.com/expressjs/expressjs.com/pull/2493/commits/07c9ecab362d607a84554b205c2d9d67dea6bb76

image

krzysdz (approved):

krzysdz:
`scroller` (from [`<div class="table-scroller">`](https://github.com/express/expressjs.com/blob/b6b1bb5b34996bd8e9a0df4b3f38b2b8e81430d4/docs/content.md?plain=1#L332) in the new docs) is not recognised by CSpell. This can be fixed by adding it to the [`project-words.txt`](https://github.com/expressjs/expressjs.com/blob/6572271a6867e16195eac14d582a48fe7374a552/project-words.txt#L91-L92).

ceeprojects-hub:
scroller (from [<div class="table-scroller">](https://github.com/express/expressjs.com/blob/b6b1bb5b34996bd8e9a0df4b3f38b2b8e81430d4/docs/content.md?plain=1#L332) in the new docs) is not recognised by CSpell. This can be fixed by adding it to the [project-words.txt](https://github.com/expressjs/expressjs.com/blob/6572271a6867e16195eac14d582a48fe7374a552/project-words.txt#L91-L92)

Issue #988, opened 2018-10-13
Establish criteria and process to review middleware listed on the site


crandmck:
http://expressjs.com/en/resources/middleware.html#additional-middleware-modules lists some middleware modules that are note maintained by the Expressjs team.  

We get requests to add more modules to this list (see #968 and #981 for example), but as @wesleytodd noted:

> Adding to the website is an explicit recommendation to the community, and I think we should be judicious about that.

So, I suggest we
1. Establish some basic criteria for listing middleware in the above page. 
1. Review the existing list to ensure that all the modules meet those criteria.
1. Add a brief explanation of the criteria and process to http://expressjs.com/en/resources/contributing.html (or elsewhere, as appropriate).

Does anyone have suggestions for basic criteria for num. 1?   @expressjs/expressjs-com

dougwilson:
So this has been a similar issue for the `express-session` module, as it relies on folks implementing store plugins to be useful to the general population. Of course, actually vetting some kind of coding quality or something for things to be listed seemed to much of a burden for both sides (our side and the author side), and so the current criteria for `express-session` is just (a) it works with the current version of the module.

So to me, that comes back around to the Express middleware modules. I think there are a lot of useful middleware out there, even serving a purpose perhaps none of the Express team is familiar with to really know if it's "good" or not. Along the lines of the `express-session` criteria I would put forth the following initial criteria set for middleware we list:

(1) It works with the current version of Express
(2) It is not known to cause undue interference to the normal Express operations and case interoperability issue with other Express middleware

Not sure what else to really add that is going to be easy to "police".

brandon-d-mckay:
@dougwilson Something easy to add would be the package quality rating/score determined by npmjs.com/npms.io (although I think the latter may no longer be updating its results... the scores of my package (krauter =]) are updated on npmjs.com but not on there). A score of 90%+ can almost guarantee that the package at least has a readme, license, tests, linting, up-to-date dependencies, and a proper package.json.

If anyone's up for it, we could also modify the source code they use ([link here](https://github.com/npms-io/npms-analyzer)), and add our own criteria.

dougwilson:
[...]

Issue #981, opened 2018-09-18
Add the casbin-express-authz middleware to provide better authorization


hsluoyz:
Hi,

We developed an authorization (authz) middleware for Expresss called ``casbin-express-authz`` (https://github.com/node-casbin/express-authz) based on Casbin (https://casbin.org/), a cross-language authorization library project. For Node.js, our main library: https://github.com/casbin/node-casbin got certain popularity.

Casbin supports many access control models like ACL, RBAC, ABAC. The policy rules can be persisted into many databases like MySQL, PostgreSQL, SQLite, Microsoft SQL Server. It's very flexible and powerful. I think it will serve the authorization need for Express users.

I noticed there is already a 3rd-party middleware called ``express-stormpath`` doing the similar thing. But it has limited functionalities (only supports Group-based Access Control). Moreover, after ``stormpath`` is acquired, the project seems no longer active. So I believe it's necessary to provide a better and up-to-date authorization middleware.

Therefore, is it OK to add ``casbin-express-authz`` to the middleware list: https://expressjs.com/en/resources/middleware.html ? Thanks.

crandmck:
Let's defer this until #988 is resolved.

hsluoyz:
Any update on this issue?

bjohansebas:
I will close it, as we have been choosing to focus only on the Express documentation.

PR #2514, opened 2026-09-13
docs: add X-Trust human presence middleware example

htl-syterme:
I've been dealing with bot abuse on a side project where bots were hitting our API endpoints and burning through our free tier budget. Didn't want to add a captcha because it kills signups. Built a small open standard called X-Trust that reads a signed header and annotates the request with a human score.

It works as Express middleware in 3 lines. The client SDK measures keystroke timing and touch entropy, signs a score with HMAC-SHA256, attaches it as an X-Trust header. The server verifies the signature. Tokens expire in 120 seconds so replay attacks don't work.

This PR adds a documentation page showing how to use it with Express. It never blocks requests — just annotates them with a trust score so the app decides what to do.

OSS spec and live endpoint: https://github.com/htl-syterme/htl-core


krzysdz:
We currently don't add middleware recommendations, because we have no requirements specifying what could be added. For earlier examples of similar requests and discussions see:

- https://github.com/expressjs/expressjs.com/pull/1516#issuecomment-2108907327
- https://github.com/expressjs/expressjs.com/pull/2227#pullrequestreview-4031120692
- https://github.com/expressjs/expressjs.com/pull/2375#issuecomment-4650899198
- https://github.com/expressjs/expressjs.com/issues/966#issuecomment-413717603
- https://github.com/expressjs/expressjs.com/issues/981#issuecomment-433803932
- https://github.com/expressjs/expressjs.com/pull/968#issuecomment-413985270
- https://github.com/expressjs/expressjs.com/issues/988

Your account and the middleware are new (20 days old), the middleware ~~looks~~ is vibe coded, has 5 downloads in the last week from NPM, is paid (**in crypto only!**)[^1] and you "market it" heavily and in rather suspicious way, so if there were any guidelines on middleware that may be featured on the Express website, this probably wouldn't satisfy them. By the way, the `docs/` directory is not even a part of the website.

[^1]: There's also no information who the seller is, no terms of service, no nothing and the process **may be illegal** - I assume that there won't be an invoice either?

PR #2515, opened 2026-09-17
Fix broken URL in morgan.mdx custom format example

brmendez:
## Summary
- Fixed a stray GitHub URL that was accidentally inserted into the `tokens['response-time'](req, res)` call in the custom format function example, breaking the code sample.
- Scoped to `en` only. The same bad URL also exists in the translated copies (`ja`, `it`, `zh-cn`, `zh-tw`, `pt-br`, `de`, `ko`, `fr`, `es`).

## Test plan
- Verified the code block now renders valid JS matching the pattern used elsewhere on the page.


krzysdz:
This is a problem with the way how the `get-readmes.sh` script converts relative GitHub URLs. The script uses simple regular expressions and cannot detect that this is code inside a code block that looks like a link `[something](URL)`.

https://github.com/expressjs/expressjs.com/blob/c936b6310349ca92cf59493d4038dd663704f1a1/.github/scripts/get-readmes.sh#L157-L159

There is a workflow which replaces the content of this file with up-to-date readmes, so manual changes are overwritten. If you have an idea how to fix the script, then go ahead and edit this PR or create a new one.

cc @bjohansebas

brmendez:
Fixed by skipping fenced code blocks in the link-rewrite regex. Doesn't cover inline single-backtick code spans, but none of the current READMEs use that.

